PR #493, opened 2015-01-07
Only skip cleanup when the according env variable is set

aduffeck:
That way the packages/ directory will always be cleaned up when the
tests are run locally but it is still possible to skip the cleanup on
the CI server where we want to keep all built packages around for
download.


suse-tests-pass:
( Success: Merged 0891f90418650a4a3a933f7cfefdf6f929b837a6 into 84ddfd0511dcd814d761ec198aaddc968c367ca5 )

cornelius:
lgtm

PR #502, opened 2015-01-08
Cache Os info in LocalSystem

thardeck:
The method os of LocalSystem is access several times during build
prepration for example which would always result in a localhost os
inspection. To prevent this the result of the os inspection is cached so
it is only called once.

### DIFF
--- a/lib/local_system.rb
+++ b/lib/local_system.rb
@@ -16,13 +16,18 @@
 # you may find current contact information at www.suse.com
 
 class LocalSystem < System
+  @@os = nil
+
   class << self
     def os
-      description = SystemDescription.new("localhost",
-        SystemDescriptionMemoryStore.new)
-      inspector = OsInspector.new
-      inspector.inspect(System.for("localhost"), description)
-      description.os
+      if !@@os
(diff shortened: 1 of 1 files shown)


suse-tests-pass:
( Success: Merged ed9c920ee42450408cda36243ca974f2bbb084e6 into 7179ec894701e9d20ec9c290a6f9297b85c9f994 )

lib/local_system.rb
```diff
@@ -16,13 +16,17 @@
 # you may find current contact information at www.suse.com
 
 class LocalSystem < System
+  @@description = nil
```
cornelius: I would only cache the os object, not the complete description.

suse-tests-pass:
( Success: Merged 1eb2883b2dc07615a3db84d0af05a96ec5c7dfca into 7179ec894701e9d20ec9c290a6f9297b85c9f994 )

cornelius:
lgtm

suse-tests-pass:
( Success: Merged f022890f11c77be84cca4fcf24516d9e569435e4 into 7179ec894701e9d20ec9c290a6f9297b85c9f994 )

PR #504, opened 2015-01-08
Show warning when AutoYaST is exported without users scope

thardeck:
If a system is installed with AutoYaST without a user section the root
account will have an empty password which prevents the user from logging
in.

### DIFF
--- a/lib/autoyast.rb
+++ b/lib/autoyast.rb
@@ -26,6 +26,15 @@ class Autoyast < Exporter
       "repositories",
       "packages"
     )
+    if !description.users
+      Machinery::Ui.puts(
+        "\nWarning: Exporting a description without the scope 'users' as AutoYaST" \
+        " profile will result in a root account without a password which prevents" \
+        " logging in.\n" \
+        "So either inspect or add the scope 'users' before the export or" \
+        " add a section for the root user to the AutoYaST profile."
+      )
+    end
   end
 
   def write(output_dir)


suse-tests-pass:
( Success: Merged 4d581b7cf7be38431dc2d18ba4c06491d67e760d into 3fd168162fd8397623ff865175442ee8758c99b9 )

lib/autoyast.rb
```diff
@@ -23,6 +23,15 @@ def initialize(description)
       "repositories",
       "packages"
     )
+    if !description.users
+      Machinery::Ui.puts(
+        "\nWarning: Exporting a description without the scope 'users' as AutoYaST" \
+        " profile will result in a root account without a password which prevents" \
+        " logging in.\n" \
+        "So either a section for the root user needs to be added manually to the" \
```
cornelius: I would write this in active form "So either add a section ... or inspect before ..."

suse-tests-pass:
( Success: Merged 0f92fe86805d3071d11cc02fa7e5f8f65437df90 into 3fd168162fd8397623ff865175442ee8758c99b9 )

suse-tests-pass:
( Success: Merged 0f92fe86805d3071d11cc02fa7e5f8f65437df90 into c0ce4bd87fdabb0641991a694d820d142bfd60bb )

lib/autoyast.rb
```diff
@@ -23,6 +23,15 @@ def initialize(description)
       "repositories",
       "packages"
     )
+    if !description.users
+      Machinery::Ui.puts(
+        "\nWarning: Exporting a description without the scope 'users' as AutoYaST" \
+        " profile will result in a root account without a password which prevents" \
+        " logging in.\n" \
+        "So either inspect the scope 'users' before the export or" \
```
tgoettlicher: For now `machinery inspect` is the most common way to get a system description. But, there will be other ways to assemble a system description in the long run, e.g. using an editor. For now the current text looks okay, if we want to have the text future proof, I'd recommend: "So either add a 'users' section to the system description (e.g. by inspecting scope users) or"

thardeck: I have talked to Cornelius about this. I would have required the users scope for exporting Autoyast, similar to packages and repositories but this would have force the user to add this section manually to the manifest if he created the file on its own.
That's why we decided to add the warning to be not that restrictive.

tgoettlicher: Let my clarify: I didn't say that we should exit with an error. My comment was about rephrasing the warning, because this current text implies that system descriptions are always generated by inspect.

suse-tests-pass:
( Success: Merged cdadbee97fba099fb258b726688caf2a39129de3 into c0ce4bd87fdabb0641991a694d820d142bfd60bb )

tgoettlicher:
lgtm

PR #509, opened 2015-01-08
Prevent missing_scope comparison messages if both descriptions lack them

### LINKED ISSUE #498: "Unable to compare" message should only be shown if one of the descriptions has the scope
Especially in case of an identical description the warning "Unable to compare" is useless but even for non-identical descriptions it doesn't make much sense to show the message if both don't have this information.

```
Compared descriptions are identical.
# Unmanaged files
  Unable to compare, no data in 'd217', 'd217-copy'

# Changed managed files
  Unable to compare, no data in 'd217', 'd217-copy'

# Users
  Unable to compare, no data in 'd217', 'd217-copy'

# Groups
  Unable to compare, no data in 'd217', 'd217-copy'

# Changed configuration files
  Unable to compare, no data in 'd217', 'd217-copy'

# Services
  Unable to compare, no data in 'd217', 'd217-copy'
```

thardeck:
In doesn't make much sense in case of a comparison to inform about
missing scopes if both don't have them.
It also looks weird to inform about missing_scopes in case of an
identical comparison.
With this commit the missing scope message is only shown if it is
missing from only on scope.

Fixes #498

### DIFF
--- a/lib/renderer.rb
+++ b/lib/renderer.rb
@@ -148,16 +148,17 @@ class Renderer
     @stack = []
     missing_descriptions = Array.new
 
-    @buffer += "# #{display_name}\n"
-
-   if !description1[scope]
+    if !description1[scope]
       missing_descriptions << description1.name
     end
     if !description2[scope]
       missing_descriptions << description2.name
     end
 
-    indent { puts "Unable to compare, no data in '#{missing_descriptions.join("', '")}'" }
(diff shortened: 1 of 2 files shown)


suse-tests-pass:
( Success: Merged 538e79348bcd996fa1fb4893b7853b81a1a06c9e into 352a3609e43d25bf21f5879fbe55a5203270543a )

PR #514, opened 2015-01-12
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150112_extract_provisioning_code'

aduffeck:
  * 9cc8272 Extract subject image provisioning code into an external script

### DIFF
--- a/spec/definitions/provision.sh
+++ b/spec/definitions/provision.sh
@@ -1,0 +1,47 @@
+# unmanaged-files
+mkdir /usr/local/magicapp
+touch /usr/local/magicapp/one
+mkdir /usr/local/magicapp/data
+touch /usr/local/magicapp/data/two
+touch /etc/magicapp.conf
+mkdir /var/lib/chroot_proc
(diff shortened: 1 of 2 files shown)


suse-tests-pass:
( Success: Merged 9cc82725ec0d22c61b0317b4a2ce0894e60ba7ee into f8e4340656da679098a54e28f5c6ee35bca93e48 )

spec/definitions/provision.sh
```diff
@@ -0,0 +1,47 @@
+# unmanaged-files
```
tgoettlicher: `set -e` or `trap 'exit' ERR` could improve the robustness of this script.

cornelius:
Looks good from my point of view

spec/definitions/provision.sh
aduffeck: You may be right, but I'd defer this change for now. The script was only moved here and it's working fine in its current form.

PR #523, opened 2015-01-15
Enabled scope services to support Red Hat

tgoettlicher:


suse-tests-pass:
( Success: Merged ea81b5bc6ff6557200882289cb756708afe5d16a into 23f8d0bf3105e4507602b1ff020b7db0258181c3 )

suse-tests-pass:
( Success: Merged 7957469bdf0416ee256840395179c494cd0457a0 into 23f8d0bf3105e4507602b1ff020b7db0258181c3 )

plugins/inspect/services_inspector.rb
```diff
@@ -42,6 +47,13 @@ def has_systemd(system)
     false
   end
 
+  def has_redhat_chkconfig(system)
+    system.run_command("chkconfig", "--version")
```
cornelius: How does this check, if it has "redhat" chkconfig? It only tests for the existence of `chkconfig`, isn't it? How is this RedHat specific?

plugins/inspect/services_inspector.rb
```diff
@@ -42,6 +47,13 @@ def has_systemd(system)
...
+    true
+  rescue Cheetah::ExecutionFailed
```
cornelius: Checking for the existence of a tool, where it not being existant is a valid case, should not be done through exceptions. I would rather have a method in `System` which encapsulates that (something like `has_command?`) and return a boolean as result.

plugins/inspect/services_inspector.rb
```diff
@@ -85,4 +97,29 @@ def inspect_sysvinit_services(system)
...
+    @summary = "Found #{output.lines.size} services."
+
+    services = output.lines.map do |line|
+      state = []
+      name, state[0], state[1], state[2], state[3], state[4], state[5], state[6] = line.split(/\s+/)
+      Service.new(name: name, state: state[runlevel.to_i].split(":")[1])
+    end
+
+    ServiceList.new(services.sort_by(&:name))
+  end
```
cornelius: There is quite some duplication in this method with the `inspect_sysvinit_services` method. It might be better to explicitly code the difference into one method.

spec/unit/services_inspector_spec.rb
```diff
@@ -71,11 +78,14 @@
...
 
-    it "returns data about SysVinit services when no systemd is present" do
+    it "returns data about SysVinit services on a suse system when no systemd is present" do
       system = double
       expect(system).to receive(:run_command).
         with("systemctl", "--version").
         and_raise(Cheetah::ExecutionFailed.new(nil, nil, nil, nil))
+      expect(system).to receive(:run_command).
+        with("chkconfig", "--version").
+        and_raise(Cheetah::ExecutionFailed.new(nil, nil, nil, nil))
```
cornelius: It's not very nice that this tests has to be adapted, when the code change actually is not related to what is tested by this test. Is there a better way?

tgoettlicher: That's because we need to find out if SUSE's or Red Hat's chkconfig is on the system. If we determined the OS by running the OS inspector first we could skip this step.

tgoettlicher:
Squashed: https://github.com/SUSE/machinery/pull/525

PR #525, opened 2015-01-16
Enabled scope services to support Red Hat

tgoettlicher:

PR #528, opened 2015-01-19
Add has_command method to system

Werni:
This method returns true if the system to inspect has the given
command else it will return false.

### DIFF
--- a/lib/system.rb
+++ b/lib/system.rb
@@ -78,4 +78,11 @@ class System
 
     run_command("bash", "-c", script, *args)
   end
+
+  def has_command?(command)
+    run_command("bash", "-c", "type -P #{command}", stdout: :capture)
+    true
+  rescue Cheetah::ExecutionFailed
+    false
+  end
 end
--- a/spec/unit/system_spec.rb
+++ b/spec/unit/system_spec.rb
@@ -117,4 +117,18 @@ describe System do
       end
     end
   end
+
+  describe "#has_command" do
+    it "returns true if the system has the command" do
+      system = LocalSystem.new
+
+      expect(system.has_command?("echo")).to be(true)
+    end
+
+    it "returns false if the system hasn't the command" do
+      system = LocalSystem.new
+
+      expect(system.has_command?("not_existing_command")).to be(false)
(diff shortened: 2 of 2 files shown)


suse-tests-pass:
( Success: Merged c86a268bfa237eb69b26119c19e07dd14284921f into 50806d9ec88d48511f31c5efd67d677e13e71072 )

thardeck:
lgtm

thardeck:
I think the method is confusing. It doesn't just test if the command exists, it also runs it which is a problem imho because it can return a value unequal zero which would look like a non existing command.

thardeck:
I would use `type -p <command>` instead which is similar to `which` but should exist on all distributions.

suse-tests-pass:
( Success: Merged d23d8a9bf8ef88f9154006172c7df29b8f975cfb into 50806d9ec88d48511f31c5efd67d677e13e71072 )

suse-tests-pass:
( Success: Merged e4690d4597a5f70f358badc8ee5468468651aa4f into 50806d9ec88d48511f31c5efd67d677e13e71072 )

tgoettlicher:
There is also a has_command? method in my open pull request https://github.com/SUSE/machinery/pull/525 . It also checks if the commands work with specific arguments. That's needed by the services inspector to figure out if it's a SUSE or Red Hat system.

suse-tests-pass:
( Success: Merged 2575f0e8321f0dd17bf9ae17c652844cfdbc183d into 50806d9ec88d48511f31c5efd67d677e13e71072 )

thardeck:
But I think this implementation was wrong, like I mentioned in my commands. I would prefer Ricks implementation and if the has_command is true, then use run_command to actually run the command.

suse-tests-pass:
( Success: Merged 2dab0147a8fbd400f0b041ce6374a58b0af8690e into 50806d9ec88d48511f31c5efd67d677e13e71072 )

cornelius:
I agree with @thardeck. The implementation using `type` is more correct as it doesn't have side effects.

Patch looks good to me.

PR #535, opened 2015-01-20
Add prepare host script

Werni:
This script is needed to setup the host for running machinery. The
script pulls the newest sources, build the package and installing
the package.


suse-tests-pass:
( Success: Merged 99378170ccc49f63294fd3cf65cbb6225ec5e33a into eb41dec50a46ac11a329f11cbd24960f9ca6b0f2 )

spec/definitions/prepair_host.sh
```diff
@@ -0,0 +1,11 @@
+#update git
+cd ~/machinery/ && git pull
+
+#build package
+cd ~/machinery/ && rake rpm:build[https://api.suse.de,Devel:Alfred,SLE_12]
+
+#install package
+package=$(find ~/machinery/package/ \
+  -name "machinery-[0-9]*.ppc64le.rpm" -or \
+  -name "machinery-[0-9]*.s390x.rpm")
```
cornelius: Why do we need to list the architectures here? Can't we do that with a wildcard?

spec/definitions/prepair_host.sh
```diff
@@ -0,0 +1,11 @@
...
+sudo zypper in --force --auto-agree-with-licenses -y $package
```
cornelius: Why is the `--force` option needed here?

cornelius:
There is a typo in the name of the script. It should be `prepare_host.sh`.

spec/definitions/prepair_host.sh
Werni: The Problem is there is also a package named machinery-*.src.rpm and a wildcard will include them. We can filter them but I think this version is more robust.

spec/definitions/prepair_host.sh
Werni: Oh, yes it's not required was only for testing. I will remove them.

PR #537, opened 2015-01-21
Add prepare host script

Werni:
This script is needed to setup the host for running machinery. The
script pulls the newest sources, build the package and installing
the package.

Superseeds #535


suse-tests-pass:
( Success: Merged 7823c3a6114d56ea678bfc6226f65ee1c0019b1a into eb41dec50a46ac11a329f11cbd24960f9ca6b0f2 )

cornelius:
As this script is only needed by the internal integration tests I think we should move it to the internal repository where these tests are located as well.

PR #546, opened 2015-01-26
Add Version change and tracking bug id to RPM_CHANGES.

greygoo:


suse-tests-pass:
( Success: Merged 56e7ab5582d46014860aa1c47d000a03b233a172 into 175cb6b349dff2b34faa6e9863613edc828ba502 )

RPM_CHANGES
```diff
@@ -1,6 +1,11 @@
 # Machinery RPM Changelog
 
 
+## Version 1.3.0 - Mon Jan 26 13:07:06 CET 2015 - chuller@suse.de
+
+* update to version 1.3.0
```
thardeck: you just need to add the tracking but, everything else is done by our release script so the above part needs to be removed

PR #559, opened 2015-01-29
Use opt instead of tmp for integration test remote-dirs

thardeck:
Using /tmp as remote-dir prevented snapper under some circumstances
from rolling back the snapshot because the files couldn't be accessed.

Switching to opt prevented this issue.


suse-tests-pass:
( Success: Merged ac5e48dea756ab84650850b340ed9d9bf781aa50 into 89940ae63e1c07c5b043f1650e6dead28cfc15e0 )

cornelius:
I trust you that it solves the problem. The change looks fine.

PR #560, opened 2015-01-29
Filter zypper repos from unmanaged-file export

thardeck:
To prevent duplication of zypper repositories the files from
/etc/zypp/repos.d are filtered.


thardeck:
I filtered it from the export only. I think this is better that way so people don't run into the bug but still don't lose information since we don't store everything in the manifest yet to my knowledge.

Werni:
lgtm

PR #561, opened 2015-01-29
Format scopes in hint cmd properly

### LINKED ISSUE #541: Hint shows command that won't work
If you do an inspection without file extraction and afterwards try to export those, you are lacking the extracted files and machinery tells you so:

```
machinery export-kiwi --kiwi-dir /tmp localhost
The following scopes 'config-files, changed-managed-files, unmanaged-files' are part of the system description but the corresponding files weren't extracted during inspection.
The files are required to continue with this command. Run `/usr/bin/machinery inspect --extract-files --scope=config-files, changed-managed-files, unmanaged-files localhost` to extract them.
```

However - if you just copy and paste the command as given by the hint - which i assume many users will do - you only get an error message:

```
/usr/bin/machinery inspect --extract-files --scope=config-files, changed-managed-files, unmanaged-files localhost
Inspecting changed-managed-files, for config-files...
Could not establish SSH connection to host 'changed-managed-files,'. Please make sure that you can connect non-interactively as root, e.g. using ssh-agent.

To copy your default ssh key to the machine run:
ssh-copy-id root@changed-managed-files
```

The reason are the spaces between the commas, we should remove thos so the command is stated correct in the hint.

thardeck:
Remove the spaces between scopes to allow the execution without manual
interaction.

Fixes #541

### DIFF
--- a/lib/exceptions.rb
+++ b/lib/exceptions.rb
@@ -64,7 +64,7 @@ module Machinery
         end.compact.first || "<HOSTNAME>"
         formatted_scopes = Machinery::Ui.internal_scope_list_to_string(@scopes)
 
-        cmd = "#{$0} inspect --extract-files --scope=#{formatted_scopes}"
+        cmd = "#{$0} inspect --extract-files --scope=#{formatted_scopes.delete(" ")}"
         cmd += " --name='#{@description.name}'" if hostname != @description.name
         cmd += " #{hostname}"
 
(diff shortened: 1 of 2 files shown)


Werni:
lgtm

lib/exceptions.rb
```diff
@@ -64,7 +64,7 @@ def to_s
         end.compact.first || "<HOSTNAME>"
         formatted_scopes = Machinery::Ui.internal_scope_list_to_string(@scopes)
 
-        cmd = "#{$0} inspect --extract-files --scope=#{formatted_scopes}"
+        cmd = "#{$0} inspect --extract-files --scope=#{formatted_scopes.delete(" ")}"
```
cornelius: I would rather implement a different method to format the scopes without spaces. Adding spaces just to remove them again doesn't make too much sense.

PR #562, opened 2015-01-30
Fix KIWI export integration test

thardeck:
The ls listing of the KIWI export wasn't adapted after adding the new
filter for /etc/zypp/repos.d.


Werni:
lgtm

PR #563, opened 2015-01-30
Inspection of yum repositories

tgoettlicher:
DON'T MERGE!

This is a proposal how to handle yum repositories. Please comment.


suse-tests-pass:
( Success: Merged 44360e6b13259d5c667c812f6efbade30f1627ae into e5c57eea49e5696500b981c161af5718849a7b10 )

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+  * yum
+    * yum 1
+    * yum 2
+```
+
+* Advantages:
+  * Lists repos separately
+  * Different repos type can contain different attributes
+* Disadvantages:
+  * Schema migration needed
```
cornelius: * No common representation, complicates code dealing with repositories in a more abstract way, e.g. for display or comparison

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+  * It's not clear how to get the repo's url using the api, yet
+  * Copying script to target needed
+
+### Calling `yum-config-manager`
+* Advantages:
+  * Lists all repo attributes (incl. defaults)
+* Disadvantages:
+  * Hard to parse
+  * Disabled repos missing
+
```
cornelius: What about the variables used in repo files? Should we store them or the expanded value? Where do the values of the variables come from, and should we capture that as well?

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+  * zypp 1
+  * zypp 2
+  * yum 1
+  * yum 2
+```
+This approach includes storing the repository type.
+That means each entry in the repository list will get another attribute `type` which can be `zypp` or `yum`.
+
+* Advantages:
+  none
```
cornelius: Common representation, which makes it easier to deal with repositories in a more abstract way, e.g. for display or comparison

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+
+```
+* repositories
+  * zypp 1
+  * zypp 2
+  * yum 1
+  * yum 2
+```
+This approach includes storing the repository type.
+That means each entry in the repository list will get another attribute `type` which can be `zypp` or `yum`.
```
cornelius: Strictly speaking this is not an attribute of the repository, but of the way the repository is used in the system. The same repository could be used with both, zypper and yum.

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+  * yum 1
+  * yum 2
+```
+This approach is similar to the approach above but doesn't include storing the repository type.
+All attributes of yum repos are mapped to zypp's attributes.
+
+* Advantages:
+  * No schema migration needed
+* Disadvantages:
+  * Not easy to build images using the available data
```
cornelius: "Not easy" means impossible, or not? If we don't know, whether to use yum or zypper in the resulting system, we have no way to replicate the system from the description.

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Considerations
+
+When reading the list of repositories we want to get all repositories, thus
+calling `yum-config-manager` or reading `/etc/yum.repos.d` doesn't work for us.
+`yum -v repolist all` is the prefered choice if we can deal with the line breaks in its output,
+because then we don't have to copy a file to the target system.
+
+The usecase for storing yum repositories is the inspection of a Red Hat system
+and creation of a human readable report. Building Red Hat images is out of scope
```
cornelius: Building Red Hat images is a less relevant use case. But there are more relevant use cases, such as migrating a Red Hat system to a SUSE system, or to compare a Red Hat system with a SUSE system. What are the implications of these?

When we only want to use the description as result of an inspection and not for anything else, we might want to think about if we need to flag the description somehow, so the tool can help the user to do the right things with it.

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+  * Lists all repos
+* Disadvantages:
+  * Parsing not trivial (line breaks depend on length of output)
+  * Not all repo attributes accessible
+
+### Parse the repo files in `/etc/yum.repos.d`
+* Advantages:
+  * list (almost) all repo attributes
+* Disadvantages:
+  * Repos from plugins missing (e.g. suse manager's repos)
```
cornelius: Where do plugins store their configuration?

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,125 @@
...
+## Reading the Repository Data from the System
+
+This section presents different ways of accessing the yum repository data and lists their advantages and disadvantages.
+
+### Parse output of `yum -v repolist all`
+* Advantages:
+  * Lists all repos
+* Disadvantages:
+  * Parsing not trivial (line breaks depend on length of output)
+  * Not all repo attributes accessible
```
cornelius: Which attributes are not accessible?

docs/Inspection-of-yum-Repositories.md
thardeck: Since yum does have neither a priority nor an autorefresh attribute they are not hard to distinguish atm so I vote for this option.

This allows us to ship the support for our users soon but still have time to prepare a more wider manifest optimization without forcing the user to run two migrations in a short time.

suse-tests-pass:
( Success: Merged 44360e6b13259d5c667c812f6efbade30f1627ae into 308e5dddbad8a50c0b3860e71aed1ab3157bea37 )

docs/Inspection-of-yum-Repositories.md
cornelius: Basing the distinction on the presence or values of some attributes is a very fragile and error-prone approach. I don't think we should rely on this.

docs/Inspection-of-yum-Repositories.md
tgoettlicher: The SUSE Manager repository plugin retrieves the data from the SUSE Manager server.

docs/Inspection-of-yum-Repositories.md
tgoettlicher: Automatic migration from one distribution to another or even comparision between different operating system versions will need abstraction. This does not only affect repositories but the other scopes as well. It's a fundamental design question which we better address in a seperate discussion.

suse-tests-pass:
( Success: Merged e03eaa2cee0ed2017b4830a5df31884ba04dc612 into 308e5dddbad8a50c0b3860e71aed1ab3157bea37 )

docs/Inspection-of-yum-Repositories.md
cornelius: Where is the information stored from which server to get the data?

suse-tests-pass:
( Success: Merged 72524d6e8ef89aa5cbf648b0ccfbc1cb8a049617 into 308e5dddbad8a50c0b3860e71aed1ab3157bea37 )

suse-tests-pass:
( Success: Merged 72524d6e8ef89aa5cbf648b0ccfbc1cb8a049617 into 78f97d20bdd3279abfbd7a7c32df05d3c52f3a92 )

docs/Inspection-of-yum-Repositories.md
```diff
@@ -0,0 +1,231 @@
...
+  * list (almost) all repo attributes
+* Disadvantages:
+  * Repos from plugins missing (e.g. suse manager's repos)
+
+### Accessing the python api
+* Advantages:
+  * All repos and repo attributes accessible (you need to know the attribute's name, though)
+* Disadvantages:
+  * It's not clear how to get the repo's url using the api, yet
+  * Copying script to target needed
```
Werni: We can run the script directly in the shell without copying.
`python -c 'print "Test"'`

PR #568, opened 2015-02-04
Define the default vagarant shar explicitly in Vagrantfile.

greygoo:
- We encountered issues with vagrant trying to install nfs packages using
  yum, which failed
- This seem to happen sometimes as we do not have the default vagrant share
  defined in the Vagrantfile
- By adding it explicitly and defining it to use rsync instead of nfs this
  issue should not occur again

### DIFF
--- a/spec/definitions/vagrant/Vagrantfile
+++ b/spec/definitions/vagrant/Vagrantfile
@@ -97,6 +97,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # the path on the guest to mount the folder. And the optional third
   # argument is a set of non-required options.
   # config.vm.synced_folder "../data", "/vagrant_data"
+  config.vm.synced_folder '.', "/vagrant", type: "rsync"
 
   # Provider-specific configuration so you can fine-tune various
   # backing providers for Vagrant. These expose provider-specific options.


suse-tests-pass:
( Success: Merged de459580162c396ef9692b3ffb83c7d18e797a27 into 78f97d20bdd3279abfbd7a7c32df05d3c52f3a92 )

suse-tests-pass:
( Success: Merged de459580162c396ef9692b3ffb83c7d18e797a27 into beb797124d1f510110c3765a18b40405c0071cef )

suse-tests-pass:
( Success: Merged de459580162c396ef9692b3ffb83c7d18e797a27 into bfffd41f64708d7ffa147994035dd519ea5009aa )

suse-tests-pass:
( Success: Merged de459580162c396ef9692b3ffb83c7d18e797a27 into 5960b08e05ca243e863d7e95313d893a69f27536 )

suse-tests-pass:
( Success: Merged de459580162c396ef9692b3ffb83c7d18e797a27 into b84e426d717244615b9c983a03284137c17411e1 )

thardeck:
According to https://docs.vagrantup.com/v2/synced-folders/basic_usage.html it syncs the folder to the vm. So what would "." be?

It seems that I already have a /vagrant with the machinery_rpm_provisioner.rb so that does explain why the tests didn't break during my tests.

greygoo:
by default vagrant shares . as /vagrant in the guest. This default is nowhere configured as far as i could see, so vagrant creates it by itself, and in some cases uses nfs , in some rsync. When using nfs - it attempts to install nfs-client in the guest using yum thus breaking the startup. By explicitly defining it we make sure it always is using rsync

thardeck:
lgtm

PR #572, opened 2015-02-05
Reworked Inspection-of-yum-Repositories

tgoettlicher:


suse-tests-pass:
( Success: Merged 92f0186329abe5fca474a1f687c8c6c37cb8f457 into db2431c46b2fa4d8ab2359905043b22fc0956636 )

docs/Inspection-of-yum-Repositories.md
```diff
@@ -42,10 +43,11 @@ You can find an example script below in the References section.
 * Disadvantages:
   * Hard to parse
   * Disabled repos missing
-  * Variables `$releasever`, `$basearch` in repo files cause raise questions:
+  * Variables `$releasever`, `$basearch` in repo files raise questions:
     * What about the variables used in repo files?
     * Should we store them or the expanded value?
     * Where do the values of the variables come from, and should we capture that as well?
```
cornelius: It would be great to have some answers to these questions, so we can decide how to deal with them.

docs/Inspection-of-yum-Repositories.md
```diff
@@ -199,6 +201,23 @@ Attributes of yum repos are mapped to zypp's attributes.
 * Disadvantages:
   * Building Red Hat images not straight forward: We can find out that it is a yum repo because it doesn't have priority nor autorefresh
 
+
+## Use Cases
```
cornelius: I think the use cases should be at the beginning of the document and define the goal we want to achieve with the implementation.

suse-tests-pass:
( Success: Merged 92f0186329abe5fca474a1f687c8c6c37cb8f457 into aa55fc1ec1c1d1a0a40131f7b5397e637cfa66ec )

docs/Inspection-of-yum-Repositories.md
```diff
@@ -199,6 +201,23 @@ Attributes of yum repos are mapped to zypp's attributes.
 * Disadvantages:
   * Building Red Hat images not straight forward: We can find out that it is a yum repo because it doesn't have priority nor autorefresh
```
cornelius: Finding out if it's a yum repo by missing attributes is a fragile and error-prone solution. I don't think we should do that. I would see it as a better solution to add an explicit attribute storing the package manager which is used to manage the repository.

If this would be an optional attribute defaulting to zypp we might even get away without schema change as the new schema would be compatible with the old one.

cornelius:
In the "flat data section with schema change" there are entries on the same level with differing content. This can't be expressed well in a schema. It could be done by defining a common element with lots of optional sub element, but this would allow mixed invalid states.

If we want to have different data for different package managers, we need separate elements. This would make it equivalent to the "separate sub-sections" variant.

The variant we didn't consider so far is extending the schema in a compatible way by adding optional entries. This might be a good compromise for storing the relevant data without having to change the schema while still being able to mostly treat the data in a general way without having to special-case package managers.

suse-tests-pass:
( Success: Merged 999347f70579fdf64ca34f1695f76910ca8f8a65 into aa55fc1ec1c1d1a0a40131f7b5397e637cfa66ec )

suse-tests-pass:
( Success: Merged 97ac455f619e506038402b30c351ffa64dea6be0 into aa55fc1ec1c1d1a0a40131f7b5397e637cfa66ec )

suse-tests-pass:
( Success: Merged 97ac455f619e506038402b30c351ffa64dea6be0 into 9799dc47c06c3f875e28ec61844a30019532f3e9 )

suse-tests-pass:
( Success: Merged 97ac455f619e506038402b30c351ffa64dea6be0 into 502554be42e09229012ef95862aa1733bec58c47 )

suse-tests-pass:
( Success: Merged 79aaf104abb52d1446b691eb03077eb885f36fd5 into 502554be42e09229012ef95862aa1733bec58c47 )

tgoettlicher:
Cornelius, your commits look good to me. Can we merge?

PR #573, opened 2015-02-06
Use a regex to check the unit test output

Werni:
The name of the fake schema can change on other systems so we must use a
regex.


thardeck:
lgtm

suse-tests-pass:
( Success: Merged 9115e2887b593aec6e66d9e7e851d76b000cbc83 into aa55fc1ec1c1d1a0a40131f7b5397e637cfa66ec )

PR #587, opened 2015-02-13
Fix repository extraction on Fedora

thardeck:
Extraction of repositories on Fedora failed because baseurl isn't
available and accessing an element in an non existing array will fail.

By switching the attributes it works for both types, baseurl and
metalink.

### DIFF
--- a/helpers/yum_repositories.py
+++ b/helpers/yum_repositories.py
@@ -28,7 +28,7 @@ for repo in yb.repos.sort():
   repo_dict["alias"] = repo.id
   repo_dict["name"] = repo.name
   repo_dict["type"] = "rpm-md"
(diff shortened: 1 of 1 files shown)


suse-tests-pass:
( Success: Merged 499ee87b525387a5bac6c6d34177027ab93c94c4 into 333011a35059c8bbcb8299d70f931adb30f4c2f9 )

aduffeck:
LGTM, but you should rather use `||` than `or` for boolean algebra, `or` is meant for flow control and doesn't always behave as expected (same for `and` vs `&&`)

aduffeck:
Examples where `and` and `or` behave differently (because of different operator precedence):

```
irb(main):022:0> a = false or true; a
=> false
irb(main):023:0> a = true and false; a
=> true
```

thardeck:
It seems that Python has no "||" operator.

PR #605, opened 2015-02-17
Clean up patterns inspector

thardeck:
Since yum doesn't support patterns and zypper is the only supported
program for inspecting patterns we only extract data for zypper and just
show an unsupported message on all other systems.

### DIFF
--- a/plugins/inspect/patterns_inspector.rb
+++ b/plugins/inspect/patterns_inspector.rb
@@ -19,12 +19,9 @@ class PatternsInspector < Inspector
   def inspect(system, description, options = {})
     if system.has_command?("zypper")
       inspect_with_zypper(system, description)
-    elsif system.has_command?("yum")
-      inspect_with_yum(description)
     else
-      raise Machinery::Errors::MissingRequirement.new(
-        "Need binary zypper or yum to be available on the inspected system."
-      )
+      description.patterns = PatternsScope.new
+      "Patterns are not supported on this system."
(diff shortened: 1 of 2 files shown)


suse-tests-pass:
( Success: Merged de7be5816781fdb163b4f56e7c3eb72796a59dea into 80511166d243670063978e103ac8f045573f96e3 )

plugins/inspect/patterns_inspector.rb
```diff
@@ -19,12 +19,9 @@ class PatternsInspector < Inspector
...
     if system.has_command?("zypper")
       inspect_with_zypper(system, description)
-    elsif system.has_command?("yum")
-      inspect_with_yum(description)
     else
-      raise Machinery::Errors::MissingRequirement.new(
-        "Need binary zypper or yum to be available on the inspected system."
-      )
+      description.patterns = PatternsScope.new
+      "Patterns are not supported on this system."
```
tgoettlicher: Nitpicking: "Inspection of patterns is not supported on this system."

tgoettlicher:
Feel free to change the wording or not. LGTM

plugins/inspect/patterns_inspector.rb
thardeck: The complete message looks like this so I think it is better that way.
```
Inspecting patterns...
 -> Patterns are not supported on this system.
```

PR #619, opened 2015-02-20
[Review] Request from 'aduffeck' @ 'SUSE/machinery/local_integration_tests'

aduffeck:
  * ef6c2ed Be more robust when cleaning up the export directories
  * 5c04a68 Encapsulate the machinery environment (path, user, group)
  * 2c0d47a For local machinery we only have to set up the subject system's host key
  * ffd3a19 Also run the autoyast export tests locally
  * 77166b8 Fix overriding the kiwi export test settings when running locally
  * f84c646 Add a new suite of specs that are run on the locally instead of in a VM
  * f339562 Parameterize KIWI export examples so that they can be used locally, too
  * 184191b Allow for overriding the machinery store using the MACHINERY_DIR env var


lib/machinery.rb
```diff
@@ -33,6 +33,7 @@
 require "find"
 require "pathname"
 require "nokogiri"
+require "etc"
```
cornelius: This is only used in the tests, isn't it? Then it should rather be included in `spec_helper.rb`.

spec/integration/machinery_local_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    owner: current_user,
+    group: group
+  }
+  let(:machinery_config) { machinery_config }
+
+  before(:all) do
+    @machinery = start_system(
+      local: true,
+      env: { "MACHINERY_DIR" => machinery_config[:machinery_dir] },
+      command_map: { "machinery" => File.join(Machinery::ROOT, "bin/machinery") }
```
cornelius: The `command_map` option looks a bit complicated. Would it also work to just add the local `bin` directory to the `PATH` environment variable?

PR #629, opened 2015-02-23
Remove autoyast integration test file comparison

thardeck:
The file content is already checked in the unit test so no need to have
two places to check the same thing.


aduffeck:
lgtm

PR #635, opened 2015-02-24
[Review] Request from 'aduffeck' @ 'SUSE/machinery/local_integration_tests'

aduffeck:
  * 8b5d2c6 command_map is not needed, we just tweak PATH to prefer local machinery
  * d6242ba Only require 'etc' for the integration tests
  * 72bb72b Be more robust when cleaning up the export directories
  * 5f346ce Encapsulate the machinery environment (path, user, group)
  * dc9bcd1 For local machinery we only have to set up the subject system's host key
  * a4bc019 Also run the autoyast export tests locally
  * 9bc5a45 Fix overriding the kiwi export test settings when running locally
  * 07b34c5 Add a new suite of specs that are run on the locally instead of in a VM
  * e6b6eb8 Parameterize KIWI export examples so that they can be used locally, too
  * b785574 Allow for overriding the machinery store using the MACHINERY_DIR env var

### DIFF
--- a/bin/machinery
+++ b/bin/machinery
@@ -22,7 +22,9 @@ require_relative '../lib/machinery'
 begin
   LocalSystem.validate_machinery_compatibility
   Machinery.initialize_logger(ENV["MACHINERY_LOG_FILE"] || Machinery::DEFAULT_LOG_FILE)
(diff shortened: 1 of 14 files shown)


aduffeck:
Supersedes #619

cornelius:
looks good to me. I think `given_filesystem` should also work across examples, but I'm fine with refactoring that later.

PR #640, opened 2015-02-25
Use absolute path to binary in tests that are run locally and remote

aduffeck:
That fixes the problem with sudo, where the PATH environment variable
was not passed through and the system machinery was used instead of the
git one.


greygoo:
lgtm but maybe somebody else has a look too. Have not yet checked out and tested.

thardeck:
lgtm

PR #641, opened 2015-02-25
Require package_manager in repositories scope

### LINKED ISSUE #630: package_manager is not required in schema 3
it was added with a type check but it is not required.

I think it should be required so it should be added to schema 4 and should be included in the migration 3to4 again just to make sure.

thardeck:
The attribute was added to the migration and inspection but it wasn't
marked as required in the schema so this was fixed now as discussed.

Fixes #630


tgoettlicher:
As there is already a migration for this attribute, your commit looks good to me.

PR #642, opened 2015-02-26
Improve `machinery list` output in error cases

aduffeck:
This commit makes the usage of the different exception classes more
consistent and handles certain cases differently in `machinery list`.

Machinery::Errors::SystemDescriptionIncompatible
  This exception class is used when a system description uses a
  different or an unknown format version.

Machinery::Errors::SystemDescriptionValidationFailed
  This class is used when a system description is compatible (same
  format version), but a validation failed (either of the JSON or the
  file structure)

Machinery::Errors::SystemDescriptionError
  This is the exception class that's used when something unexpected
  happened while loading the system description, e.g. when a scope can
  not be mapped to a class.


lib/list_task.rb
```diff
@@ -23,10 +23,23 @@ def list(store, options = {})
...
+          Machinery::Ui.puts " #{name}: format version #{e.format_version}. " \
+            "Please upgrade Machinery to the latest version.\n\n"
+        end
+        next
+      rescue Machinery::Errors::SystemDescriptionValidationFailed
+        Machinery::Ui.puts " #{name}: This description is broken. Use " \
+          "`#{$0} validate #{name}` to see the error message.\n\n"
+        next
       rescue Machinery::Errors::SystemDescriptionError
-        Machinery::Ui.puts " #{name}:\n"
```
thardeck: Machinery::Errors::SystemDescriptionError seems to be used quite often. Does this only happen if the json parser has issues?

spec/unit/system_description_spec.rb
```diff
@@ -68,7 +68,7 @@ class SystemDescriptionFooConfig < Machinery::Object; end
           SystemDescriptionMemoryStore.new,
```
thardeck: Please change the title of the it block according to your change.

thardeck:
lgtm

PR #643, opened 2015-02-26
Fix 638

### LINKED ISSUE #638: Option to get a short list of descriptions
`machinery list` lists all descriptions including all the scopes which are contained in each description. This gives quite long output, which is hard to read, when you are just looking for the name of a description. It would be good to have an option to get a short list only showing one line per description.

thardeck:
Fixes #638


tgoettlicher:
I'd expect the `--short` option to only affect the command's output. The rest (e.g. error checking) should be the same as without `--short`.

PR #645, opened 2015-02-27
Do not choke on broken JSON.

aduffeck:
Fixes #584

PR #649, opened 2015-02-27
[Review] Request from 'tgoettlicher' @ 'SUSE/machinery/review_150227_export_autoyast_profile_which_only_installs_required_packages'

tgoettlicher:
  * f93512f Export autoyast profile which only installs required packages
  * 6e1de0c Merge pull request #647 from SUSE/fix_638_2
  * e12a539 Add list --short option to changelog
  * 540a1de Update manpage for list --short option
  * 7b2ec5c Add short option to list
  * 513dcd1 Remove obsolete option description name cleanup in list
  * ffca70a Replace hashes in cli.rb through Ruby 1.9 syntax
  * 8a5b940 Replace verbose string option by symbol
  * dcfa701 Merge pull request #645 from SUSE/review_150227_fix_json_error_handling
  * 7c047cf Do not choke on broken JSON.

### DIFF
--- a/lib/autoyast.rb
+++ b/lib/autoyast.rb
@@ -74,6 +74,7 @@ class Autoyast < Exporter
         apply_basic_network(xml)
         apply_repositories(xml)
         xml.software do
+          apply_software_settings(xml)
           apply_packages(xml)
           apply_patterns(xml)
         end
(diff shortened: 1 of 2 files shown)


thardeck:
Shouldn't this behavior documented somewhere?

tgoettlicher:
If a user inspects a system and exports it as autoyast profile, they will expect the system to be as similar as possible. Thus, this behaviour does not need to be documented.

thardeck:
True, we would install everything else anyway so lgtm.

Issue #644, opened 2015-02-26
autoyast fails when hostip is specified


tgoettlicher:
The exported autoyast installation fails when the hostip incl. netmask is specified behind the autoyast parameter:
``autoyast=http://192.168.1.1/install/autoyast/autoinst.xml hostip=192.168.1.2/24``

This makes the sed expression fail: 
 https://github.com/SUSE/machinery/blob/master/lib/autoyast.rb#L253

The result is corrupted config files on the installed system.

thardeck:
Fixed

PR #651, opened 2015-02-27
Fix parsing kernel command line (issue #644)

tgoettlicher:
This commit fixes parsing the kernel command line. Parsing failed
when there was another option after the autoyast= option which
contained a slash character ('/'). The new regular expression
fixes this issue.


thardeck:
lgtm

PR #657, opened 2015-03-03
[Review] Request from 'aduffeck' @ 'SUSE/machinery/rbr_filtering'

aduffeck:
Review-By-Refactoring branch for #654


lib/filter.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def matches?(path, value)
+    filter = filter_for(path)
+    return false if !filter
+
+    filter.matches?(value)
+  end
+
+  def filter_for(path)
```
cornelius: This should probably be `element_filter_for`.

plugins/inspect/unmanaged_files_inspector.rb
```diff
@@ -273,9 +304,13 @@ def inspect(system, description, options = nil)
     unmanaged_links = {}
     remote_dirs = mount_points.remote
     special_dirs = mount_points.special
-    ignore_list.each do |ignore|
-      remote_dirs.delete_if { |e| e.start_with?(File.join("/", ignore, "/")) }
-    end
+
+    filter_locator = "/unmanaged_files/files/name"
+    filter_dirs = ElementFilter.new(filter_locator)
```
cornelius: Maybe we could actually leave out the filter locator here (according to our new terminology this should be a path anyway). If the `ElementFilter` is used directly as in this inspector, the path is not relevant. Would it just work to allow a constructor without arguments?

This makes me think a bit, if the path actually should be part of the `ElementFilter` at all, as it's not used in the class itself.

cornelius:
Generally this looks like a good set of classes now. There certainly will be further refinements necessary as we move forward, but for a start it looks good.

When we agree on this as a starting point it would be good to document the responsibilities of the classes and the terminology we use, so that we have a common and consistent view on that.

As a further step It would be great to simplify the unmanaged files inspector code to be able to use more simple filters. There is some unnecessary duplication in the two list of matchers we use there right now.

aduffeck:
Merging as discussed in person.

PR #665, opened 2015-03-04
Don't extract the root directory as changed managed file

tgoettlicher:
If the mode of the root directory ('/') has changed we don't want to
extract the root directory because this will lead to extraction of
its sub-directories. This approach excludes the root directory from
extraction, the meta information is still stored in the manifest.


thardeck:
lgtm

PR #667, opened 2015-03-06
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150306_store_used_filter_in_description'

aduffeck:
  * cfaf825 Extract filter preparation and handling code into separate methods
  * 8d28733 Adapt schema to allow storing the used filter during inspection
  * 4a49a06 Adapt to new signature of InspectTask#inspect_system
  * 42b59fe Store filter which was used for inspection in the description metadata
  * 42a0f6c Add Filter#to_array which returns an array of element filter definitions
  * 42c76a0 Fix parsing of filter definitions with simple lists
  * 0c93bc5 Add Filter#add_filter_definition which adds new element filters
  * 85a2ec3 Make Filter#parse_filter_definition a class method

### DIFF
--- a/lib/inspect_task.rb
+++ b/lib/inspect_task.rb
@@ -16,11 +16,11 @@
 # you may find current contact information at www.suse.com
 
 class InspectTask
-  def inspect_system(store, host, name, current_user, scopes, options = {})
+  def inspect_system(store, host, name, current_user, scopes, filter_definition, options = {})
     system = System.for(host)
     check_root(system, current_user)
 
-    description, failed_inspections = build_description(store, name, system, scopes, options)
+    description, failed_inspections = build_description(store, name, system, scopes, filter_definition, options)
 
     if !description.attributes.empty?
       print_description(description, scopes) if options[:show]
@@ -57,7 +57,7 @@ class InspectTask
     end
   end
 
-  def build_description(store, name, system, scopes, options)
+  def build_description(store, name, system, scopes, filter_definition, options)
     begin
       description = SystemDescription.load(name, store)
     rescue Machinery::Errors::SystemDescriptionNotFound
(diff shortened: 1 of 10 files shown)


spec/unit/filter_spec.rb
```diff
@@ -2,12 +2,46 @@
 
 describe Filter do
   let(:definition1) { "/unmanaged_files/files/name=/home/alfred" }
-  let(:definition2) { "/unmanaged_files/files/name=/home/alfred,/var/cache" }
+  let(:definition2) { "\"/unmanaged_files/files/name=/home/alfred,/var/cache\"" }
```
cornelius: We defined the syntax in the design document a little bit different. In this form it would match an element in the JSON where the value is an array of `["/home/alfred", "/var/cache"]` according to the spec. But the syntax you use here is a shortcut to having to filter deinitions `/unmanaged_files/files/name=/home/alfred` and `/unmanaged_files/files/name=/var/cache`, right?

At the current point in time I wouldn't introduce these kind of shortcuts, but first get the basic filtering implemented how we planned, including command line interface and persistent storage. Then we are in a better position to judge what is needed and practical to improve the syntax.

cornelius:
We should add some API documentation to the `Filter` class, which describes how you create filters, what filter definitions are supported, and how they are used.

lib/inspect_task.rb
```diff
@@ -16,11 +16,12 @@
 # you may find current contact information at www.suse.com
 
 class InspectTask
-  def inspect_system(store, host, name, current_user, scopes, options = {})
+  def inspect_system(store, host, name, current_user, scopes, filter_definition, options = {})
```
cornelius: Wouldn't it make more sense to pass a filter object here instead of a filter definition? Then the preparation of the filter could happen in the code calling the `InspectTask` and there would be no need to pass the `:skip_files` option.

PR #668, opened 2015-03-06
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150306_store_used_filter_in_description'

aduffeck:
  * e2b1627 Add documentation for the Filter class
  * 0c1861b Add copyright notes
  * df6bed7 Fix filter behavior regarding arrays in the filter definition
  * b00bc21 Do not require quoting of array filters when there's only one filter
  * 3debe88 Improve method name
  * 89bacf1 Fix hound issues
  * cfaf825 Extract filter preparation and handling code into separate methods
  * 8d28733 Adapt schema to allow storing the used filter during inspection
  * 4a49a06 Adapt to new signature of InspectTask#inspect_system
  * 42b59fe Store filter which was used for inspection in the description metadata
  * 42a0f6c Add Filter#to_array which returns an array of element filter definitions
  * 42c76a0 Fix parsing of filter definitions with simple lists
  * 0c93bc5 Add Filter#add_filter_definition which adds new element filters
  * 85a2ec3 Make Filter#parse_filter_definition a class method

### DIFF
--- a/helpers/default_filters.json
+++ b/helpers/default_filters.json
@@ -1,0 +1,24 @@
+{
+  "inspect": [
+    "/unmanaged_files/files/name=/etc/passwd",
+    "/unmanaged_files/files/name=/etc/shadow",
+    "/unmanaged_files/files/name=/etc/group",
+    "/unmanaged_files/files/name=/tmp",
+    "/unmanaged_files/files/name=/var/tmp",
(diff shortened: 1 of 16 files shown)


cornelius:
The patch looks good to me know. We should do careful QA, though.

PR #675, opened 2015-03-09
Fix 674

### LINKED ISSUE #674: other_rpm_changes is only used if there are not other known changes
The rpm changed file tag `other_rpm_changes` is only used if there is no other known rpm change.

This doesn't fit the acceptance criteria `All other changes supported by RPM are reported as "other_rpm_changes" `

thardeck:
The code isn't that great and I would be happy about reviews by refactoring. :)

Fixes #674


lib/changed_rpm_files_helper.rb
```diff
@@ -16,10 +16,21 @@
 # you may find current contact information at www.suse.com
 
 module ChangedRpmFilesHelper
+  def expected_tag?(rpm_changes, character, position)
+    if rpm_changes[position] == character
+      true
+    else
+      @unknown_tag = true if ![".", "?"].include?(rpm_changes[position])
```
aduffeck: The `true if` part is superfluous.

aduffeck: I'd prefer `unknown_tag ||= ...` here.

tgoettlicher:
It's critical to have issue #674 fixed, lgtm.

PR #686, opened 2015-03-10
Make sure that filtering directories works with and without trailing '/'

### LINKED ISSUE #685: Filters with trailing `/` don't match directories
`machinery inspect --skip-files=/var/cache/` doesn't filter `/var/cache` and its sub directories. My expectation as a user would be that it excludes these directories from the inspection.

aduffeck:
Fixes #685


thardeck:
lgtm besides my comment

thardeck:
lgtm

PR #691, opened 2015-03-11
Fix 688

### LINKED ISSUE #688: other_rpm_changes recognition is broken on sles11sp3
SLES11SP3 has, unless all other supported distributions, only eight possible rpm changes instead of nine.

thardeck:
Fixes #688

### DIFF
--- a/lib/changed_rpm_files_helper.rb
+++ b/lib/changed_rpm_files_helper.rb
@@ -52,7 +52,7 @@ module ChangedRpmFilesHelper
       changes << "user" if expected_tag?("U", 5)
       changes << "group" if expected_tag?("G", 6)
       changes << "time" if expected_tag?("T", 7)
-      changes << "capabilities" if expected_tag?("P", 8)
+      changes << "capabilities" if @rpm_changes.size > 8 && expected_tag?("P", 8)
     end
 
     if @unknown_tag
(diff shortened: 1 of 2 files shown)


tgoettlicher:
At the review meeting there was a little of confusion figuring out the difference between:
https://github.com/SUSE/machinery/blob/fix_688/spec/unit/changed_rpm_files_helper_spec.rb#L63
and
https://github.com/SUSE/machinery/blob/fix_688/spec/unit/changed_rpm_files_helper_spec.rb#L71

You might to improve this with this commit as it covers the same area.

lib/changed_rpm_files_helper.rb
```diff
@@ -52,7 +52,7 @@ def parse_rpm_changes_line(line)
       changes << "user" if expected_tag?("U", 5)
       changes << "group" if expected_tag?("G", 6)
       changes << "time" if expected_tag?("T", 7)
-      changes << "capabilities" if expected_tag?("P", 8)
+      changes << "capabilities" if @rpm_changes.size > 8 &&  expected_tag?(["P", "C"], 8)
```
tgoettlicher: According to the rpm man page 'P' means 'capabilities differ'. Doesn't 'C' mean "SELinux context differs'?

tgoettlicher:
LGTM

PR #692, opened 2015-03-11
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150311_implement_global_exclude_option'

aduffeck:
  * 5256627 Fix hound issues
  * dd78768 DRY up FilterOptionParser
  * 5ccb269 Clean up FilterOptionParser class
  * 885d463 Add support for filter definition files in the global --exclude option
  * 5c16872 Use given_filesystem for creating example files
  * b2a4321 Add changelog entry for global --exclude option
  * 98ed7e3 Implement global --exclude option for applying filters globally
  * 5ac3f44 Handle filter definitions from the global "exclude" option
  * 0e8cd3c Extract filter option handling into a separate FilterOptionParser class
  * 53bc119 Add a "experimental-features" option (disabled by default)
  * 6e12a85 Add support for configuration options with '-'s


aduffeck:
I currently can't reproduce the unit test failure locally, will investigate.

*Note*: This shouldn't be merged to master currently because a release is pending.

PR #701, opened 2015-03-13
Fix warning that --skip-files is only available in for unmanaged-files

### LINKED ISSUE #699: no warning for scope os that --skip-files is not supported
The card for the feature states:

'Add warning for files scopes where the `--skip-files` switch isn't supported yet'

however:

-------------------skip---------------------
./bin/machinery inspect --skip-files "/tmp" -s os 192.168.121.244
Inspecting 192.168.121.244 for os...
Warning: System Description validation errors:
In scope unmanaged_files: The property (files) of type Array matched more than one of the required schemas.
Inspecting os...
 -> Found operating system 'SUSE Linux Enterprise Server 12' version '12'.

Hint: To show the data of the system you just inspected run:
./bin/machinery show 192.168.121.244

Hint: To do a full inspection containing all scopes and to extract files run:
./bin/machinery inspect 192.168.121.244 --name 192.168.121.244 --extract-files
-------------------skip---------------------

The expected warninng seems to be missing.

aduffeck:
Fixes #699

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -455,6 +455,11 @@ class Cli
         inspect_options[:extract_unmanaged_files] = true
       end
 
+      if options["skip-files"] && !(scope_list & ["config_files", "changed_managed_files"]).empty?
+        Machinery::Ui.warn("Warning: The --skip-files option is currently only supported for the " \
+          "\"unmanaged-files\" scope")
+      end
+
       filter = prepare_filter("inspect", options)
 
       inspector_task.inspect_system(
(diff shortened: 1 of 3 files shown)


tgoettlicher:
Will there be a warning shown when a non-files scope (e.g. `os`) is specified and the ``--skip-files`` option is passed?

aduffeck:
No. We also don't show a warning when `--extract-files` is used with a non-files inspector, for eample. I think the risk that a user expects that the `os` inspector would skip files with that option is pretty small :)

tgoettlicher:
Yes, that makes sense. lgtm

PR #702, opened 2015-03-13
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150311_implement_global_exclude_option'

aduffeck:
  * 87be26b Fix problem with filtering wildcard matches
  * db927d9 Ask the inspector for the summary only after filtering
  * 9140770 Apply filters when they are set using #set_filter
  * ba82814 Add Filter#apply! which applies its filters to a system description
  * eacf0f2 Implement array matchers in ElementFilters
  * 02bb85d Fix hound issues
  * 792b3f4 Clean up FilterOptionParser class
  * 64c91d2 Add support for filter definition files in the global --exclude option
  * d26b31d Use given_filesystem for creating example files
  * 3bb0024 Add changelog entry for global --exclude option
  * 6c9f54d Implement global --exclude option for applying filters globally
  * 3382b47 Handle filter definitions from the global "exclude" option
  * f5c6c49 Extract filter option handling into a separate FilterOptionParser class
  * 1ca23f4 Add a "experimental-features" option (disabled by default)
  * ec62110 Add support for configuration options with dashes in the name


lib/cli.rb
```diff
@@ -25,6 +25,9 @@ class Cli
   switch :version, negatable: false, desc: "Show version"
   switch :debug, negatable: false, desc: "Enable debug mode"
   switch [:help, :h], negatable: false, desc: "Show help"
+  if Machinery::Config.new.experimental_features
+    flag :exclude, negatable: false, desc: "Exclude elements matching the filter criteria"
```
tgoettlicher: What does "filter criteria" mean in this context? Woudn't following description explain it better?
"Exclude experimantal features (see: http://link-to-wiki-page )"

lib/config.rb
```diff
@@ -25,9 +25,13 @@ def define_entries
       default_config_file(Machinery::DEFAULT_CONFIG_FILE)
 
       entry("hints",
-        default:     true,
+        default: true,
         description: "Show hints about usage of Machinery in the context of the commands ran by the user"
       )
+      entry("experimental-features",
+        default: false,
+        description: "Enable experimental features. Use at your own risk."
```
tgoettlicher: Would a link to the wiki page make sense here? It's not that I'm requesting such a link, but I wonder if that could increase usability.

spec/unit/config_spec.rb
```diff
@@ -21,6 +21,18 @@
   include FakeFS::SpecHelpers
   subject { Machinery::Config.new }
 
+  it "works with keys with '-'s" do
```
tgoettlicher: I had to read it twice to get it. How about:
`it "works with keys containing hyphens ('-')" do`

cornelius:
This pull request is hard to review because it mixes quite a number of different things. It would probably speed up the review, if we could separate it in different chunks. Would it be possible to split it up in the following pieces:

* Refactoring (e.g. the extraction of the `FilterOptionParser` class)
* Implementation of the experimental features switch (this could include the `--exclude` option as a stub)
* Implementation of the `Filter.apply!` method.
* Implementation of the `--exclude` option.

The first two can probably quickly be reviewed and merged. The third one might need a little bit of discussion, but could also be quickly merged. The fourth one will need a bit of testing, which might take a bit longer.

aduffeck:
I extracted https://github.com/SUSE/machinery/pull/703 and https://github.com/SUSE/machinery/pull/704 from this pull request. It would be good if those could be reviewed and merged into master since the rest of this pull request partly depends on those changes and can't be easily extracted (at least against master).

PR #703, opened 2015-03-13
[Review] Request from 'aduffeck' @ 'SUSE/machinery/implement_filter_apply'

aduffeck:
  * 84b1395 Implement array matchers in ElementFilters
  * 2f991d0 Add Filter#apply! which applies its filters to a system description


spec/unit/filter_spec.rb
```diff
@@ -122,4 +122,57 @@
...
+      expect(description.unmanaged_files.files.map(&:name)).to match_array([
+        "/etc/another-unmanaged-file"
+      ])
+    end
+
+    it "filters by array matches" do
+      description = create_test_description(scopes: ["changed_managed_files"])
+
+      expect(description.changed_managed_files.files.map(&:name)).to match_array([
+        "/etc/deleted changed managed",
```
cornelius: Would it be possible to express that in a way which resembles the system description a bit closer? It's hard to understand the test without context.

spec/unit/filter_spec.rb
```diff
@@ -122,4 +122,57 @@
...
+        "/etc/cron.daily/logrotate"
+      ])
+
+      filter = Filter.new("/changed_managed_files/files/name=/etc/c*")
+      filter.apply!(description)
+
+      expect(description.changed_managed_files.files.map(&:name)).to match_array([
+        "/etc/deleted changed managed"
+      ])
+    end
```
cornelius: Could it make sense to dry up these tests a bit and instead of repeating the code have a table of something to define the different test cases?

Something like:

```
{
  before: [
    "/etc/deleted changed managed",
    "/etc/cron.daily/cleanup",
    "/etc/cron.daily/logrotate"
  ],
  filter: "/changed_managed_files/files/name=/etc/c*",
  after: [
    "/etc/deleted changed managed"
  ]
}

cornelius:
agree with tim, otherwise looks good to me.

thardeck:
lgtm besides my comment

aduffeck:
Merged

PR #706, opened 2015-03-13
Use FakeFS for hint spec

thardeck:
Prevent the hint test from reading the actual machinery config.


aduffeck:
I think we're trying to get away from FakeFS where possible, so would it be possible to use something like given_filesystem here as well?

PR #707, opened 2015-03-13
[Review] Request from 'aduffeck' @ 'SUSE/machinery/global_exclude_option'

aduffeck:
  * 67c3684 Add changelog entry about global --exclude option
  * dae641e --skip-files works for all file inspectors now
  * 83ab784 Ask the inspector for the summary only after filtering
  * cff4e72 Add global --exclude option
  * 0c712b8 Apply filters when they are set on a SystemDescription
  * 2119b75 Improve descriptiong of the "experimental-features" option

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Add experimental `--exclude` option to exclude elements from inspection.
+  Run `machinery config experimental-features on` to enable the option
 
 ## Version 1.5.0 - Fri Mar 13 13:03:47 CET 2015 - thardeck@suse.de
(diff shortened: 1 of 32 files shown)


greygoo:
lgtm but maybe someone else can have a look as it is quite a big change and I might have overseen something.

tgoettlicher:
Using @config for getting configuration values is implemented and .parse is used instead of .to_filter which makes the code more understandable, now. Furthermore aduffeck prommised to discuss the other optimizations with cornelius when he's back in the office. In order to not block further development, this pull request looks good to me.

PR #709, opened 2015-03-13
Do not read real machinery config during test

aduffeck:


aduffeck:
RBR branch for #706 which doesn't rely on FakeFS

thardeck:
Thx, but I am not sure if this is better. It is a lot of more stubbing and changing so I leave both there until someone else decides.

aduffeck:
Not sure what you mean, it's the same amount of stubbing as before and as
in your branch, just on a higher level where it makes more sense. It's a
cleaner fix than resetting the config by throwing FakeFS at it although the
actual data is still being stubbed imo.
Am 13.03.2015 16:52 schrieb "Tim Hardeck" <notifications@github.com>:

thardeck:
lgtm

PR #711, opened 2015-03-16
Adapt to new buildservice ruby environment

aduffeck:
It no longer provides the binary at /usr/bin/machinery at build time,
but only at /usr/bin/machinery2.0


thardeck:
lgtm

PR #715, opened 2015-03-16
Add hint for upgrade-format --force

vwallfahrer:


aduffeck:
Extracting the hint information in the error handler is complicated and error prone imo, my counter proposal is #720

PR #717, opened 2015-03-16
Upgrade system description

vwallfahrer:


aduffeck:
I think you only wanted to submit the  5af22fd commit, right?

The code looks good to me except for the hound issues and the failing test.

PR #718, opened 2015-03-16
Add inspect verbose option to list used filters

thardeck:
I was trying to use `captured_machinery_output` but it was always empty besides initializing it in the describe block. I am not sure why but the tests work this way too.


aduffeck:
Using the `capture_machinery_output` helper worked for me: #721

PR #722, opened 2015-03-17
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150317_add_verbose_option_to_show'

aduffeck:
  * 06ad287 Add "--verbose" option to "show" command to display used show filters
  * 1efe606 Do not choke on non-existing elements when applying filters

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -545,7 +545,7 @@ class Cli
     c.switch "html", required: false, negatable: false,
       desc: "Open system description in HTML format in your web browser."
     c.switch "verbose", required: false, negatable: false,
-      desc: "Show the list of filters that is applied before showing the description."
+      desc: "Show the filters that were applied before showing the description."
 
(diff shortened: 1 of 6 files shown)


lib/cli.rb
```diff
@@ -544,6 +544,8 @@ def self.parse_scopes(scope_string)
       desc: "Show diffs of configuration files changes."
     c.switch "html", required: false, negatable: false,
       desc: "Open system description in HTML format in your web browser."
+    c.switch "verbose", required: false, negatable: false,
+      desc: "Show the list of filters that is applied before showing the description."
```
thardeck: I would rephrase it to get rid of the singular plural confusion and to keep the same tense as in all other descriptions.
`Show the filters that were applied ...`

thardeck:
lgtm

aduffeck:
Merged.

PR #723, opened 2015-03-17
Add inspect verbose option to list used filters

thardeck:
Supersedes #718 and #721

### DIFF
--- a/man/machinery-inspect.1.md
+++ b/man/machinery-inspect.1.md
@@ -68,6 +68,10 @@ trigger errors.
 
       $ `machinery` inspect --skip-files=/file\\,with_comma myhost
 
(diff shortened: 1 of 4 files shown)


man/machinery-inspect.1.md
```diff
@@ -68,6 +68,10 @@ trigger errors.
 
       $ `machinery` inspect --skip-files=/file\\,with_comma myhost
 
+  * `--verbose` (optional):
+    Display the filters which are used during inspection at the beginning.
```
aduffeck: I find the `at the beginning` a bit confusing. Can we just remove it here and in the man page?

aduffeck:
lgtm

PR #720, opened 2015-03-17
[Review] Request from 'aduffeck' @ 'SUSE/machinery/rbr_show_hint_for_upgrade_format_force'

aduffeck:
Review-By-Refactoring branch for #715

PR #726, opened 2015-03-17
Show hint for upgrade format force 

vwallfahrer:
Rebased #720


lib/upgrade_format_task.rb
```diff
@@ -44,7 +44,9 @@ def upgrade(store, name, options = {})
 
     if !errors.empty?
       Machinery.logger.error errors.join("\n")
-      raise Machinery::Errors::UpgradeFailed.new(errors.join("\n"))
+      exception = Machinery::Errors::UpgradeFailed.new(errors.join("\n"))
+      exception.hint_proc = -> { Hint.upgrade_format_force(name: name || "--all") }
```
cornelius: It needs some effort to understand this code and you need to know what the `hint_proc` method is supposed to do. Wouldn't it be easier to not introduce special magic here and use something like:

```
exception = Machinery::Errors::UpgradeFailed.new(errors.join("\n") + Hint.upgrade_format_force(name: name || "--all", to_string: true)
```

where the `to_string` option would get the hint returned as string instead of being directly put out on STDERR.

aduffeck: I don't really like changing the behavior of the hint methods that fundamentally depending on an option. What about

`exception = Machinery::Errors::UpgradeFailed.new(errors.join("\n") + Hint.to_string(:upgrade_format_force, name: name)`

instead?

cornelius: Looks good to me as well. It creates a slight asymmetry, though. To be consequent, we should then express printing a hint as something like

```ruby
Hint.print(:upgrade_format_force, name: name)
```

aduffeck: :+1:

PR #727, opened 2015-03-17
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150317_refactor_inspector_interface'

aduffeck:
  * 1f0de3b Give system and description to the Inspector constructor already
  * d3a1a88 Make Inspector.for return classes instead of instances

### DIFF
--- a/lib/inspect_task.rb
+++ b/lib/inspect_task.rb
@@ -89,10 +89,11 @@ class InspectTask
       filter_in_metadata = Filter.new
     end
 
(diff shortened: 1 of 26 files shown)


cornelius:
I like this change. It makes the code a good bit cleaner.

The first commit can't work independent of the second one. I'm fine with that as we apply it in one go, but it still creates a point in the history, where the code is broken. I think we should avoid that in the future.

PR #731, opened 2015-03-17
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150317_refactor_systemdescription_set_filter'

aduffeck:
  * 9c6e3a8 Rename SystemDescription#set_filter,take a String array instead of Filter
  * d008b58 Don't apply the filter in SystemDescription#set_filter


cornelius:
Looks good to me.

PR #733, opened 2015-03-17
Fix rpm building

thardeck:
Fixing the original issue that caused us to create the commit
dd4761784697aadfc7f7acc66cfcae85a33923fb
    
Building machinery on 13.1 failed because the ruby macros were changed
recently and an issue in the 13.1 /etc/gemrc resulted in always renaming
machinery to machinery2.0 which normally should only happen for symlink
builds.


tgoettlicher:
After getting the change explained face to face it looks good to me.

PR #748, opened 2015-03-19
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150319_only_show_used_filters_when_there_are_filters_being_applied'

aduffeck:
  * 77b03e6 Only show used filters when there are filters being applied
  * e0b09a2 Implement Filter#empty? method


vwallfahrer:
lgtm

PR #750, opened 2015-03-19
Sort descriptions before upgrading

vwallfahrer:
Upgrading path succession was non deterministic which resulted in test failure.


aduffeck:
We also have to manually sort in `ListTask`, so should we maybe already sort in `SystemDescriptionStore#list` instead?

PR #757, opened 2015-03-23
Upgrade to RSpec 3.2

aduffeck:
See http://rspec.info/blog/2015/02/rspec-3-2-has-been-released/ for the
changelog.


cornelius:
I'm fine with upgrading to RSpec 3.2. I trust that the tests run when it actually is done.

PR #758, opened 2015-03-23
Using generate-html creates an output now

vwallfahrer:
Get an output where the generated html file is stored.


spec/unit/generate_html_task_spec.rb
```diff
@@ -30,5 +32,12 @@
 
       subject.generate(description)
     end
+
+    it "shows an output where the html file is stored" do
+      subject.generate(description)
+      expect(captured_machinery_output).to match(
+        "The generated HTML file is stored in: \n"
```
cornelius: The more important part would be to check that the right file name is printed.

PR #761, opened 2015-03-24
Run all local commands with C locale to prevent language issues

### LINKED ISSUE #689: Tarball library is language dependent
When running the unit test on a non-english system the tarball test fails. The reason for this is, that we parse the output of the `tar` command. 

       - {:path=>"hardlink",
       + {:path=>"hardlink Verknüpfung zu file\n",

aduffeck:
Fixes #689


aduffeck:
Please ignore the failing integration test, it is caused by some left-overs from #760 which weren't cleaned up because pennyworth did not have the needed functionality yet.

cornelius:
Looks good to me.

PR #769, opened 2015-03-26
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150326_implement_negated_filters'

aduffeck:
  * c9a070f Adapt Filtering-Design document to reflect the inverted filters changes
  * fbf4a35 Add changelog entry for negated filter expressions in --exclude
  * 0ecffdf Consider ElementFilter operators in Filter#to_array
  * 56265eb Remove superfluous linebreaks
  * 72c69e6 Add test for defition parser with '!=' operator
  * 629790a Implement :!= operator in ElementFilters
  * cb204bc ElementFilters have an operator now (always :== currently)


lib/element_filter.rb
```diff
@@ -16,13 +16,15 @@
...
 
-  def initialize(path, matchers = nil)
+  def initialize(path, operator, matchers = nil)
     @path = path
+    @operator = operator
     @matchers = []
 
-    raise("Wrong type") if ![NilClass, String, Array].include?(matchers.class)
+    raise("Wrong filter type") if ![NilClass, String, Array].include?(matchers.class)
+    raise("Wrong filter operator '#{@operator}'") if !["!=", "="].include?(@operator)
```
cornelius: Maybe it would be good to introduce constants for the operators, so that they don't have to be literally repeated in the code.

lib/filter.rb
```diff
@@ -49,9 +49,9 @@ class Filter
   def self.parse_filter_definitions(filter_definitions)
     element_filters = {}
     Array(filter_definitions).each do |definition|
-      path, matcher_definition = definition.split("=", 2)
+      path, operator, matcher_definition = definition.scan(/(.*?)(=|!=)(.*)/)[0]
```
cornelius: A filter definition like `/path>=matcher` will silently have no effect as a path `/path>` can't exist in the data. Maybe it would be good to catch these kind of cases, so that users trying different operators get proper feedback.

lib/filter.rb
```diff
@@ -89,7 +89,7 @@ def add_element_filter_from_definition(filter_definition)
     new_element_filters = Filter.parse_filter_definitions(filter_definition)
 
     new_element_filters.each do |path, element_filter|
-      @element_filters[path] ||= ElementFilter.new(path)
+      @element_filters[path] ||= ElementFilter.new(path, element_filter.operator)
       @element_filters[path].add_matchers(element_filter.matchers)
```
cornelius: This will fail, when there are element filters for the same path with different operators.

PR #771, opened 2015-03-26
Get rid of Ui.print_output, Ui.puts supports paging transparently now

aduffeck:
Machinery::Ui.print_output was used to print output either on the
console (when no pager was available or when the user gave the
--no-pager option) or a pager. That approach was not very flexible,
though, because the whole output had to be given as a string at once.

The new approach now uses the normal Machinery::Ui.puts method. It
writes to the console by default, but a pager can be enabled using

  Machinery::Ui.use_pager = true

That way the output can be composed from several places and can be
streamed to the user instead of having to wait for the whole output
before showing it.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -51,6 +51,7 @@ class Cli
 
       Hint.print(:get_started)
     end
+    Machinery::Ui.close_pager
   end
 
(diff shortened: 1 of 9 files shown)


cornelius:
lgtm

PR #773, opened 2015-03-27
Show filters which were applied during inspection

vwallfahrer:


lib/cli.rb
```diff
@@ -571,9 +571,18 @@ def self.parse_scopes(scope_string)
...
       filter = FilterOptionParser.parse("show", options, global_options)
 
-      if options["verbose"] && !filter.empty?
-        Machinery::Ui.puts "\nThe following filters were applied before showing the description:"
-        Machinery::Ui.puts filter.to_array.join("\n") + "\n\n"
+      inspected_filters = description.filter_definitions("inspect")
+
+      if options["verbose"]
+        if !inspected_filters.empty?
+          Machinery::Ui.puts "\nThe following filters were applied during inspectation:"
```
cornelius: It should be "inspection", not "inspectation".

lib/cli.rb
```diff
@@ -571,9 +571,18 @@ def self.parse_scopes(scope_string)
...
+          Machinery::Ui.puts inspected_filters.join("\n") + "\n\n"
+        end
+
+        if !filter.empty?
+          Machinery::Ui.puts "\nThe following filters were applied before showing the description:"
+          Machinery::Ui.puts filter.to_array.join("\n") + "\n\n"
+        end
```
cornelius: Wasn't this already added somewhere? We had a card for that during the last sprint.

vwallfahrer: Yes, but I had to add another if-statement. There is only an output, if the inspected_filter or/and the filter array are not empty.

PR #776, opened 2015-03-30
Fix ugly error messages in RSpec "after" hooks when cleaning up

aduffeck:
We now use the new Runner#cleanup_directory method in pennyworth which
cleans up the directory on runners working with real machines, but does
not run into error when working with VMs.

Depends on https://github.com/SUSE/pennyworth/pull/59


cornelius:
This looks nice now. :shipit: when the depending pull request is in.

PR #777, opened 2015-03-30
Document that --skip-files does not expand paths

### LINKED ISSUE #695: --skip-files does not handle ../ in directory name
Following line does not skip the /opt directory:

```
# machinery inspect hostname -x  --scope unmanaged-files --skip-files=/etc/../opt
```

This is a minor bug. However, specifying files like this is valid in UNIX/Linux. This might occure for example when scripts call machinery.

vwallfahrer:
Fix #695


thardeck:
Since bold is used for the other Note in this man page part I would format it the same way.

man/machinery-inspect.1.md
```diff
@@ -68,6 +68,8 @@ trigger errors.
 
       $ `machinery` inspect --skip-files=/file\\,with_comma myhost
 
+    Note: Paths are not expanded.
```
cornelius: We should make it clear that this is only true for the file or directory names which are given as argument, not if a file is given with the `@` prefix. So maybe saying "file or directory names" instead of paths (as it is done in the documentation in other places as well) would be better. And maybe it would make it more clear what is meant by providing an example (e.g. `../path`).

PR #778, opened 2015-03-30

aduffeck:
  * 7cb2028 Show filters used during inspection in HTML view
  * 9e11063 Merge pull request #774 from SUSE/show_inspecting_filters_2
  * 30c82cf Refactored Code: correct spelling mistake
  * 07f67b3 Show filters which were applied during inspect

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -15,6 +15,7 @@
   user a hint that the option --force can be used.
 * Add option to show command to display filters used during inspection
 * Support negated filter expressions in --exclude option
+* Show filters used during inspection in HTML view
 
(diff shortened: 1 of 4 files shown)


cornelius:
It's not obvious that the "(inspection details)" text is hoverable. Not sure, if people will find it easily. Also the popup can be a bit confusing, because it disappears quickly when the mouse is moved to the wrong place.

Maybe it would be better to make the "inspection details" a proper link and only show the popup when the link is actually clicked. It should then get a "close" icon to close it again.

PR #780, opened 2015-03-30
Document that --skip-files does not expand paths

### LINKED ISSUE #695: --skip-files does not handle ../ in directory name

vwallfahrer:
Fix #695
Supersedes #779


thardeck:
lgtm

Issue #781, opened 2015-03-31
Feedback on clickable elements in HTML view


cornelius:
At the moment there is no consistent feedback on which element are clickable in the HTML view:

Links turn the mouse pointer in a hand. This is useful. It would be good to use this consistently for clickable elements. The following elements don't have this feedback right now:

* [x] Reset button
* [x] Collapse/Expand control
* [x] "Inspection details" toggle (from #782)

The icon on the left indicating the scope changes the mouse pointer, but clicking it does not always do something, and when it does it only moves up the content to the top of the window. It also has a hover popup window. Maybe it would be better to not make it clickable.

PR #782, opened 2015-04-01

aduffeck:
  * cf2e301 Close inspection details popup using a close button
  * 7219d65 Fix hound issues
  * c91f5f2 Show filters used during inspection in HTML view

### DIFF
--- a/html/index.html.haml
+++ b/html/index.html.haml
@@ -12,6 +12,9 @@
     %script{:src => "assets/collapse.js"}
 
     %script#content{:type => "text/html"}
+      .hidden
+        {{>inspection_details_template}}
+
       {{>os}}
       {{>packages}}
       {{>patterns}}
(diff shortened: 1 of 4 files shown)


html/index.html.haml
```diff
@@ -23,6 +26,31 @@
       {{>config_files}}
       {{>services}}
 
+    %script#inspection_details_template{:type => "text/html"}
+      #inspection_details
+        #filters
+          .row
+            .col-xs-1
+            .col-xs-11
+              %h3 Filters used during Inspection
```
cornelius: I would lower-case "inspection" here.

cornelius:
I think it works better now. There is an issue with discoverability of the control (see #781), but this could treated independently as well. It's also a bit annoying that you have to scroll to see all the content, even, if there would be enough space on the screen to show all at once. But this also could be fixed later.

PR #784, opened 2015-04-01

### LINKED ISSUE #781: Feedback on clickable elements in HTML view

aduffeck:
  * cfbb099 Fix usability issues regarding links

Fixes #781

### DIFF
--- a/html/index.html.haml
+++ b/html/index.html.haml
@@ -57,8 +57,7 @@
       #os_container.scope
         .row
           .col-xs-1
-            %a{:href => "#os"}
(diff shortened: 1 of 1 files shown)


thardeck:
lgtm

PR #786, opened 2015-04-01
Package for Red Hat to make unmanaged files test work on SUSE and Redhat

tgoettlicher:


greygoo:
lgtm, was tested for rhel5 integration tests and passed

PR #788, opened 2015-04-02
[Review] Request from 'aduffeck' @ 'SUSE/machinery/compare_html'

aduffeck:
  * d462ac5 Add summary information and links to the "common" sections
  * 1de79c2 Add headers for the different tables describing the origin of the data
  * bdf705c Optimize for smaller screen sizes
  * 0d57df5 Implement --html option for compare command


lib/html.rb
```diff
@@ -47,6 +47,24 @@ def self.generate(description)
     target
   end
 
+  def self.generate_comparison(diff, target)
+    FileUtils.mkdir_p(File.join(target, "assets"))
+    template = Haml::Engine.new(
+      File.read(File.join(Machinery::ROOT, "html", "comparison_index.html.haml"))
```
cornelius: What about naming it just `comparison.html.haml`?

html/assets/compare/machinery.js
```diff
@@ -0,0 +1,69 @@
+$(document).ready(function () {
+  // Render the system description
```
cornelius: This should be "render the comparison", shouldn't it?

html/comparison_index.html.haml
```diff
@@ -0,0 +1,1010 @@
...
+                %span.summary-part {{meta.description_a}}: {{packages.0.length}} packages
+                {{/packages.0.length}}
+                {{#packages.1.length}}
+                %span.summary-part {{meta.description_b}}: {{packages.1.length}} packages
+                {{/packages.1.length}}
+                {{#packages.2.length}}
+                %span.summary-part
+                  %a{href: "#packages_both"}
+                    both
+                  = ": {{packages.2.length}} patterns"
```
cornelius: This should be "packages" not "patterns".

cornelius:
I would suggest to fix the small issues I commented on, and then merge the code as experimental feature. This way we can do QA and gather more feedback without blocking releases, and we still can refine the view before we make it generally available.

PR #792, opened 2015-04-02
Improve format upgrade output 

vwallfahrer:
Fix  #789


spec/unit/upgrade_format_task_spec.rb
```diff
@@ -87,22 +85,28 @@ def migrate; end
       end
     end
 
+    it "lists each system description and its status during upgrade" do
+      expected_output = <<-EOF
+Reading "description1" ... successfully upgraded from version 1 to #{SystemDescription::CURRENT_FORMAT_VERSION}.
+Reading "description2" ... no upgrade necessary.
+Upgraded 1 system descriptions.
```
cornelius: This should be singular: "Upgraded 1 system description."

lib/migration.rb
```diff
@@ -88,8 +88,7 @@ def migrate_description(store, description_name, options = {})
       end
 
       if current_version == SystemDescription::CURRENT_FORMAT_VERSION
-        Machinery::Ui.puts "System description \"#{description_name}\" is up to " \
-          "date, no upgrade necessary."
+        Machinery::Ui.puts "no upgrade necessary."
```
tgoettlicher: As this is a sentence, I'd write 'No' with an upper case 'N'.

lib/upgrade_format_task.rb
```diff
@@ -51,13 +57,12 @@ def upgrade(store, name, options = {})
     end
 
     if options[:all]
+      migrations_done == 1 ? frequency = "description" : frequency = "descriptions"
       if migrations_done > 0
-        Machinery::Ui.puts "Upgraded #{migrations_done} system descriptions successfully."
+        Machinery::Ui.puts "Upgraded #{migrations_done} system #{frequency}."
```
tgoettlicher: The variable name frequency is misleading. I'd expect it to contain a number.

tgoettlicher: Would the following construct be easier to read?
```
if migrations_done == 1
  Machinery::Ui.puts "Upgraded 1 system description."
elsif migrations_done > 1
  Machinery::Ui.puts "Upgraded #{migrations_done} system descriptions."
else
  Machinery::Ui.puts "No system descriptions were upgraded."
end
```

spec/unit/migration_spec.rb
```diff
@@ -101,7 +101,7 @@ class MigrateFoo < Migration; end
       Migration.migrate_description(store, "v3_description")
       description = SystemDescription.load("v3_description", store)
       expect(description.format_version).to eq(SystemDescription::CURRENT_FORMAT_VERSION)
-      expect(captured_machinery_output).to include("is up to date, no upgrade necessary")
+      expect(captured_machinery_output).to include("no upgrade necessary")
```
tgoettlicher: If you change the message to begin with an uppercase letter you need to adapt this line, too.

tgoettlicher:
There is a recommendation by Cornelius how to pluralize. See: https://github.com/SUSE/machinery/pull/792#commitcomment-10548080

lib/upgrade_format_task.rb
```diff
@@ -52,12 +58,10 @@ def upgrade(store, name, options = {})
 
     if options[:all]
       if migrations_done > 0
-        Machinery::Ui.puts "Upgraded #{migrations_done} system descriptions successfully."
-      else
+        Machinery::Ui.puts "Upgraded #{migrations_done} #{"system description".pluralize(migrations_done)}."
+      elsif migrations_done = 0
```
cornelius: This is using the assignment operator `=`, not the comparison operator `==`, so it actually doesn't check the condition but sets `migrations_done` to 0. This is not what is supposed to happen, although it doesn't have any effect here.

Actually I would leave out the condition and just leave the plain `else`. This way the if statement covers all cases.

spec/unit/list_task_spec.rb
```diff
@@ -65,11 +65,11 @@
     it "lists the system descriptions with scopes" do
       system_description.save
       expected_output = <<-EOF
-
  foo:
    * packages
    * repositories
 
+
```
cornelius: Is this intentional that there are two blank lines here now?

spec/unit/ui_spec.rb
```diff
@@ -57,7 +65,7 @@
       allow($stdout).to receive(:tty?).and_return(true)
       allow(LocalSystem).to receive(:validate_existence_of_package).
         and_raise(Machinery::Errors::MissingRequirement)
-      expect($stdout).to receive(:puts).with(output)
+      expect($stdout).to receive(:print).with(output + "\n")
```
cornelius: This test is not ideal because it tests the implementation, not the result. You can tell that because you needed to change the test, although the behavior of the method hasn't changed. It would be better to write the test in a way that is more independent of the implementation.

I would be ok with leaving that to a later refactoring, though.

spec/unit/upgrade_format_task_spec.rb
```diff
@@ -87,22 +85,28 @@ def migrate; end
       end
     end
 
+    it "lists each system description and its status during upgrade" do
+      expected_output = <<-EOF
+Reading "description1" ... successfully upgraded from version 1 to #{SystemDescription::CURRENT_FORMAT_VERSION}.
```
cornelius: I think we are usually using single quotes around names in these kind of messages. Could you check, if that is true, and make sure that the patch is consistent with the rest of the code?

spec/unit/upgrade_format_task_spec.rb
```diff
@@ -87,22 +85,28 @@ def migrate; end
...
+Reading "description2" ... No upgrade necessary.
```
cornelius: There is some inconsistency in capitalization. In the first part of the message it says "... successfully" in the second part "... No". It should either both be upper case or both be lower case. I think I would prefer upper case, as it actually is two sentences.

PR #793, opened 2015-04-02
Show expected variable type in config error

### LINKED ISSUE #362: When invalid config value is given, show what is valid
`machinery config hints 42` shows an error message, that 42 is not a valid value. The message should also show, what values are valid.

keberle94:
Fixes #362


tgoettlicher:
LGTM

PR #798, opened 2015-04-09
Rename ScopeMixin to Scope

cornelius:
It is an implementation detail that the Scope is a mixin. For the
clarity of the API it is more straight forward to just name it
Scope.

This also gives more flexibility towards future change in case it
would make sense at some time to turn Scope into a class.


aduffeck:
lgtm

PR #803, opened 2015-04-20
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150420_restructure_plugins_directory'

aduffeck:
  * bab6e06 Move schema definitions to the according plugin directories
  * 798d7d2 Restructure plugins into one directory per plugin
  * 9ddf75f Merge branch 'scope_renaming'
  * b81f8a9 Rename ScopeMixin to Scope


cornelius:
There still is a reference in `editing/machinery.sh` to the old plugin structure.

Other than that the patch looks fine to me.

PR #805, opened 2015-04-20
Fixed config file inspector test

greygoo:
Change config file inspector tests by testing lower methods first and then methods building on top of them with stubbing the lower methods instead of the called system commands.


cornelius:
We can close this pull request, right?

PR #807, opened 2015-04-20
Fixed config file inspector test

keberle94:
superseeds #805


spec/unit/config_files_inspector_spec.rb
```diff
@@ -96,231 +118,349 @@
...
+          extracted: false,
+          files: ConfigFileList.new
+        )
+        expect(description["config_files"]).to eq(expected)
+      end
 
-    it "returns sorted data" do
-      system = double
-      expect_inspect_configfiles(system, true)
+      it "raise an error when requirements are not fulfilled" do
```
cornelius: Typo, it should be "raises".

spec/unit/config_files_inspector_spec.rb
```diff
@@ -96,231 +118,349 @@
...
-      inspector.inspect(filter, :extract_changed_config_files => true)
-      names = description["config_files"].files.map(&:name)
+        inspector = ConfigFilesInspector.new(system, description)
+        expect { inspector.inspect(filter) }.to raise_error(
+          Machinery::Errors::MissingRequirement
+        )
+      end
+
+      it "extracts changed configuration files" do
+        cfdir = File.join(store.description_path(name), "config_files")
```
cornelius: `cfdir` is not a very expressive name. Better not use abbreviations in variable names. This makes the code easier to read.

spec/unit/config_files_inspector_spec.rb
```diff
@@ -96,231 +118,349 @@
...
+        expect(system).to receive(:retrieve_files).with(
+          config_paths,
+          cfdir
+        )
+        expect_data_gather_cmds(
```
cornelius: `expect_data_gather_cmds` is a bit of a confusing name. It stubs the call to the `stat` command. So maybe better rename it to something which says this more explicitly.

Why is this needed here, anyway? It makes the test a bit asymmetric, because it stubs the `retrieve_files` call on the `System` level, but lets through the call to the `stat` command.

keberle94: Thanks for the feedback.
Andre explained me in this case it is needed, because system is not a real system but a double and that both stubs are on the same level but retrieve_files is written in the text and expect_data_gather_cmds is extracted to the top of the file.

PR #808, opened 2015-04-21
Show an error, when filter doesn't contain an operator

### LINKED ISSUE #770: No error, when filter doesn't contain an operator
The command `machinery --exclude=abc show NAME` runs without warning or error. It should fail with an error, as `abc` is not a valid filter expression as it doesn't contain an operator.

vwallfahrer:
Fixed  #770


thardeck:
lgtm besides my comment

PR #809, opened 2015-04-21
Make empty array case explicit in the schema definition

### LINKED ISSUE #680: Validation error when 0 unmanaged files were inspected
A description which has an unmanaged files section with 0 files (can be created with `machinery inspect --skip-files=/*`) has validation errors:

```
In scope unmanaged_files: The property (files) of type Array matched more than one of the required schemas.
Warning: File validation errors:
Error validating description 'xyz'

Scope 'unmanaged_files':
  * File '/home/cs/.machinery/xyz/unmanaged_files/files.tgz' doesn't have meta data
```

aduffeck:
Fixes #680

### DIFF
--- a/plugins/schema/v3/system-description-unmanaged-files.schema.json
+++ b/plugins/schema/v3/system-description-unmanaged-files.schema.json
@@ -11,8 +11,13 @@
       "oneOf": [
         {
           "type": "array",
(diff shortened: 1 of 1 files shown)


thardeck:
lgtm

PR #811, opened 2015-04-21
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150420_restructure_plugins_directory'

aduffeck:
  * 1efee18 Remove dead code
  * 888dcf5 Move schema definitions to the according plugin directories
  * f959dbd Restructure plugins into one directory per plugin

Supersedes #803


cornelius:
Change looks fine. Prophet should also give his ok, though.

PR #814, opened 2015-04-22
Adding remote-user option to machinery.config

keberle94:


lib/cli.rb
```diff
@@ -432,6 +432,8 @@ def self.parse_scopes(scope_string)
     c.flag "skip-files", required: false, negatable: false,
       desc: "Do not consider given files or directories during inspection. " \
         "Either provide one file or directory name or a list of names separated by commas."
+    c.flag ["remote-user", :u], type: String, required: false, default_value: @config.remote_user,
+      desc: "Uses the following remote user to inspect system", arg_name: "USER"
```
cornelius: "uses the following" sounds weird. It probably should just be "remote user"

aduffeck:
LGTM besides the nitpicks, but it should only be merged when we also have implemented the according logic.

keberle94:
Close reviewed and accepted Pull-Request. -Waiting for implementation- Then re-open and merge before implementation.

PR #821, opened 2015-04-23
Support tumbleweed

thardeck:


boblmartens:
Thank you!

spec/unit/os_inspector_spec.rb
```diff
@@ -142,6 +142,21 @@
...
+    it "is able to recognize the rolling openSUSE release Tumbleweed" do
+      FakeFS::FileSystem.clone("spec/data/os/openSUSETumbleweed/etc/os-release",
+        "/etc/os-release")
+
+      expect(inspector).to receive(:get_arch).and_return("x86_64")
+
+      inspector.inspect(filter)
+
+      expect(description.os.name).to eq "openSUSE Tumbleweed"
+      expect(description.os.version).to include(" (Tumbleweed)")
```
aduffeck: Why don't we match the complete version string instead?

man/machinery-build.1.md
```diff
@@ -49,20 +49,6 @@ built image. The list of filters is shown at the beginning of the build.
   * All repositories in the system description must be accessible from the
     build machine on which `machinery build` is called.
 
-### BUILD SUPPORT MATRIX
```
aduffeck: Can't we keep this section? It is helpful to the users imo...

plugins/inspect/os_inspector.rb
```diff
@@ -92,6 +92,11 @@ def get_os_from_os_release
     end
     if result["pretty_name"]
       result["pretty_name"] = strip_arch_from_name(result["pretty_name"])
+
+      # remove always changing Tumbleweed version from pretty name
+      if result["pretty_name"].include?("Tumbleweed")
```
aduffeck: I'd prefer to have make the ckeck a bit more specific, e.g. 

```
if result["pretty_name"] =~ /^openSUSE.*Tumbleweed$/
```

or something like that.

PR #827, opened 2015-04-24
Move scrub method to library and make use of it in html rendering

tgoettlicher:
Please enter the commit message for your changes. Lines starting


thardeck:
I would add it to the machinery module (move it upwards) and write a test.

keberle94:
I'm working on this now.

PR #830, opened 2015-04-27
Escape '\' in javascript

keberle94:
Fixes bnc#925732


tgoettlicher:
As the list of characters which need to be escaped is getting longer and longer you could put it in a separate method to increase the readabilty.

keberle94:
Ok I'll try that. Thanks.

thardeck:
lgtm

Issue #831, opened 2015-04-27
xinetd services on RHEL5 can not be inspected


greygoo:
When running the services inspector on a RHEL5 system it failed when xinetd was installed on such system due to the changed output from chkconfig. This was filed under https://github.com/SUSE/machinery/issues/815. The proposed fix for https://github.com/SUSE/machinery/issues/815 is currently to skip xinetd services status and implement proper xinetd support. That is what this bug is for.

Output from a chkconfig when xinetd is installed:
```
-------------------------snip--------------------------
[root@dhcp86 ~]# chkconfig --list
NetworkManager 	0:off	1:off	2:off	3:off	4:off	5:off	6:off
acpid          	0:off	1:off	2:on	3:on	4:on	5:on	6:off
anacron        	0:off	1:off	2:on	3:on	4:on	5:on	6:off
atd            	0:off	1:off	2:off	3:on	4:on	5:on	6:off
auditd         	0:off	1:off	2:on	3:on	4:on	5:on	6:off
autofs         	0:off	1:off	2:off	3:on	4:on	5:on	6:off
avahi-daemon   	0:off	1:off	2:off	3:on	4:on	5:on	6:off
avahi-dnsconfd 	0:off	1:off	2:off	3:off	4:off	5:off	6:off
bluetooth      	0:off	1:off	2:on	3:on	4:on	5:on	6:off
capi           	0:off	1:off	2:off	3:off	4:off	5:off	6:off
conman         	0:off	1:off	2:off	3:off	4:off	5:off	6:off
cpuspeed       	0:off	1:on	2:on	3:on	4:on	5:on	6:off
crond          	0:off	1:off	2:on	3:on	4:on	5:on	6:off
dc_client      	0:off	1:off	2:off	3:off	4:off	5:off	6:off
dc_server      	0:off	1:off	2:off	3:off	4:off	5:off	6:off
dnsmasq        	0:off	1:off	2:off	3:off	4:off	5:off	6:off
dund           	0:off	1:off	2:off	3:off	4:off	5:off	6:off
firstboot      	0:off	1:off	2:off	3:on	4:off	5:on	6:off
gpm            	0:off	1:off	2:on	3:on	4:on	5:on	6:off
haldaemon      	0:off	1:off	2:off	3:on	4:on	5:on	6:off
hidd           	0:off	1:off	2:on	3:on	4:on	5:on	6:off
httpd          	0:off	1:off	2:off	3:off	4:off	5:off	6:off
[...]

Issue #815, opened 2015-04-23
RHEL 5.x inspection: "Need binary 'chkconfig' to be available on the inspected system"


gsanso:
Target OS: RHEL 5.10 x86_64

Machinery is showing a message saying it can't find chkconfig binary on the target system. 

```

peon@dhcp95:~> machinery inspect -n rhel510 10.100.103.86
Inspecting 10.100.103.86 for changed-managed-files, config-files, groups, os, packages, patterns, repositories, services, unmanaged-files, users...
Inspecting changed-managed-files...
 -> Found 1 changed files.
Inspecting config-files...
 -> Found 9 changed configuration files.
Inspecting groups...
 -> Found 53 groups.
Inspecting os...
 -> Found operating system 'Red Hat Enterprise Linux Server' version '5.10 (Tikanga)'.
Inspecting packages...
 -> Found 742 packages.
Inspecting patterns...
 -> Patterns are not supported on this system.
Inspecting repositories...
 -> Found 1 repositories.
Inspecting services...
 -> Inspection failed!
Inspecting unmanaged-files...
The content of the following remote directories is ignored: /misc, /net.
The content of the following special directories is ignored: /dev, /proc, /sys, /dev/shm.
 -> Found 936 unmanaged files and trees.
Inspecting users...
 -> Found 38 users.

Errors while inspecting services:
 -> Need binary 'chkconfig' to be available on the inspected system.

```
But the 'chkconfig' binary is already installed.

```
[root@dhcp86 ~]# rpm -q chkconfig
chkconfig-1.3.30.2-2.el5

[root@dhcp86 ~]# chkconfig 
chkconfig version 1.3.30.2 - Copyright (C) 1997-2000 Red Hat, Inc.
This may be freely redistributed under the terms of the GNU Public License.

usage:   chkconfig --list [name]
         chkconfig --add <name>
         chkconfig --del <name>
         chkconfig [--level <levels>] <name> <on|off|reset|resetpriorities>


```

greygoo:
PR to fix the problem by skipping xinetd service status:

https://github.com/SUSE/machinery/commit/b2a6d98ac5b734f2903c5b7518faea912afb331d

ajaeger:
Fixed

PR #834, opened 2015-04-27
Add select statement to only parse chkconfig output for inetd services.

greygoo:
- fix for #815
- when xinetd is installed, chkconfig output has also xinetd values that
  are printed differently than the original inetd services output. This
  broke the chkconfig parser in the services inspector.
- the PR makes sure that only lines matching what th eparser can parse are
  actually parsed.
- the PR does not fix the lack of xinetd services support in machinery.
- the missing support for inspecting xinetd services is tracked in
  #831


tgoettlicher:
It could be helpful to show a warning if xinetd services are on the system but ignored.

tgoettlicher:
Do you want to add tests to make sure parsing the output works correctly?

spec/unit/services_inspector_spec.rb
```diff
@@ -187,5 +187,40 @@
...
+      expect(system).to receive(:check_requirement).
+        with("runlevel")
+      expect(system).to receive(:run_command).
+        with("runlevel", stdout: :capture).
+        and_return("N 5")
+      expect(system).to receive(:run_command).
+        with("chkconfig", "--list", stdout: :capture).
+      and_return(chkconfig_redhat_xinetd_output)
+
+      expect(Machinery::Ui).to receive(:warn)
```
tgoettlicher: I wouldn't check only for some warning to occur but for the actual message.

spec/unit/services_inspector_spec.rb
```diff
@@ -187,5 +187,40 @@
...
+      services = inspector.send(:parse_redhat_chkconfig)
```
tgoettlicher: Lines 218 to 223 look redundant as they are covered be the test above and are not needed for testing if a warning is shown. I don't insist on removing these lines, though.

PR #838, opened 2015-04-28
Add new show_filter_note method

vwallfahrer:
Show note if there are filters applied during inspection.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -126,6 +126,12 @@ class Cli
     Cli.handle_error(e)
   end
 
+  def self.show_filter_note(scopes, filter)
+    if scopes.any? { |scope| !filter.element_filters_for_scope(scope).empty? }
+      Machinery::Ui.puts "\nFilters are applied during inspection.\n\n"
+    end
+  end
+
   def self.shift_arg(args, name)
     if !res = args.shift
       raise GLI::BadCommandLine.new("You need to provide the required argument #{name}.")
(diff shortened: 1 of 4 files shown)


lib/cli.rb
```diff
@@ -116,6 +116,15 @@ def self.handle_error(e)
     Cli.handle_error(e)
   end
 
+  def self.show_filter_note(scopes, filter)
+    scopes.each do |scope|
```
aduffeck: I think this code would look a little bit nicer like that (untested):

```ruby
if scopes.any? { |scope| !filter.element_filters_for_scope(scope).empty? }
  Machinery::Ui.puts "\nFilters are applied during inspection.\n\n"
end
```

lib/cli.rb
```diff
@@ -477,9 +489,11 @@ def self.parse_scopes(scope_string)
 
       filter = FilterOptionParser.parse("inspect", options, global_options)
 
-      if options["verbose"] && !filter.empty?
+      if options[:verbose] && !filter.empty?
```
aduffeck: You're using a symbol here but a string a few lines above. Do both work? But even if both work I think it would be cleaner to use one or the other consistently.

spec/unit/cli_spec.rb
```diff
@@ -73,6 +73,24 @@
         allow(SystemDescription).to receive(:delete)
       end
 
+      describe "filter-note" do
```
aduffeck: Your'e not really describing `filter-note` here but the behavior of `#inspect`, so I would get rid of that line.

PR #841, opened 2015-04-28
Improve format upgrade output

keberle94:
This is the rebased version of #818 

This branch had to rebased because of an movement of the scrub method in the master-branch.


lib/upgrade_format_task.rb
```diff
@@ -52,12 +58,13 @@ def upgrade(store, name, options = {})
 
     if options[:all]
       if migrations_done > 0
-        Machinery::Ui.puts "Upgraded #{migrations_done} system descriptions successfully."
+        Machinery::Ui.puts(
+          "Upgraded #{migrations_done} " \
+            "#{Machinery::pluralize("system description", migrations_done)}."
```
greygoo: I think having a test case testing pluralisation might make sense.

PR #845, opened 2015-04-29
Support xinetd on rhel5 with service inspection fix for 815

greygoo:


plugins/services/services_inspector.rb
```diff
@@ -109,11 +109,20 @@ def parse_redhat_chkconfig
       stdout: :capture
     )
 
-    services = output.lines.map do |line|
+    services = output.lines.select { |line|
```
tgoettlicher: Please use `do` instead of `{` for multi-line blocks.

tgoettlicher:
Looks good to me now. You may want to squash the commits.

PR #846, opened 2015-04-29
Added build matrix to `machinery build --help` output

### LINKED ISSUE #145: Show which systems can be built in output of command line tool
Currently the command line tool refers to the build matrix in the man page. This has the problem that the output is not related to the context and that the man page has to manually be synchronized with the code, which defines the support matrix.

As we know on which platform we are running, we can show which platforms are supported to build. We could also show, on platform the requested build would be supported.

As output of `machinery build --help` we could show the complete build matrix, generated from the definition of build support in the code.

vwallfahrer:
Fix #145


man/machinery-build.1.md
```diff
@@ -53,23 +53,7 @@ built image. The list of filters is shown at the beginning of the build.
...
-    Can build SUSE Linux Enterprise 11 and openSUSE 13.1
-
-  * openSUSE 13.2
-
-    Can build SUSE Linux Enterprise 11, openSUSE 13.1 and openSUSE 13.2
-
-  * openSUSE Tumbleweed
-
-    Can build openSUSE 13.2
+To see the whole build matrix run `machinery build --help`.
```
aduffeck: I don't think a user would know what's meant by `build matrix`.

lib/cli.rb
```diff
@@ -56,6 +56,17 @@ class Cli
 
   GLI::Commands::Help.skips_post = false
 
+  def self.buildable_distributions
+    distribution_string = ""
+    Os.supported_host_systems.each do |distribution|
+      distribution_string += "#{distribution.canonical_name}\n\n"
```
aduffeck: Should there really be a blank line here?

PR #847, opened 2015-04-29
Parse RHEL xinetd service state output from chkconfig.

greygoo:
- fix for https://github.com/SUSE/machinery/issues/815
- when xinetd is installed on a RHEL5 system then chkconfig output
  has additional xinetd service state in a different format
- this change selects chkconfig output by regular expressions and
  applys different extraction fitting different output formats
- adjust services_inspector testcase to also test xinetd services


tgoettlicher:
lgtm

PR #849, opened 2015-04-29
Add dedicated Tumbleweed integration test

thardeck:
Since Tumbleweed will be updated regularly we can't use our regular
integration tests with output matching.

So we use for each inspector some regular expressions to make sure that
they are still able to extract valid data from the system.


aduffeck:
I'd prefer to have Ruby 1.9 style hash syntax in the `test_data` hash, but besides that it lgtm.

thardeck:
I can't use it since we use strings. I could use symbols but then I need to convert them to strings which would mess up the code. That's also the reason why hound doesn't complain.

aduffeck:
It's merged now, but in general it would have been quite easy to use `scope.to_sym` when accessing the hash.

PR #851, opened 2015-04-30
[Review] Request from 'aduffeck' @ 'SUSE/machinery/inspect_as_non_root'

aduffeck:
  * 6307ebc Fix hound issues
  * 864045b Fix tests
  * 6daf8c4 Connect via machinery user in the 13.2 inspection integration test
  * a248d81 Allow for configuring machines which use non-root accounts for SSH
  * 766e7ab Fix double escaping in find command
  * e687ecf Move test RPM injection to the provisioning step
  * ee100b9 Document the required sudo commands when inspecting as non-root
  * 0a388dc Only use sudo where necessary for non-root inspections
  * 3a57388 Change "find" command so that it doesn't rely on bash


thardeck:
lgtm

aduffeck:
Squashed and merged

PR #855, opened 2015-05-04
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150504_fix_integration_tests'

aduffeck:
  * cfcae0c Add missing space when providing inspect options
  * 9d48f5d Provide default username for `prepare_remote_machinery_for_host` (root)
  * 5aaf3b7 Enforce deterministic user ids, otherwise tests may fail after rebuilds


keberle94:
lgtm

PR #856, opened 2015-05-04
Inserting another condition for showing hints

### LINKED ISSUE #369: `machinery help <command>` should not show hints
When calling `machinery help <command>` with a given command the hint help is shown (that there are hints and how to turn them off). When calling `machinery <command> --help` this text is not shown.

The hint help should only be shown when calling `machinery help` without a specific command. This makes the help consistent, and the information is probably more confusing than helpful for the individual command help.

keberle94:
Fixes #369

Hints in the help command are only shown when the help command is called
without any command specification.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -40,7 +40,7 @@ class Cli
   end
 
   post do |global_options,command,options,args|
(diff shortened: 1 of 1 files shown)


aduffeck:
lgtm

PR #864, opened 2015-05-06
Improve format upgrade output

keberle94:
Squashes #841 and removes not branch-related commits, which were pushed by accident.


greygoo:
Now the extra line in the migration output that i saw earlier is gone and the output looks like the acceptance criteria say, so beside the hound issues LGTM.

PR #869, opened 2015-05-07
Show hint for generate html task

mauromorales:
Fix broken test


thardeck:
lgtm besides my comment

PR #871, opened 2015-05-08
Restructuring changelog creation

keberle94:


spec/unit/changelog_source_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+        }
+      }
+    }
+
+    before do
+      subject.parse
+    end
+
+    context "when parsing the rpm changes" do
+      let(:file) { File.join(File.dirname(__FILE__), "..", "data", "changelog", "RPM_CHANGES") }
```
cornelius: It would be nice to have a helper method to find a file in the `spec/data/` directory. This is repeated at several places in this patch. Extracting it as a method would make the tests easier to read.

tools/release.rb
```diff
@@ -130,49 +129,9 @@ def copy_rpmlintrc
...
-
-    changelog = ""
-    changes.each do |version, value|
-      changelog += value
-    end
-
-    File.write(RPM_CHANGES_FILE, changelog)
+  def write_changelog
+    changelog = ChangelogOutput.new(CHANGELOG_FILES)
+    File.write(RPM_CHANGES_FILE, changelog.generate)
```
cornelius: Wouldn't the more natural API look like something like this:

```ruby
changelog = ChangelogOutput.new(CHANGELOG_FILES)
changelog.write_rpm_changes(RPM_CHANGES_FILE)
```

spec/unit/changelog_source_spec.rb
```diff
@@ -0,0 +1,178 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+require_relative "../../lib/changelog_source"
```
cornelius: Would't it be better to require the file in `spec_helper.rb`, so that the test itself don't have to know where the dependencies are, but this an be centrally handled by the spec helper?

spec/unit/changelog_output_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+require_relative "../../lib/changelog_output"
```
cornelius: Same comment as for the require in the other spec file.

lib/changelog_output.rb
```diff
@@ -0,0 +1,71 @@
...
+    rpm_changes = ChangelogSource.new(@rpm_changes)
+    rpm_changes.parse
+    rpm_changes
+  end
+
+  def news
+    news = ChangelogSource.new(@news)
+    news.parse
+    news
+  end
```
cornelius: The `rpm_changes` and `news` method are only used once. It would be simpler, if the code would just be folded back into the methods where they are called, I think.

lib/changelog_output.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+require_relative "./changelog_source"
+
+class ChangelogOutput
```
cornelius: Some basic documentation about the responsibility of this class would be nice. Same for `ChangelogSource`.

spec/unit/changelog_source_spec.rb
mauromorales: I like the idea to require only what is needed for the test in order to make the test run leaner when run individually. But definitely can be changed to just require the helper

cornelius: Does it actually make a difference, when only the one file is included and not the helper?

mauromorales:
Working on a new proposition for this PR now that I have worked a bit more on the release process. Closing for now.

PR #875, opened 2015-05-08
[Review] Request from 'tgoettlicher' @ 'SUSE/machinery/review_150508_improve_error_handling_when_user_lacks_sudo_privileges'

tgoettlicher:
  * 4bdff6c Add changelog for sudo error handling
  * e26632f Improve error handling when user lacks sudo privileges

### DIFF
--- a/helpers/changed_managed_files.sh
+++ b/helpers/changed_managed_files.sh
@@ -15,7 +15,7 @@
 #
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
-
+set -e
 
 # Print a list of each package with changed managed files followed by a list of
 # the changed files, e.g.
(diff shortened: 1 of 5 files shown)


helpers/changed_managed_files.sh
```diff
@@ -15,7 +15,7 @@
 #
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
-
+set -e
```
thardeck: I think there should be a comment why this was added.

helpers/changed_managed_files.sh
```diff
@@ -26,12 +28,18 @@
...
 
 for package in `rpm -qa --queryformat "%{NAME}-%{VERSION}\\n"`; do
-  CHANGES=`$SUDOPREFIX rpm -V --nodeps --nodigest --nosignature --nomtime --nolinkto $package`;
-  if [ -n "$CHANGES" ]; then
-    echo -e "$package:\\n$CHANGES";
+  # rpm returns 1 as exit code when modified config files are detected
+  # that's why we explicitly detect if sudo failed
+  OUTPUT=`$SUDOPREFIX rpm -V --nodeps --nodigest --nosignature --nomtime --nolinkto $package 2>&1 || true`;
+  if [ -n "$OUTPUT" ]; then
+    if [[ "$OUTPUT" ==  *"password is required"* ]]; then
```
thardeck: This is way to dangerous that somewhere in the regular line an output could appear like this.

I would check for the exact output if it is the same on each distribution and of course make sure that the correct language environment is set.

"sudo: a password is required"

lib/remote_system.rb
```diff
@@ -72,9 +72,15 @@ def run_command(*args)
       cheetah_class = LoggedCheetah
     end
 
-    sudo = "sudo" if options[:privileged] && remote_user != "root"
-    cmds = ["ssh", "#{remote_user}@#{host}", sudo, "LC_ALL=C", *piped_args, options].compact
+    sudo = ["sudo", "-n"] if options[:privileged] && remote_user != "root"
```
thardeck: Why do you convert it to an array?

spec/unit/remote_system_spec.rb
```diff
@@ -80,12 +80,23 @@
...
+
+      it "raises an exception if the user is not allowed to run sudo" do
+        expect(Cheetah).to receive(:run).with(
+          "ssh", "machinery@remotehost", "sudo", "-n", "LC_ALL=C", "ls", "/tmp", privileged: true
+        ).and_raise(Cheetah::ExecutionFailed.new(nil, 1, "", "sudo: a password is required"))
+
+        remote_system.remote_user = "machinery"
+        expect {
+          remote_system.run_command("ls", "/tmp", privileged: true)
+        }.to raise_error(Machinery::Errors::InsufficientPrivileges)
```
thardeck: I would check the error output too with since you don't check it anywhere else I think.

cornelius:
lgtm

PR #876, opened 2015-05-08
Filter /root/.local from unmanaged-file tests

thardeck:
The existence of "/root/.local" seems to be quite non-deterministic.
That is why we filter it in the unmanaged-file integration tests.


aduffeck:
lgtm

PR #877, opened 2015-05-11
Added script which creates the inspector files

vwallfahrer:


cornelius:
What's the purpose of this script?

vwallfahrer:
The script extracts the test data from a new system. It creates files which include the scope information in spec/data/.

cornelius:
The script should have some documentation, for what it is supposed to be used, and how to do that.

Also the commit message should be more descriptive and state why this was added and what it is.

helpers/inspector_files.rb
```diff
@@ -0,0 +1,42 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+require "fileutils"
+require "cheetah"
+
+class WriteInspectorFiles
```
cornelius: The name of the class should be a noun which describes the responsibility of the class.

helpers/inspector_files.rb
```diff
@@ -0,0 +1,42 @@
...
+  @@machinery = File.expand_path("../bin/machinery", File.dirname(__FILE__))
+  @@path = File.expand_path("../spec/data/", File.dirname(__FILE__))
```
cornelius: `path` is a very generic name. It would be better to reflect in the name of the variable what kind of path this is.

helpers/inspector_files.rb
```diff
@@ -0,0 +1,42 @@
...
+      )
+      File.write(file_path(inspector, distribution), output)
+    end
+  end
+
+  def self.file_path(inspector, distribution)
+    FileUtils.mkdir_p("#{@@path}/#{inspector}")
+    File.join(@@path, inspector, distribution)
+  end
+end
```
cornelius: The class does everything in class methods and variables. It would probably be better and simpler to encapsulate the state in the state of an object of the class and use normal methods and member variables.

helpers/inspector_files.rb
vwallfahrer: What do you propose?

cornelius: I would propose to document and structure the code first. Maybe the best name of the class will naturally follow from that.

PR #878, opened 2015-05-11
Allow tumbleweed build with new snapshot

thardeck:


machinery.spec.erb
```diff
@@ -113,22 +113,24 @@ bundle install --standalone --local
...
 # E: file-contains-buildroot (Badness: 10000)
 rm -rf bundle/ruby/%{rb_ver}/extensions/*/%{rb_ver}/nokogiri-*/mkmf.log
 rm -rf bundle/ruby/%{rb_ver}/extensions/*/%{rb_ver}/nokogiri-*/gem_make.out
+sed -i '\|%{buildroot}|d' bundle/bundler/setup.rb
 
 popd
 
 # Adapt the binary
 
+# make sure the machinery executable is always called `machinery`
```
cornelius: I would add a comment why it is not called `machinery` in what cases

thardeck: I actually had this before but I thought there were to many comments in a row.

thardeck:
I don't know them myself, it is gem2rpm magic or something like this.

machinery.spec.erb
```diff
@@ -113,22 +113,26 @@ bundle install --standalone --local
 
 rm -rf vendor .bundle Gemfile Gemfile.lock
 
-# remove Nokogiri build logs to prevent the following RPMLint error:
+# Remove references to the build directory to prevent the following RPMLint error:
 # E: file-contains-buildroot (Badness: 10000)
 rm -rf bundle/ruby/%{rb_ver}/extensions/*/%{rb_ver}/nokogiri-*/mkmf.log
 rm -rf bundle/ruby/%{rb_ver}/extensions/*/%{rb_ver}/nokogiri-*/gem_make.out
+# the nokogiri gem is mentioned twice in the setup.rb in Tumbleweed (once with
```
cornelius: Ok, to make the patch perfect start this sentence with an upper-case letter as well. Then it looks good to me.

PR #881, opened 2015-05-11
Add Tumbleweed rpm spec fixes to RPM_CHANGES

thardeck:


RPM_CHANGES
```diff
@@ -1,5 +1,7 @@
 # Machinery RPM Changelog
 
+* remove absolute build path from bundler setup.rb (Tumbleweed)
+* remove ruby version suffix from machinery executable (Tumbleweed and 13.1)
```
tgoettlicher: Writing `openSUSE 13.1` instead of just `13.1` would be more precise. If you think it's clear in this context feel free to keep it as is.

tgoettlicher:
LGTM

PR #882, opened 2015-05-11
Always remove all gems under package/

thardeck:
Having more than on gem in the package directory messes up the
rpm build process in Tumbleweed and also 13.1 since we use the
Devel repository for it.

It should be OK to just archive the rpms so the gems are always
removed.


tools/release.rb
```diff
@@ -101,8 +101,12 @@ def self.generate_release_version(release_type)
 
   private
 
-  def remove_old_releases
-    FileUtils.rm Dir.glob(File.join(Machinery::ROOT, "package/*"))
+  def remove_old_releases(skip_cleanup)
```
tgoettlicher: `package/*.gem` and `/var/tmp/.../*.rpm` are removed although the `skip_cleanup` option is set. 
If that's on purpose, the option name is misleading.

cornelius: Make this a named parameter. `remove_old_releases(true)` is much less expressive than `remove_old_releases(skip_cleanup: true)`.

thardeck: I think it is Ok, since we only clean up the tmp directory there which should be always fine. The reason why we added skip_cleanup was to preserve the packages for our integration test which should still work.

tools/release.rb
```diff
@@ -101,8 +101,12 @@ def self.generate_release_version(release_type)
 
   private
 
-  def remove_old_releases
-    FileUtils.rm Dir.glob(File.join(Machinery::ROOT, "package/*"))
+  def remove_old_releases(options = {})
```
cornelius: Using actual named parameters makes it better to read:

```ruby
def remove_old_releases(skip_cleanup: false)
  if skip_cleanup
    ...
```

tgoettlicher:
From my perspective it looks good now. Please check if you also fixed the issues Cornelius mentioned.

cornelius:
Looks good to me now as well.

PR #885, opened 2015-05-12
Review 150508 uniform file extraction

aduffeck:
This pull request abstracts the file extraction and handling of resulting files and makes use of it in the Kiwi export.

Later the autoyast export and changed_managed_files scope will be ported to the new classes as well.


lib/array.rb
```diff
@@ -49,11 +49,19 @@ def from_json(json_object)
     end
 
     attr_reader :elements
+    attr_accessor :scope
 
     def initialize(elements = [])
       @elements = self.class.convert_raw_array(elements)
     end
 
+    def set_scope(scope)
```
cornelius: Wouldn't it be better to use the `scope=` method to set the scope?

lib/object.rb
```diff
@@ -55,11 +55,19 @@ def from_json(json_object)
     end
 
     attr_reader :attributes
+    attr_accessor :scope
 
     def initialize(attrs = {})
       set_attributes(attrs)
     end
 
+    def set_scope(scope)
```
cornelius: Same here as in the `Machinery::Array` class.

lib/system_description.rb
```diff
@@ -91,7 +91,7 @@ def validate_name(name)
     def from_hash(name, store, hash)
       begin
         json_format_version = hash["meta"]["format_version"] if hash["meta"]
-        description = SystemDescription.new(name, store, create_scopes(hash))
+        description = SystemDescription.new(name, store, create_scopes(store, name, hash))
```
cornelius: It looks wrong that `store` is used twice in the parameter list in this line. I suppose this is an indication that the code isn't structured optimally yet.

lib/kiwi_config.rb
```diff
@@ -104,16 +104,21 @@ def inject_extracted_files(output_location)
...
+      destination = File.join(output_location, "root", "tmp", "unmanaged_files")
       FileUtils.mkdir_p(destination, mode: 01777)
-      FileUtils.cp_r(unmanaged_files_path, destination)
+      filter = "unmanaged_files_#{@name}_excludes"
+
+      @system_description.unmanaged_files.write_files_tarball(destination)
+      @system_description.unmanaged_files.files.each do |file|
+        next if !file.directory?
+
+        file.utils.write_tarball(File.join(destination, "trees"))
```
cornelius: I would prefer 

```ruby
  if file.directory?
    file.utils.write_tarball...
  end
```

here. Less indirect.

aduffeck: Counter proposal: 

```ruby

      @system_description.unmanaged_files.files.select(&:directory?).each do |directory|
        directory.utils.write_tarball(File.join(destination, "trees"))
      end
```
:)

lib/kiwi_config.rb
```diff
@@ -104,16 +104,21 @@ def inject_extracted_files(output_location)
...
+      end
```
cornelius: Maybe the writing of the tarballs could be extracted into an own method, so it would just become

```ruby
@system_description.unmanaged_files.export_files_as_tarballs(destination)
```

lib/kiwi_config.rb
cornelius: Or:

``ruby
@system_description.unmanaged_files.write_directory_tarballs(destination)
```

File: lib/system_description.rb [hunk 6]
```diff
@@ -111,19 +111,21 @@ def from_hash(name, store, hash)
...
-      scopes = hash.map do |scope_name, scope_json|
-        next if scope_name == "meta"
+    def create_scopes(store, name, hash)
+      scopes = hash.map do |scope, json|
+        next if scope == "meta"
 
-        scope_class = Machinery::Scope.class_for(scope_name)
-        scope_object = scope_class.from_json(scope_json)
+        if store.persistent?
+          scope_file_store = ScopeFileStore.new(store.description_path(name), scope)
```
cornelius: We have `SystemDescription#scope_file_store` already which does this, don't we?

lib/kiwi_config.rb
aduffeck: :+1:

lib/system_description.rb
```diff
@@ -111,19 +111,21 @@ def from_hash(name, store, hash)
 
     private
 
-    def create_scopes(hash)
-      scopes = hash.map do |scope_name, scope_json|
-        next if scope_name == "meta"
+    def create_scopes(store, name, hash)
+      scopes = hash.map do |scope, json|
```
cornelius: This actually is a `scope_name`, not a `scope` object, right?

lib/scope.rb
```diff
@@ -26,9 +26,14 @@ def self.all_scopes
       @scopes
     end
 
-    def self.class_for(scope_name)
+    def self.initialize_scope(scope_name, json, scope_file_store)
```
cornelius: In other places we use the convention to name factory methods `for`, so could we do the same here?

lib/system_description.rb
aduffeck: We do, but only as an instance method. This code is defined on the class instead.

lib/file_scope.rb
```diff
@@ -16,6 +16,13 @@
 # you may find current contact information at www.suse.com
 
 class FileScope < Machinery::Object
+  def write_files_tarball(destination)
+    FileUtils.cp(
+      File.join(scope_file_store.path, "files.tgz"),
+      destination
+    )
+  end
```
cornelius: Doesn't this method belong to the `SystemFileUtils` class? I think we should have only one class which is responsible for the structure of how extracted files are stored on disk as part of the system description.

lib/system_description.rb
cornelius: Another indication that `create_scopes` might better be an instance method as well.

lib/scope.rb
aduffeck: Those are a little bit different, because they only return the appropriate class, not an instance like here.

lib/file_scope.rb
aduffeck: Right, we were also pondering about that. But wrinting a tarball doesn't make sense for `SystemFile`s of type "file", because all of them are encapsulated in only one tarball.

We put it in the scope itself because the result will be that one tarball with all files of that scope (and links, but not directories) will be written somewhere.

cornelius: Maybe the utils class should have a broader scope and not only handle one file, but also the collection of files.

lib/scope.rb
cornelius: Seems we are not handling this consistently. `Renderer::for` instantiates the object, `Inspector::for` returns the class.

What about a naming convention of using `class_for` to return the class, and `for` to return an instantiated object?

cornelius:
I still see some overlap between the `FileExtractor`, `SystemFileUtils` and `ScopeFileStore` classes. They are coupled in some explicit and some implicit ways. But I think the pull request as it is improves the situation already quite a bit, because it makes it much more explicit than before, where things are, and it makes it easier to change things in the future.

So I would be fine with merging it for now. We should keep in mind that we might want to further refactor the classes handling the internal storage of files in the system description, though.

cornelius:
Prophet should be green again before we merge it, of course.

aduffeck:
Ok, I fixed the tests and merged that branch into the uniform_file_extraction feature branch.

PR #888, opened 2015-05-13
Added script for integration tests

vwallfahrer:
Superseds #877 

The script generates files with scope informations. Usefull for the integration tests.


helpers/inspector_files.rb
```diff
@@ -0,0 +1,52 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+
```
vwallfahrer: I added the blank line for better overview.

cornelius:
The documentation might be more easy to understand if you frame it a little bit differently. The key point is to create reference data for integration tests from inspecting existing machines. So maybe you can change the wording to express this a little bit better. I find the information that it creates reference data for tests as the most important one.

Rakefile
```diff
@@ -140,6 +141,19 @@ task :release, [:type] do |task, args|
   release.publish
 end
 
+desc "Build files with scope information for integration tests"
+
+# This command can be used to generate files with information
+# of each scope of an inspected system automatically.
+# This is usefull for the integration tests.
+
+task :inspector_files, [:ip_adress, :distribution] do |task, args|
```
cornelius: What does `distribution` mean and what are the allowed values for that? This should be made clearer.

helpers/inspector_files.rb
```diff
@@ -0,0 +1,52 @@
...
+require "cheetah"
+
+class CreateScopeInfoTestData
+  def initialize
+    @machinery = File.expand_path("../bin/machinery", File.dirname(__FILE__))
+    @test_data_path = File.expand_path("../spec/data/", File.dirname(__FILE__))
+  end
+
+  def inspect_system(ip_adress)
+    Cheetah.run(@machinery, "inspect", ip_adress, "-n", "machine", stdout: :capture)
```
cornelius: This writes a description named `machine` to the users machinery directory. This could cause conflicts. Maybe better use a random name and remove the description after its use.

helpers/inspector_files.rb
```diff
@@ -0,0 +1,52 @@
...
+# This script can be used to generate files with information
+# of each scope of an inspected system automatically.
+# This is usefull for the integration tests.
+
+# Usage: rake inspector_files[ip-adress,distribution]
+
+require "fileutils"
+require "cheetah"
+
+class CreateScopeInfoTestData
```
cornelius: Class names should always be nouns. The verbs are the methods.

I would probably call it `ReferenceTestData` and have two methods `inspect(ip_address)` and `write(destination)`.

PR #891, opened 2015-05-13
Added script for integration tests

vwallfahrer:
Superseds #888


cornelius:
Looks good to me. I would suggest to merge it into one commit before merging to master.

PR #892, opened 2015-05-15
Fix integration tests unfiltered tgz files

mauromorales:


cornelius:
The refactoring definitely should go in, and the additional filter also looks fine. So I'm fine with merging this.

For the future shouldn't we move to using filters during inspection (using the `--exclude` option), so that we don't have to filter any results ourselves?

mauromorales:
@cornelius makes sense ... we changed it to use the `--skip-files` options instead

PR #893, opened 2015-05-18
[Review] Request from 'aduffeck' @ 'SUSE/machinery/abstract_changed_managed_files_file_access'

aduffeck:
  * 53110eb Do not require the "type" attribute for "deleted" changed managed files
  * ae24f09 Use new SystemFile/SystemFileUtils APIs in kiwi and autoyast
  * 82b72e7 Implement SystemFileUtils.write_file for copying untared extracted files
  * 33f362f Also store the file type (dir, link, file) when inspecting RPM files
  * 6c86428 Apply links and link attributes in kiwi and autoyast export
  * 275f62c Add support for links to the changed-managed-files inspector
  * 62a68ab Apply links in changed managed files in the kiwi export
  * f35fbdc Add support for links in the changed-managed-files data model
  * f4ac5b9 Fix script for upgrading test descriptions to work with the current data
  * 7a60296 Upgrade to format version 4 (support links in changed managed files)
  * ee770fc Implement SystemFile#deleted?


lib/autoyast.rb
```diff
@@ -266,6 +266,12 @@ def apply_extracted_files(scope)
     return if !@system_description[scope] || !@system_description[scope].extracted
 
     base = Pathname(@system_description.scope_file_store(scope).path)
+
+    @system_description[scope].files.select { |file| file.link? }.each do |link|
+      @chroot_scripts << "rm -rf '#{link.name}'"
+      @chroot_scripts << "ln -s '#{link.target}' '#{link.name}'"
+    end
```
cornelius: Is this code used at all? This is only called for the config-files scope, where we haven't done the changes to extract the necessary informatioin, isn't it?

cornelius:
In general this patch looks fine. I definitely think we need to review the API to the storage of extracted files. This isn't completely done and is not consistent yet. But in the context of this patch I think it's ok, as it allows to change it easily and in the changed code the intention is expressed through explicit API.

I haven't tested the code. I would suggest that we merge it and then start QA on the branch, so we get some test coverage before it goes to master.

lib/autoyast.rb
aduffeck: Good point, that's a left-over from some intermediate step I had in there. I'll remove it before merging the branch.

PR #911, opened 2015-05-21
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150521_fix_missing_link_target'

### LINKED ISSUE #908: `machinery -s changed-managed-files` creates invalid system description
Running `machinery inspect host -s changed-managed-files` without `--extract-files` option create an invalid system description.

```
In scope changed_managed_files: The property #17 (files/changes) of type Hash did not match any of the required schemas.
```

aduffeck:
  * 13caf88 Store link targets in manifest.json
  * c55cec5 Improve check which determines whether something needs to be extracted

Fixes #908


spec/unit/changed_rpm_files_helper_spec.rb
```diff
@@ -151,3 +151,12 @@
     )
   end
 end
+
+describe "#get_link_target" do
+  it "returns the link target" do
+    system = "double"
+    expect(system).to receive(:run_command).and_return("/etc/foo\n")
```
tgoettlicher: Wouldn't it make sense to check that run_command is called with "readlink" as argument?

lib/changed_rpm_files_helper.rb
```diff
@@ -93,6 +93,10 @@ def parse_stat_line(line)
     ]
   end
 
+  def get_link_target(system, link)
+    system.run_command("readlink", link, stdout: :capture).strip
```
tgoettlicher: Can you please add `readlink` to list of required commands on the man page https://github.com/SUSE/machinery/blob/master/man/machinery-inspect.1.md ?

spec/unit/changed_rpm_files_helper_spec.rb
aduffeck: I think whether the inspector uses `readlink` to get the link target or something else is not of the test's concern. It only couples the test more tightly to the implementation without adding any value imo.

PR #915, opened 2015-05-26
Adding line break and result if upgrade failed for `machinery upgrade-format --all`

### LINKED ISSUE #914: `machinery upgrade-format --all` misses line break and result if upgrade failed
If everything works well, there is one system description incl. the result per line:
```
$ machinery upgrade-format --all
Reading "host1" ... Successfully upgraded from version 3 to 4.
Reading "host2" ... Successfully upgraded from version 3 to 4.
Upgraded 2 system descriptions.
```

If the upgrade of system descriptions fails the result and line break are missing:
```
$ machinery upgrade-format --all
Reading "host1" ... Reading "host2" ... Reading "host3" ... 
...
```

keberle94:
Fixes #914


lib/upgrade_format_task.rb
```diff
@@ -46,12 +46,13 @@ def upgrade(store, name, options = {})
         end
       rescue StandardError => e
         errors.push("Upgrading description \"#{description}\" failed:\n#{e}")
+        puts "Upgrade failed."
```
aduffeck: This should use `Machinery::Ui.puts` instead.

aduffeck:
lgtm

PR #917, opened 2015-05-26
Just testing the prophet server

mauromorales:


mauromorales:
tests are passing after vagrant and libvirt plugin upgrade

PR #918, opened 2015-05-27
[Review] Request from 'aduffeck' @ 'SUSE/machinery/fix_repositories_inspection_with_sudo'

aduffeck:
  * cad7033 Fix reading zypp credentials when using a remote user


thardeck:
lgtm

PR #925, opened 2015-05-29
Do not execute rpm verification scripts

tgoettlicher:
This commit adds the `--noscripts` argument to rpm verification calls, which makes verification scripts not to be executed. The output of verification scripts is not standardized and therefor cannot be parsed by Machinery. That's why we don't want verification scripts to be executed.


thardeck:
Have you tested it with our oldest distributions, rhel5 and sles11sp3?

tgoettlicher:
It works with old RHEL. SLES11SP3 works if the latest update of `rpm` is installed. I'll better add a check for the version.

tgoettlicher:
squashed is: https://github.com/SUSE/machinery/pull/930

PR #926, opened 2015-05-29
Move html of compared machines to the users home

mauromorales:
issue: https://github.com/SUSE/alfred/issues/1108


mauromorales:
The problem right now is that the first user who creates the directory gets full ownership of it so the next user cannot overwrite it. The idea to move it to the user home path is that each user can do the comparison with the benefit that they won't destroy other users data either

lib/compare_task.rb
```diff
@@ -41,7 +41,7 @@ def render_html_comparison(description1, description2, scopes, options)
       end
     end
 
-    target = "/tmp/machinery-html-comparison"
+    target = File.join(Machinery::DEFAULT_CONFIG_DIR, "machinery-html-comparison")
```
cornelius: As this is a Machinery specific directory, it would probably be enough to just call it `html-comparison`.

The constant name `DEFAULT_CONFIG_DIR` is misleading, as this is not the directory for configuration, but for all Machinery data. I think we should change that. It makes this patch a bit confusing to read, but if we change it we should do it in a separate patch.

mauromorales: @cornelius got it. How about I put the full path, that should make it easier to spot what is meant to be done there. See next commit

mauromorales:
@cornelius my bad, completely missunderstood that :/ ... switched to use the constant again and will work on the refactoring of it into a different branch

cornelius:
lgtm now. should be squashed before merging for a cleaner history, though.

mauromorales:
squash done on PR: https://github.com/SUSE/machinery/pull/938

PR #930, opened 2015-06-02
Do not use rpm verification scripts

tgoettlicher:
This commit adds the `--noscripts` argument to rpm verification calls, which makes verification scripts not to be executed. The output of verification scripts is not standardized and therefor it cannot be parsed by Machinery. That's why we don't want verification scripts to be executed. The version of rpm initially shipped with SLES11SP3 doesn't support `--noscripts`. A workaround is used in this case.


tgoettlicher:
Variables by the operating system or start up scripts are usually all in CAPITALS, these are called 'environment variables'. To prevent your own variables to conflict with environment variables, it is a good practice to use lower case.

thardeck:
I think changing the existing changed-managed-file script and adding a new one should be a separate task.

But in general I would prefer to create one for both. I am willing to do this since I am planning to refactor both inspectors anyway as soon as the new link recognition is merged.

thardeck:
Ok, we are not planning to refactor the inspector since they will be replaced soon but then the two bash scripts should be merged in this commit.

tgoettlicher:
See: https://github.com/SUSE/machinery/pull/936

PR #932, opened 2015-06-02
 Refactoring: used new ruby hash syntax 

vwallfahrer:
Superseded #931


spec/unit/html_spec.rb
```diff
@@ -75,36 +75,36 @@
...
+          { type: "common",   new_line_number: 7, original_line_number: 7, content: "========" },
+          { type: "common",   new_line_number: 8, original_line_number: 8, content: "- This software is free. You can do with it whatever you want." },
+          { type: "common",   new_line_number: 9, original_line_number: 9, content: "  I would however kindly ask you to acknowledge the use of this" },
+          { type: "deletion", original_line_number: 10, content: "  package, if you are going use it in your software, which you might" },
+          { type: "deletion", original_line_number: 11, content: "  be going to distribute. I would also like to receive a note if" },
+          { type: "common",   new_line_number: 10, original_line_number: 12, content: "  you are a satisfied user :-)" },
+          { type: "addition", new_line_number: 11, content: "  Change two" },
+          { type: "common",   new_line_number: 12, original_line_number: 13, content: nil },
+          { type: "common",   new_line_number: 13, original_line_number: 14, content: "Acknowledgements:" },
+          { type: "common",   new_line_number: 14, original_line_number: 15, content: "=================" },
```
vwallfahrer: Ignored Hound messages because with line breaks it gets unclear.

lib/build_task.rb
```diff
@@ -35,8 +35,12 @@ def build(system_description, output_path, options = {})
...
 
       begin
-        LoggedCheetah.run("sudo", tmp_script.path, :stdout => $stdout,
-          :stderr => $stderr)
+        LoggedCheetah.run(
+          "sudo",
+          tmp_script.path,
+          stdout: $stdout,
+          stderr: $stderr
+        )
```
thardeck: I would prefer something like this but it is not worth changing.
```
        LoggedCheetah.run(
          "sudo",  tmp_script.path, stdout: $stdout, stderr: $stderr
        )
```

thardeck:
lgtm

Issue #927, opened 2015-06-01
kernel_require.rb:55:in `require': cannot load such file -- abstract_method (LoadError)


schwindp:
Note: I'm trying to run machinery on a FC21 box, haven't found any explicit warnings against it...
```
[pschwindt@korn:~/src/machinery]$ ruby bin/machinery 
/usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require': cannot load such file -- abstract_method (LoadError)
	from /usr/share/rubygems/rubygems/core_ext/kernel_require.rb:55:in `require'
	from /home/pschwindt/src/machinery/lib/machinery.rb:20:in `<top (required)>'
	from bin/machinery:20:in `require_relative'
	from bin/machinery:20:in `<main>'
```

machinery source was checked out right before, so it's up to date.

```
[pschwindt@korn:~/src/machinery]$ rpm -qa|grep ruby
rubygem-psych-2.0.5-29.fc21.x86_64
rubygem-json-1.8.1-29.fc21.x86_64
rubygem-io-console-0.4.3-29.fc21.x86_64
ruby-2.1.6-29.fc21.x86_64
rubypick-1.1.1-2.fc21.noarch
rubygems-2.2.2-101.fc21.noarch
rubygem-bigdecimal-1.2.4-29.fc21.x86_64
rubygem-rdoc-4.1.1-2.fc21.noarch
ruby-libs-2.1.6-29.fc21.x86_64
ruby-irb-2.1.6-29.fc21.noarch
```

thardeck:
Thanks for your report. Fedora is atm not supported to run machinery although inspection of it should work.

How did you install machinery? The reason for your issue is the missing abstract_method gem.

But as soon as you have all required gems installed machinery would exit with a warning that the platform is not supported.

schwindp:
Thanks for coming back that quickly: I pulled machinery from git this morning. 

And thanks for enlightening me regarding the missing gem, I'll have a look then.

(closing since unsupported on FC21, thanks for helping anyway)

thardeck:
If you are using the git tree you can run `bundle install` in the machinery directory to install all required gems.

thardeck:
So the error should now be a warning which also can be disabled by running `machinery config perform-support-check false`

PR #935, opened 2015-06-03
Platformcheck is a warning instead of an error

keberle94:
Related to #927


spec/unit/local_system_spec.rb
```diff
@@ -102,33 +102,23 @@
       }.not_to raise_error
     end
 
-    it "raises Machinery::Errors::IncompatibleHost on hosts that can not run machinery" do
+    it "puts an IncompatibleHost message on hosts that can not run machinery" do
```
thardeck: I would prefer something like "shows a warning on hosts that can not run machinery"

spec/unit/local_system_spec.rb
```diff
@@ -102,33 +102,23 @@
...
       allow(LocalSystem).to receive(:os).and_return(OsSles11.new)
+      LocalSystem.validate_machinery_compatibility
 
-      expect {
-        LocalSystem.validate_machinery_compatibility
-      }.to raise_error(Machinery::Errors::IncompatibleHost)
+      expect(captured_machinery_output).to include(
+        "Running Machinery is not supported on this system.\nSupported operating systems are:"
+        )
```
thardeck: I would add an additional expect for at least one supported OS which won't change in the nearby future like SLES12

lib/local_system.rb
```diff
@@ -50,7 +50,7 @@ def validate_machinery_compatibility
         message = "Running Machinery is not supported on this system.\n" \
           "Supported operating systems are: #{supported_oses}"
 
-        raise(Machinery::Errors::IncompatibleHost.new(message))
+        Machinery::Ui.puts message
```
cornelius: I would prefer a more welcoming text here. Maybe something like "You are running Machinery on a platform we don't explicitly support and test. It still could work very well. If you run into issues or would like to provide us feedback, you are welcome to file an issue at https://github.com/SUSE/machinery/issues/new or write an email to machinery@lists.suse.com."

Any maybe it should be possible to switch this message off, or we should only show it once, because when you run Machinery regularly it might be very annoying.

keberle94:
Good idea, should I implement a new option for that, or should I implement the message as a hint ?

keberle94:
Superseeded by #942

PR #937, opened 2015-06-03
Do not use rpm verification scripts

tgoettlicher:
This commit adds the `--noscripts` argument to rpm verification calls, which makes verification scripts not to be executed. The output of verification scripts is not standardized and therefor it cannot be parsed by Machinery. That's why we don't want verification scripts to be executed. The version of rpm initially shipped with SLES11SP3 doesn't support `--noscripts`. A workaround is used in this case.


helpers/inspect_files.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
thardeck: I would prefer the file to be called "inspect_changed_files.sh" or "changed_files.sh" but it is not that important.

tgoettlicher:
Renamed the script and squashed: https://github.com/SUSE/machinery/pull/953

PR #938, opened 2015-06-03
Move html of compared machines to the users home

mauromorales:


mauromorales:
lgtm is on https://github.com/SUSE/machinery/pull/926

PR #940, opened 2015-06-05
Check number of parameters

mauromorales:


mauromorales:
switched to https://github.com/SUSE/machinery/pull/943

PR #943, opened 2015-06-05
Check number of parameters

mauromorales:


aduffeck:
LGTM, but do you mind adding a test for that?

mauromorales:
@aduffeck ^__^ of course ... totally missed that :/

PR #944, opened 2015-06-05
Platformcheck is a warning instead of an error

keberle94:
Superseeds #942


lib/config.rb
```diff
@@ -38,6 +38,10 @@ def define_entries
         description: "Enable experimental features. See " \
           "https://github.com/SUSE/machinery/wiki/Experimental-Features for more details"
       )
+      entry("perform_supportability_check",
```
aduffeck: I know I suggested it but after reading it here I think that `supportability` does not really make sense. `perform_support_check`?

spec/unit/local_system_spec.rb
```diff
@@ -94,41 +94,49 @@
   end
 
   describe ".validate_machinery_compatibility" do
-    it "returns true on hosts that can run machinery" do
-      allow(LocalSystem).to receive(:os).and_return(OsOpenSuse13_1.new)
+    context "on hosts that can run machinery" do
+      it "returns true" do
```
aduffeck: The question is whether a warning is put out, not the return value, right?

spec/unit/local_system_spec.rb
```diff
@@ -94,41 +94,49 @@
...
+        allow(LocalSystem).to receive(:os).and_return(OsOpenSuse13_1.new)
 
-      expect {
-        LocalSystem.validate_machinery_compatibility
-      }.not_to raise_error
+        expect {
+          LocalSystem.validate_machinery_compatibility
+        }.not_to raise_error
```
aduffeck: The method never raises an exception, so that test does not make sense. You should rather check that `captured_machinery_output` is empty.

aduffeck:
lgtm, but the branch needs to be squashed and rebased.

keberle94:
Squashed in to #946

PR #945, opened 2015-06-05
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150605_fix_797'

aduffeck:
  * b6b5901 Don't alter SystemDescription's Filter in the unmanaged-files inspector
  * a7b1fc5 Make sure that dupped ElementFilters don't affect the original


mauromorales:
lgtm

PR #951, opened 2015-06-08
Show message in comparison output when scopes are identical

keberle94:
Superseeds #950


aduffeck:
Some of the commits contain debug statements or bugs which were fixed in later commits. I think it would be good to clean that up and squash those commits a bit.

keberle94:
Squashed by #952

PR #952, opened 2015-06-08
Show message in comparison output when scopes are identical

keberle94:
Squashes #951


aduffeck:
The only issue I see is that the messages says "Following scopes" also when it is only one scope. I'm fine with merging this branch like that and fixing the pluralization separately though, if you prefer.

lgtm

keberle94:
Superseeded by case sensitive #955

PR #955, opened 2015-06-08
Show case sensitive message in comparison output when scopes are identical

keberle94:
Superseeds #952 with case sensitivity


keberle94:
Superseeded by #958

PR #957, opened 2015-06-08
Test changed_managed_files in build integration test

thardeck:
We accidentally tested the config-files twice.


keberle94:
Related integration test fails locally. Is that another issue or is this commit-related ?

keberle94:
Now tests are green. LGTM

PR #958, opened 2015-06-08
Show numerus sensitive message in comparison output when scopes are identical 

keberle94:
Superseeds #955


lib/compare_task.rb
```diff
@@ -94,9 +97,13 @@ def render_comparison(description1, description2, scopes, options = {})
         identical = false if description1[scope] || description2[scope]
       end
     end
+    plural = "scopes are"
+    singular = "scope is"
+    count = identical_scopes.count
```
thardeck: Why use variables if the are only used once?

lib/compare_task.rb
```diff
@@ -94,9 +97,13 @@ def render_comparison(description1, description2, scopes, options = {})
...
 
     output = "Compared descriptions are identical.\n" + output if identical && common_scopes
-
+    output += "Following #{Machinery::pluralize(count, singular, plural)} " \
+      "identical in both descriptions: " + identical_scopes.join(",") if !identical_scopes.empty?
```
thardeck: Even though this is for the interpreter only one line these are two for us so I would prefer `if` before the the output statement.

spec/unit/helper_spec.rb
```diff
@@ -18,6 +18,33 @@
...
+      end
+      it "returns given plural text if the number is zero" do
+        expect(Machinery::pluralize(0, "text")).to eq "texts"
+      end
+      it "returns given plural text if the number is more than one" do
+        expect(Machinery::pluralize(5, "text")).to eq "texts"
+      end
+    end
+
+    context "pluralform is given" do
```
thardeck: a "when" is missing

thardeck:
lgtm besides my comment and the last two commits should be merged

keberle94:
Superseeded with commented changes and squashed in #959

PR #558, opened 2015-01-27
Cache descendants of Os class

cornelius:
This speeds up unit tests by a factor of three.

PR #959, opened 2015-06-09
Show numerus sensitive message in comparison output when scopes are identical

keberle94:
Superseeds #958


aduffeck:
lgtm

keberle94:
Merged #558

PR #960, opened 2015-06-09
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150609_inspection_progress_outpu'

aduffeck:
  * f7b3179 Add changelog entry for inspection progress output
  * 63558b4 Adapt progress output to match the other output
  * 1924903 Add progress output to the config-files inspector
  * 55fb63d Add Inspector#run_inspect_script
  * 778bd7e Show progress information while inspecting changed-managed-files
  * 885e3a5 Only show progress information when outputting to a terminal
  * bb137fe Show progress information when inspecting unmanaged files
  * d968d0b Add Machinery::Ui.progress() for progress info, e.g. when inspecting


greygoo:
lgtm from a functional perspective. I checked out the code and tried several things mentioned in the comments and it looks good to me.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -90,7 +90,10 @@ def extract_unmanaged_files(files, trees, excluded, store_name)
     archive_path = File.join(store_path, "files.tgz")
     @system.create_archive(files.join("\0"), archive_path, excluded)
 
+    extracted_count = files.length
+
     trees.each do |tree|
+      Machinery::Ui.progress(" -> Extracted #{extracted_count} files and trees...")
```
tgoettlicher: I wonder if you considered to use `Machinery::pluralize` here?

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -343,6 +348,10 @@ def inspect(filter, options = {})
       unmanaged.map!{ |d| find_dir + d }
       unmanaged_files.push(*unmanaged)
       links.each { |d| unmanaged_links[find_dir + d] = "" }
+
+      Machinery::Ui.progress(
+        " -> Found #{unmanaged_files.length + unmanaged_trees.length} files and trees..."
```
tgoettlicher: I wonder if you considered to use `Machinery::pluralize` here, too?

lib/inspector.rb
```diff
@@ -89,4 +89,30 @@ def scope
...
+  # periodically with new output when it occurs.
+  #
+  # Example:
+  #
+  #     count = 0
+  #     raw_list = run_inspect_script("changed_managed_files.sh") do |chunk|
+  #       count += chunk.lines.count
+  #       Machinery::Ui.progress("Found #{count} changed files...")
+  #     end
+  def run_inspect_script(*script, &callback)
```
cornelius: What about calling the method `run_script_with_progress`? That would make it clearer why there is an additional method.

lib/ui.rb
```diff
@@ -79,7 +79,15 @@ def print(output)
...
 
+      def progress(output)
+        return if !$stdout.tty?
+
+        reset_line
+        print output
+      end
+
       def puts(output)
+        reset_line if $stdout.tty?
```
cornelius: This adds a reset of the line for every `puts` call, even if the line doesn't need to be cleaned. Would it be possible to instead clean the line at the end of the progress output?

cornelius:
I like the implementation and the output. That gives a better impression about what Machinery is doing when it takes some time.

It would be nice to have a bit more test coverage for this code. We talked about this in person. Let's see if we can come up with a way how to improve that.

PR #966, opened 2015-06-11
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150609_inspection_progress_outpu'

aduffeck:
  * 7e2d9da Add changelog entry for the progress output
  * dd080da Add integration tests for the file inspection progress output
  * 29f7c16 Show progress information while inspecting the file scopes
  * 2768f15 Add Machinery::Ui.progress() for progress info, e.g. when inspecting
  * ae1a817 Add Inspector#run_script_with_progress

### DIFF
--- a/lib/ui.rb
+++ b/lib/ui.rb
@@ -49,7 +49,7 @@ module Machinery
       end
 
       def print(output)
-        reset_line if progress_enabled?
+        reset_line if progress_visible?
 
         if !use_pager || !$stdout.tty?
           begin
@@ -84,7 +84,7 @@ module Machinery
       def progress(output)
         return if !progress_enabled?
 
-        reset_line
+        reset_line if progress_visible?
         print output
 
         @progress_visible = true
@@ -108,9 +108,11 @@ module Machinery
         $stdout.tty? || ENV["FORCE_MACHINERY_PROGRESS_OUTPUT"]
       end
 
-      def reset_line
-        return if !@progress_visible
+      def progress_visible?
+        @progress_visible
+      end
 
+      def reset_line
         STDOUT.print "\r"     # Move cursor to beginning of line
         STDOUT.print "\033[K" # Clear line content
         @progress_visible = false
--- a/lib/ui.rb
+++ b/lib/ui.rb
@@ -113,7 +113,7 @@ module Machinery
       end
 
       def reset_line
-        STDOUT.print "\r"     # Move cursor to beginning of line
+        STDOUT.print "\r" # Move cursor to beginning of line
         STDOUT.print "\033[K" # Clear line content
         @progress_visible = false
       end
--- a/lib/ui.rb
+++ b/lib/ui.rb
@@ -105,7 +105,7 @@ module Machinery
       private
 
       def show_progress
-        $stdout.tty? || ENV["SHOW_PROGRESS_OUTPUT"]
+        $stdout.tty? || ENV["FORCE_MACHINERY_PROGRESS_OUTPUT"]
       end
 
       def reset_line
--- a/lib/ui.rb
+++ b/lib/ui.rb
@@ -49,7 +49,7 @@ module Machinery
       end
 
       def print(output)
-        reset_line if show_progress
+        reset_line if progress_enabled?
 
         if !use_pager || !$stdout.tty?
           begin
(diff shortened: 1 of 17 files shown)


lib/ui.rb
```diff
@@ -90,6 +101,20 @@ def warn(s)
       def error(s)
         STDERR.puts s
       end
+
+      private
+
+      def show_progress
+        $stdout.tty? || ENV["SHOW_PROGRESS_OUTPUT"]
```
tgoettlicher: Wouldn't it make sense to have a machinery specific prefix to the environment variable in order to avoid clashes with other programs: e.g. `MACHINERY_SHOW_PROGRESS_OUTPUT`.

tgoettlicher: I'd expect setting `SHOW_PROGRESS_OUTPUT=false` on a tty would suppress progress output.

aduffeck: I'll rename the variable.

This is not a variable that should be used by users, though, and it will not be documented. It's merely used for forcing the progress output to be shown in integration tests where there is no tty present. The negated version does not really make sense. Maybe we should call it `FORCE_MACHINERY_PROGRESS_OUTPUT` instead to make that more clear.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -90,7 +90,11 @@ def extract_unmanaged_files(files, trees, excluded, store_name)
     archive_path = File.join(store_path, "files.tgz")
     @system.create_archive(files.join("\0"), archive_path, excluded)
 
+    extracted_count = files.length
+
     trees.each do |tree|
+      phrase = Machinery::pluralize(extracted_count, "file or tree", "files and trees")
```
cornelius: I would not use an intermediate variable here. That makes it harder to see what the text would look like.

lib/ui.rb
```diff
@@ -49,6 +49,8 @@ def kill_pager
       end
 
       def print(output)
+        reset_line if show_progress
```
cornelius: This resets the line with every `print` call regardless if progress is actually shown or not. On one hand `show_progress` is misleading, it should probably rather be something like `progress_enabled?`. On the other hand it might be better to only reset the line, if progress actually has been shown and the line needs cleaning.

spec/unit/ui_spec.rb
```diff
@@ -91,4 +96,23 @@
       Machinery::Ui.warn("foo")
     end
   end
+
+  describe ".reset_line" do
+    it "sends reset escape sequences if there progress output that needs to be cleared" do
+      allow($stdout).to receive(:tty?).and_return(true)
+      allow(Machinery::Ui).to receive(:use_pager).and_return(false)
+      expect(STDOUT).to receive(:print).exactly(4).times # 1 for progress, 1 for print, 2 for reset
```
cornelius: Wouldn't it be better to just explicitly check for the four calls and their exact arguments?

spec/unit/ui_spec.rb
```diff
@@ -91,4 +96,23 @@
...
+
+      Machinery::Ui.progress("some_progress")
+      Machinery::Ui.print("")
+    end
+
+    it "does not send unneccessary reset escape sequences if there's nothing to reset" do
+      allow($stdout).to receive(:tty?).and_return(true)
+      allow(Machinery::Ui).to receive(:use_pager).and_return(false)
+      expect(STDOUT).to receive(:print).once
```
cornelius: Same here. Wouldn't it be better to check that the actual argument is correct?

lib/ui.rb
aduffeck: Ok, I'll rename the method. It's not true that the line is always reset, the check whether it's necessary or not is inside of `reset_line`.

plugins/unmanaged_files/unmanaged_files_inspector.rb
aduffeck: So you'd prefer

```ruby
      Machinery::Ui.progress(" -> Extracted #{extracted_count} " \
        "#{Machinery::pluralize(extracted_count, "file or tree", "files and trees")}...")
```
over
```ruby
      phrase = Machinery::pluralize(extracted_count, "file or tree", "files and trees")
      Machinery::Ui.progress(" -> Extracted #{extracted_count} #{phrase}...")
```

No super strong feelings here, but I tend to disagree.

cornelius:
Beautiful. Ship it. :fireworks:

PR #970, opened 2015-06-11
File access api

cornelius:
This pull request is for reviewing the file access API. It's not meant to directly be merged.

Access to the scope file storage is abstracted through the `ScopeFileAccess` class, which is obtained through the scope object. There are two different backends for storing scope file data and according to scope the right one is returned from the scope.

This pull requests consolidates the API to consolidate the different classes accessing scope file into one API.

There still are a couple of places which use no abstraction at all and directly work on the file system, e.g. `AutoYasT#apply_unmanaged_files` or `UnmanagedFilesInspector#extract_tar_metadata`. These places can all be abstracted in the same way going through the `ScopeFileAccess` class. It might need additional methods to fit all use cases, though.

When the API from the application's point of view is ok, the next steps would be to remove redundant, obsolete abstractions, add tests for `ScopeFileAccess` and abstract the places, where there is no abstraction yet.

Unit tests work, I haven't run integration tests. If there are integration tests failing, we should probably add unit tests to capture these cases, so we have full test coverage of the scope file access.

### DIFF
--- a/lib/kiwi_config.rb
+++ b/lib/kiwi_config.rb
@@ -124,7 +124,8 @@ class KiwiConfig < Exporter
       FileUtils.mkdir_p(destination, mode: 01777)
       filter = "unmanaged_files_#{@name}_excludes"
 
-      @system_description.unmanaged_files.export_files_as_tarballs(destination)
+      @system_description.unmanaged_files.file_access.
+        export_files_as_tarballs(destination)
 
       FileUtils.cp(
         File.join(Machinery::ROOT, "export_helpers/#{filter}"),
--- a/lib/kiwi_config.rb
+++ b/lib/kiwi_config.rb
@@ -108,8 +108,7 @@ class KiwiConfig < Exporter
           @sh << "chmod #{file.mode} '#{file.name}'\n"
           @sh << "chown #{file.user}:#{file.group} '#{file.name}'\n"
         elsif file.file?
-          file_access = @system_description[scope].file_access
-          file_access.write_file(file, output_root_path)
+          @system_description[scope].write_file(file, output_root_path)
           @sh << "chmod #{file.mode} '#{file.name}'\n"
           @sh << "chown #{file.user}:#{file.group} '#{file.name}'\n"
         elsif file.link?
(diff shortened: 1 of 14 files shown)


lib/scope_file_access.rb
```diff
@@ -0,0 +1,65 @@
+class ScopeFileAccess
+  def initialize(scope, scope_file_store)
+    @scope = scope
+    @scope_file_store = scope_file_store
+  end
+end
+
+class ScopeFileAccessFlat < ScopeFileAccess
+  def retrieve_files_from_system(system, paths)
+    system.retrieve_files(paths, @scope_file_store.path)
```
thardeck: Shouldn't this part be moved over to have one file access class?

lib/kiwi_config.rb
```diff
@@ -124,7 +125,8 @@ def inject_extracted_files(output_location)
       FileUtils.mkdir_p(destination, mode: 01777)
       filter = "unmanaged_files_#{@name}_excludes"
 
-      @system_description.unmanaged_files.export_files_as_tarballs(destination)
+      @system_description.unmanaged_files.file_access.
+        export_files_as_tarballs(destination)
```
aduffeck: Ok, so you're consolidating what was spread across the `FileScope` objects and `SystemFileUtils`, which is nice, but what's the reason of having a separate object for the file access. Couldn't that functionality be included in the `FileScope` instances directly as modules? 

That would simplify the consumer code and get rid of the `if self.is_a?(UnmanagedFilesScope) ... else ... end` conditional in `Scope#file_access`.

So 
```ruby
file_access = @system_description[scope].file_access
file_access.write_file(file, output_root_path)
# and
@system_description.unmanaged_files.file_access.export_files_as_tarballs(destination)
```

would become
```ruby
@system_description[scope].write_file(file, output_root_path)
# and
@system_description.unmanaged_files.export_files_as_tarballs(destination)
```

cornelius: This would not provide any separation between the two ways how files are stored, and there would not be any encapsulation by name space. The `self.is_a?` is a shortcut for now. This in the end should come from the scopes itself.

For me the most important step is to collect all the code accessing the files in one place. We then still can see what is the best way to make the API accessible.

aduffeck: I think I'm not following. The separation would be have two different modules, just like the two classes we have right now. 
The encapsulation by name space (I think you're referring to the `.file_access` part?) is what I don't like, I think the file scopes could know how to write their files on their own.

aduffeck:
lgtm :+1:

PR #969, opened 2015-06-11
Autoyast export files can contain quotes

vwallfahrer:

PR #972, opened 2015-06-12
Autoyast can handle quotes

vwallfahrer:
Superseds #969


spec/unit/autoyast_spec.rb
```diff
@@ -44,6 +44,27 @@
...
+          mode: "777",
+          user: "root",
+          group: "root",
+          type: "link",
+          target: "/opt/test-quote-char/target-with-quote'-foo"
+        )
+      autoyast = Autoyast.new(description)
+
+      expect(autoyast.profile).to include(
+        "ln -s '/opt/test-quote-char/target-with-quote\\'-foo' '/mnt/opt/test-quote-char/link'"
```
aduffeck: Would it make sense to add  a tick to the filename in the RPM as well? Then we could test that it is handled properly in the link name and the location.

lib/autoyast.rb
```diff
@@ -267,20 +267,26 @@ def apply_changed_files(scope)
 
     @system_description[scope].files.each do |file|
       if file.deleted?
-        @chroot_scripts << "rm -rf '#{file.name}'"
+        @chroot_scripts << "rm -rf '#{Machinery.escape_single_quotes(file.name)}'"
```
aduffeck: That escape call is quite long and used pretty often, but only with `file.name`. What about adding a method to `SystemFile` which returns the quoted name instead? (e.g. `file.quoted_name`)

lib/changed_rpm_files_helper.rb
```diff
@@ -74,14 +74,14 @@ def parse_stat_line(line)
...
 
-    type = case(type)
-    when "directory"
-      "dir"
-    when "symbolic link"
-      "link"
-    when "regular file"
-      "file"
-    end
+    type = case type
```
aduffeck: The whitespacing was correct before I think. I'll try to addapt hound accordingly, but please revert this change.

PR #975, opened 2015-06-15
Enable SimpleCov code coverage analysis

aduffeck:


cornelius:
Could we also use an online service providing code coverage? I'm using coveralls.io in another project, which is quite nice. CodeClimate can also do it, I think.

aduffeck:
Good point, I'll open a new pull request for codeclimate integration.

PR #977, opened 2015-06-15
Do not add "file" type to deleted config or changed-managed files

aduffeck:
We don't know if it was a file or a directory that was deleted.


schema/migrations/migrate3to4.rb
```diff
@@ -24,9 +24,10 @@ class Migrate3To4 < Migration
   def migrate
     if @hash.has_key?("changed_managed_files")
       @hash["changed_managed_files"]["files"].each do |file|
-        file["type"] = if File.directory?(File.join(@path, "changed_managed_files", file["name"]))
+        path = File.join(@path, "changed_managed_files", file["name"])
+        file["type"] = if File.directory?(path)
           "dir"
-        else
+        elsif File.exists?(path)
           "file"
```
thardeck: We should also add a file type if the files weren't extracted.

PR #979, opened 2015-06-15
Do not add "file" type to deleted config or changed-managed files

aduffeck:
We don't know if it was a file or a directory that was deleted.

### DIFF
--- a/spec/unit/migrations/migrate3to4_spec.rb
+++ b/spec/unit/migrations/migrate3to4_spec.rb
@@ -1,0 +1,149 @@
+# Copyright (c) 2013-2014 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or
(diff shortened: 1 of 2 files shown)


spec/unit/migrations/migrate3to4_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+require File.join(Machinery::ROOT, "schema/migrations/migrate3to4")
+
+describe Migrate3To4 do
+  initialize_system_description_factory_store
+
+  let(:description_hash) {
+    JSON.parse(<<-EOT)
+      {
+        "changed_managed_files": {
+          "extracted": false,
```
thardeck: in this case extracted should be activated

thardeck:
lgtm

PR #983, opened 2015-06-15
Autoyast can handle quotes

vwallfahrer:
Superseds #972


thardeck:
The commit needs to be renamed.

lib/autoyast.rb
```diff
@@ -267,20 +267,26 @@ def apply_changed_files(scope)
...
       if file.deleted?
-        @chroot_scripts << "rm -rf '#{file.name}'"
+        @chroot_scripts << "rm -rf '#{quoted(file.name)}'"
       elsif file.directory?
-        @chroot_scripts << "chmod #{file.mode} '#{File.join("/mnt", file.name)}'"
-        @chroot_scripts << "chown #{file.user}:#{file.group} '#{File.join("/mnt", file.name)}'"
+        @chroot_scripts << <<EOF.strip
+chmod #{file.mode} '#{File.join("/mnt", quoted(file.name))}'
+chown #{file.user}:#{file.group} '#{File.join("/mnt", quoted(file.name))}'
+EOF
```
thardeck: I think an EOF is not as readable but it is not that important especially if the line length restriction was the problem.

lib/changed_rpm_files_helper.rb
```diff
@@ -106,9 +106,9 @@ def get_link_target(system, link)
   def get_file_properties(system, cur_files)
     ret = {}
     out = system.run_command(
-        "stat", "--printf", "%a:%U:%G:%u:%g:%F:%n\\n",
-        *cur_files,
-        :stdout => :capture
+      "stat", "--printf", "%a:%U:%G:%u:%g:%F:%n\\n",
+      *cur_files,
+      stdout: :capture
```
thardeck: These beautifications would be better in a separate commit but also not that important.

lib/helper.rb
```diff
@@ -45,6 +45,10 @@ def self.pluralize(count, singular, plural = nil)
 
     val.gsub("%d", count.to_s)
   end
+
+  def self.escape_single_quotes(text)
```
thardeck: This method is obsolete and can be removed.

lib/exporter.rb
```diff
@@ -21,4 +21,8 @@ class Exporter
 
   abstract_method :write
   abstract_method :export_name
+
+  def quoted(name)
```
thardeck: I would prefer `quote` as method name.

lib/kiwi_config.rb
```diff
@@ -103,18 +103,24 @@ def inject_extracted_files(output_location)
...
       @system_description[scope].files.each do |file|
         if file.deleted?
-          @sh << "rm -rf '#{file.name}'\n"
+          @sh << "rm -rf '#{quoted(file.name)}'\n"
         elsif file.directory?
-          @sh << "chmod #{file.mode} '#{file.name}'\n"
-          @sh << "chown #{file.user}:#{file.group} '#{file.name}'\n"
+          @sh << <<EOF
+chmod #{file.mode} '#{quoted(file.name)}'
+chown #{file.user}:#{file.group} '#{quoted(file.name)}'
```
thardeck: I think with the new method name an EOF shouldn't bee needed.

spec/data/changed_managed_files/opensuse131
```diff
@@ -1,6 +1,7 @@
 # Changed managed files [192.168.121.137] (2014-11-25 18:21:52)
 
   Files extracted: yes
+  * /opt/test-quote-char/link (link_path)
```
thardeck: Why is there a link without a single quote? What does it add to our test coverage?

spec/definitions/provision.sh
```diff
@@ -9,6 +9,8 @@ mount --bind /proc /var/lib/chroot_proc
 rpm -i /vagrant/SUSE-test-quote-char-1.0-1.noarch.rpm || rpm -i /vagrant/RedHat-test-quote-char-1.0-1.noarch.rpm
 echo 42 > /opt/test-quote-char/test-dir-name-with-\'\ quote-char\ \'/unmanaged-file-with-\'\ quote\ \'
 mkdir /opt/test-quote-char/test-dir-name-with-\'\ quote-char\ \'/unmanaged-dir-with-\'\ quote\ \'
+rm /opt/test-quote-char/link
+ln -s /opt/test-quote-char/target-with-quote\'-foo /opt/test-quote-char/link
```
thardeck: Shouldn't we use single quotes around the paths?

spec/unit/changed_managed_files_inspector_spec.rb
```diff
@@ -33,10 +33,12 @@
       expect(script.first).to eq("changed_files.sh")
       options[:stdout].puts rpm_result
     end
+
```
thardeck: Why a newline?

spec/unit/changed_managed_files_inspector_spec.rb
```diff
@@ -52,7 +54,6 @@
         subject.inspect(filter)
         expect(description["changed_managed_files"].files.map(&:name)).
           to_not include("/usr/share/man/man1/time.1.gz")
-
```
thardeck: These all are beautifcations again and could be moved to a separate commit.

spec/unit/system_file_spec.rb
```diff
@@ -1,3 +1,4 @@
+
```
thardeck: This should be removed

spec/unit/system_file_spec.rb
```diff
@@ -64,6 +65,16 @@
       type: "remote_dir"
     )
   }
+  let(:single_quote_file) {
+    Machinery::SystemFile.new(
+      name: "/bla/single-quote-file'-foo/"
```
thardeck: This looks like a directory but is called file so I would remove the backslash at the end.

spec/unit/system_file_spec.rb
```diff
@@ -93,6 +104,22 @@
     specify { expect(remote_dir.directory?).to be(false) }
   end
 
+  describe "#quoted_name" do
+    it "returns name with escaped single quotes" do
+      expect(single_quote_file.quoted_name).to eq(
+        "/bla/single-quote-file\\'-foo/"
+      )
```
thardeck: Maybe one example with an existing example path should be added so it is tested that paths without quotes are not changed.

thardeck:
Having the beautifications separated would be nice to have but is not that important.

Renaming the mentioned method does make sense in my oppinion. Lgtm besides that.

PR #985, opened 2015-06-16
Autoyast can handle quotes

vwallfahrer:
Superseds #983


lib/system_file.rb
```diff
@@ -33,6 +33,14 @@ def remote_directory?
...
     end
 
+    def quoted_name
+      name.gsub("'", "\\\\'")
+    end
+
+    def quoted_target
+      target.gsub("'", "\\\\'")
+    end
+
```
thardeck: these methods are obsolete and should be removed

spec/unit/exporter_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+require_relative "spec_helper"
+
+describe Exporter do
+  describe "#quoted" do
```
thardeck: this should be "quote"

PR #990, opened 2015-06-16
Remove without parameters shows help

### LINKED ISSUE #976: machinery remove without args not showing help
When running machinery remove without arguments it should show the help for that subcommand:

```

peon@dhcp150:~> machinery remove
peon@dhcp150:~> 


```

keberle94:
Fixes #976


lib/cli.rb
```diff
@@ -567,7 +572,8 @@ def self.parse_scopes(scope_string)
 
     The success of a removal can be shown with the verbose option.
   LONGDESC
-  arg "NAME..."
+  arg "NAME...", :multiple
```
aduffeck: You should remove the `...` now that we specify the `:multiple` option.

spec/integration/support/command_line_examples.rb
```diff
@@ -108,5 +108,20 @@
         end
       end
     end
+
+    describe "remove" do
+      context "when no arguments given" do
+        it "shows the help message for the remove command" do
```
aduffeck: Do you want to test that the command fails or that the help is shown? Because right now the test does something different than the name says it does.

lib/cli.rb
```diff
@@ -57,11 +58,15 @@ class Cli
 
   GLI::Commands::Help.skips_post = false
 
-  def self.check_exceeding_arguments(defined, parsed)
-    if parsed.size > defined.size
+  def self.validate_number_of_arguments(defined, parsed)
+    if defined.any?(&:multiple?) && parsed.size == 0
```
aduffeck: `parsed.empty?`?

keberle94:
Superseeded by #1007

PR #992, opened 2015-06-16
Check for available online repositories

vwallfahrer:
Superseds #991


lib/analyze_config_file_diffs_task.rb
```diff
@@ -23,6 +23,18 @@ def analyze(description)
       "repositories",
       "config_files"
     )
+    online_repo = 0
+    description["repositories"].each do |repo|
+      if !repo["url"].start_with?("dvd:///") &&
+        !repo["url"].start_with?("cd:///") &&
+        repo["enabled"] == true
+        online_repo =+ 1
```
aduffeck: `=+` seems to be a typo.

lib/analyze_config_file_diffs_task.rb
```diff
@@ -23,6 +23,18 @@ def analyze(description)
       "repositories",
       "config_files"
     )
+    online_repo = 0
+    description["repositories"].each do |repo|
+      if !repo["url"].start_with?("dvd:///") &&
```
aduffeck: We already have a method in the data model which performs this check. I'd rewrite lines 26-36 as

```ruby
if !description["repositories"].any? { |repo| repo.enabled && !repo.external_medium? }
  raise Machinery::Errors::UpgradeFailed.new("No online repositories available.")
end
```

lib/analyze_config_file_diffs_task.rb
```diff
@@ -23,6 +23,18 @@ def analyze(description)
...
+        !repo["url"].start_with?("cd:///") &&
+        repo["enabled"] == true
+        online_repo =+ 1
+      end
+    end
+    if online_repo == 0
+      raise Machinery::Errors::UpgradeFailed.new("No online repositories available.")
```
aduffeck: `Machinery::Errors::UpgradeFailed` doesn't seem to be an appropriate exception class for that.

spec/unit/analyze_config_file_diffs_task_spec.rb
```diff
@@ -101,6 +101,83 @@
...
+            "alias": "dvd_entry_alias",
+            "name": "dvd_entry",
+            "type": "yast2",
+            "url": "dvd:///?devices=/dev/disk/by-id/ata-Optiarc_DVD+_-RW_AD-7200S,/dev/sr0",
+            "enabled": true,
+            "autorefresh": false,
+            "gpgcheck": true,
+            "priority": 2
+          }
+        ], "config_files": {
```
aduffeck: I think you could omit the `config_files` data here, it's not used anyway.

lib/analyze_config_file_diffs_task.rb
aduffeck: I'd maybe use a friendlier message, e.g. `Can not analyze the system description because it does not contain any online repository` or so.

PR #993, opened 2015-06-16
Fix faulty escaping of single quotes

thardeck:
There is no escape character in single quote strings so our approach
didn't work.

Instead I am just replacing a single quote by '\'' .


aduffeck:
lgtm

PR #994, opened 2015-06-16
Fix build validation error message

thardeck:
The content was moved from the man page to the help output some time
ago.


vwallfahrer:
lgtm

Issue #995, opened 2015-06-16
Improve error message when system description isn't found


tgoettlicher:
https://github.com/SUSE/machinery/blob/master/lib/manifest.rb#L27

This error message should:

- be phrased in active voice.
- print single quotes around `#{name}`.

mauromorales:
Fixed

PR #999, opened 2015-06-16
Improve error message

mauromorales:


thardeck:
Have you checked that we use single quotes everywhere else instead of double quotes?

mauromorales:
@thardeck not everywhere but at least in this specific file we aren't. There is still issue https://github.com/SUSE/machinery/issues/1000 so we could look at that there I think.

tgoettlicher:
lgtm

mauromorales:
closes https://github.com/SUSE/machinery/issues/995

PR #1002, opened 2015-06-16
  Check for available online repositories 

vwallfahrer:
Superseds #998


aduffeck:
lgtm (but some squashing would be nice)

Issue #997, opened 2015-06-16
Improve error message when too many arguments are passed


tgoettlicher:
https://github.com/SUSE/machinery/blob/master/lib/cli.rb#L62

This error message should be precise on which arguments are expected and which arguments were received but not expected.

PR #1004, opened 2015-06-17
Add specific command arguments into error message

mauromorales:
When passing more arguments than expected, show the number that were
passed and the ones that are expected from the command.

Issue #997


aduffeck:
lgtm in general, but I would also add a test for the case where the command expects some arguments.

I think this will conflict with #990, so maybe you should coordinate the merging of these two pull requests.

mauromorales:
@aduffeck thanks, will add the test and coordinate with Kevin to merge this

PR #1005, opened 2015-06-18
Adapt to the pennyworth change of failing on stderr output by default

aduffeck:


spec/integration/integration_spec_helper.rb
```diff
@@ -31,8 +31,8 @@ def prepare_machinery_for_host(system, ip, opts = {})
 end
 
 def prepare_local_machinery_for_host(system, ip)
-  system.run_command("ssh-keygen -R #{ip}")
-  system.run_command("ssh-keyscan -H #{ip} >> ~/.ssh/known_hosts")
+  system.run_command("ssh-keygen -R #{ip}", fail_on_stderr_output: false)
+  system.run_command("ssh-keyscan -H #{ip} >> ~/.ssh/known_hosts", fail_on_stderr_output: false)
```
cornelius: This code is confusing. Running a command should fail, if the command fails, not for any other reason. So having to provide an option to get the expected behavior sounds counter-intuitive to me.

spec/integration/support/inspect_unmanaged_files_examples.rb
```diff
@@ -61,7 +62,8 @@ def parse_md5sums(output)
 Inspecting unmanaged-files...
  -> Found 0 files and trees...\r\033\[K -> Found 0 files and trees...\r\033\[K -> Extracted 0 unmanaged files and trees.
 EOF
-      expect(normalize_inspect_output(@machinery_output)).to include(expected)
+      expect(normalize_inspect_output(@machinery_output[0])).to include(expected)
+      expect(@machinery_output[1]).to include("contains invalid UTF-8 characters")
```
cornelius: Indexing the output is not very explicit. I would very much prefer something like `expect(@machinery_output.stdout).to include(expected_stdout)`.

cornelius:
For some inspiration here is a link to tests I wrote for another project: https://github.com/cornelius/cli_tester/blob/master/spec/cli_tester_spec.rb. Maybe we can apply a similar model here.

PR #1008, opened 2015-06-18
Remove command errors without arguments

mauromorales:
superseeds #990


spec/integration/support/command_line_examples.rb
```diff
@@ -120,6 +120,19 @@
           }.to raise_error(Pennyworth::ExecutionFailed, message)
         end
       end
+
+      context "when multiple (undefined) number of arguments are expected" do
+        it "shows the help message for the remove command" do
```
aduffeck: This test still doesn't test that the help is shown, no?

aduffeck:
lgtm besides the confusing test

spec/integration/support/command_line_examples.rb
mauromorales: ah my bad, yup it was ment to say fails with message :/ fixing it now

PR #1010, opened 2015-06-18
Fix sed Gem unshift statement

thardeck:
The sed statement did just search for "gem" to add an additional line
above it.

Since there were three places in the machinery script where gem was
mentioned three lines were added.


aduffeck:
lgtm

PR #1011, opened 2015-06-18
Improve message when running machinery on an unsupported platform

### LINKED ISSUE #1003: Improve message when running machinery on an unsupported platform
https://github.com/SUSE/machinery/blob/master/lib/local_system.rb#L50-L54

This message should:
- have opening quotes, as it has closing quotes
- have the email address easily copyable (without the period at the end of the line).
- explain that there is a config option to disable this message

keberle94:
Fixes #1003

### DIFF
--- a/lib/local_system.rb
+++ b/lib/local_system.rb
@@ -50,7 +50,9 @@ class LocalSystem < System
       message = <<EOF
 You are running Machinery on a platform we do not explicitly support and test.
 It still could work very well. If you run into issues or would like to provide us feedback, you are welcome to file an issue at https://github.com/SUSE/machinery/issues/new or write an email to machinery@lists.suse.com.
-Oficially supported operating systems are: #{supported_oses}"
+Oficially supported operating systems are: '#{supported_oses}'
+
+To disable this message in the machinery configuration use 'machinery config perform-support-check false'
 EOF
 
       Machinery::Ui.warn message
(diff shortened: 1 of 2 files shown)


lib/local_system.rb
```diff
@@ -49,8 +49,10 @@ def validate_machinery_compatibility
       supported_oses = Os.supported_host_systems.map { |o| o.canonical_name }.sort.join(", ")
       message = <<EOF
 You are running Machinery on a platform we do not explicitly support and test.
-It still could work very well. If you run into issues or would like to provide us feedback, you are welcome to file an issue at https://github.com/SUSE/machinery/issues/new or write an email to machinery@lists.suse.com.
```
thardeck: I wouldn't remove the period at the end but instead just add a space between the mail address and the period.

spec/unit/local_system_spec.rb
```diff
@@ -133,10 +133,12 @@
 
       LocalSystem.validate_machinery_compatibility
 
+      supported_oses = Os.supported_host_systems.map(&:canonical_name).sort.join(", ")
+      expect(captured_machinery_output).to include(supported_oses)
       expect(captured_machinery_output).to include(
-        ": SUSE Linux Enterprise Server 12, openSUSE 13.1 (Bottle)," \
```
thardeck: I would check not just for the variable but at least some operating systems which we will be supporting for quite some time, like SLES12 and 13.2

thardeck:
lgtm

PR #1013, opened 2015-06-19
Implementation of  `machinery config foo=bar` syntax

### LINKED ISSUE #1001: `machinery config` should accept `=` when setting values
Right now `machinery config foo bar` is used to set the key `foo` to value `bar`.

`machinery config foo=bar` should be additionally supported.

vwallfahrer:
Fixed #1001


tgoettlicher:
The CLI doesn't complain if you provide too many args:
`./bin/machinery config hints=true superfluous`

lib/cli.rb
```diff
@@ -717,14 +717,20 @@ def self.parse_scopes(scope_string)
   arg "VALUE", :optional
   command "config" do |c|
     c.action do |global_options,options,args|
-      key = args[0]
-      value = args[1]
+      if !args[0].nil? && args[0].include?("=")
```
thardeck: Instead of `!args[0].nil?` you can check for `args[0]`.

lib/cli.rb
```diff
@@ -717,14 +717,20 @@ def self.parse_scopes(scope_string)
...
+        parameter = args[0].split("=")
+        key = parameter[0]
+        value = parameter[1]
```
thardeck: I would use
```
key, value = args[0].split("=")
```

lib/local_system.rb
```diff
@@ -52,7 +52,7 @@ def validate_machinery_compatibility
 It still could work very well. If you run into issues or would like to provide us feedback, you are welcome to file an issue at https://github.com/SUSE/machinery/issues/new or write an email to machinery@lists.suse.com.
 Oficially supported operating systems are: '#{supported_oses}'
 
-To disable this message in the machinery configuration use 'machinery config perform-support-check false'
+To disable this message in the machinery configuration use 'machinery config perform-support-check=false'
```
thardeck: Can you replace `machinery` by `#{$0}` so it gets extended. It is not directly related to your pr but if you change the line anyway.

spec/unit/cli_spec.rb
```diff
@@ -559,6 +559,12 @@
         with("foo", "bar")
       run_command(["config", "foo", "bar"])
     end
+
+    it "handles parameter with equal syntax" do
```
thardeck: s/equal/equal sign/

thardeck:
lgtm

PR #1014, opened 2015-06-19
Adapt to Pennyworth 0.2 syntax/matchers

aduffeck:
Depends on https://github.com/SUSE/pennyworth/pull/87


aduffeck:
The integration tests are failing because the new pennyworth version is not installed on the prophet host yet.

cornelius:
The tests read much nicer now. Excellent job. :clap: 

The only comments I have are in the pennyworth pull request, which might affect this patch as well. So once these are sorted out I'm fine with merging this one as well.

PR #1021, opened 2015-06-22
Fix test

aduffeck:


spec/integration/support/build_examples.rb
```diff
@@ -163,7 +163,7 @@
             "cat '/usr/local/magicapp/weird-filenames/spacy file name'",
             as: "root"
           )
-        ).to succeed.and have_stdout("This is a file with spaces in its name.\n")
+        ).to succeed.with_stderr.and have_stdout("This is a file with spaces in its name.\n")
```
thardeck: I would add a comment why we allow stderr here.

thardeck:
lgtm

PR #1023, opened 2015-06-22
Include integration test examples as per matrix

mauromorales:
This pull request is for discussion only, don't merge.

The idea behind this PR is to modify our test base to run tests depending on the system and architecture it is being run at.

* [x] Use data from matrix
* [ ] separate tests between alfred and machinery
* [ ] define a naming scheme for boxes that can be reused
* [ ] Parse definition to get the host machine
* [ ] Test knows about the test group (full, min, acc)
* [ ] Parametrize the architecture
* [ ] Adjust export
* [ ] Integrate s390 and power  
* [ ] Add tests for deploy group
* [ ] Define in the test matrix what will run on the local tests

https://github.com/SUSE/alfred/blob/updated_new_matrix_representation/documents/source/integration_tests.yml


spec/integration/machinery_opensuse13.1_spec.rb
```diff
@@ -17,7 +17,7 @@
 
 require_relative "integration_spec_helper"
 
-describe "machinery@openSUSE 13.1" do
+describe "machinery@openSUSE 13.1", host: "openSUSE_13_1" do
```
aduffeck: Would it be possible to get rid of the extra argument and extract the host from the example name?

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,48 @@ def normalize_inspect_output(output)
...
+
+def boxes
+  {
+    "openSUSE_13_2" => "opensuse132",
+    "openSUSE_13_1" => "opensuse131"
+  }
+end
+
+def test_according_to_matrix(currently_running_on)
+  path = File.join(Machinery::ROOT, "..", "alfred", "documents", "source", "integration_tests.yml")
```
cornelius: We probably should split up the test matrix in an external and an internal document, so we don't have to cross-reference repositories.

mauromorales: Yes agreed. We said with Chris that this definitely needed to be slitted we just wanted to have a way to share the branch for now.

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,48 @@ def normalize_inspect_output(output)
...
+    "Tumbleweed" => "machinery_tumbleweed"
+  }
+end
+
+def boxes
+  {
+    "openSUSE_13_2" => "opensuse132",
+    "openSUSE_13_1" => "opensuse131"
+  }
+end
```
cornelius: It might be nice to have a naming scheme which is systematic enough to not require extra mappings like the ones above. But if it makes things more explicit and readable for now, this is also something we can fix later.

spec/integration/machinery_opensuse13.1_spec.rb
mauromorales: Yup definitely could be taken from the description

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,48 @@ def normalize_inspect_output(output)
...
+def test_according_to_matrix(currently_running_on)
+  path = File.join(Machinery::ROOT, "..", "alfred", "documents", "source", "integration_tests.yml")
+  matrix = YAML.load_file(path)
+
+  matrix.each do |command|
+    command_name, support_level = command
+    support_level.each do |_type, test_relations|
+      test_relations.each do |relation|
+        relation.each do |host, guests|
+          next unless host.include?(currently_running_on)
```
cornelius: Is there a reason to not do a lookup here based on the system the test runs on instead of iterating through all fields?

spec/integration/integration_spec_helper.rb
mauromorales: Makes lot's of sense, I've added a task to make this change

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,48 @@ def normalize_inspect_output(output)
...
+end
+
+def boxes
+  {
+    "openSUSE_13_2" => "opensuse132",
+    "openSUSE_13_1" => "opensuse131"
+  }
+end
+
+def test_according_to_matrix(currently_running_on)
```
cornelius: A name like `include_examples_for_platform` might be more consistent.

spec/integration/integration_spec_helper.rb
mauromorales: No reason at all, just a first approach, definitely needs to be polished.

spec/integration/machinery_opensuse13.1_spec.rb
mauromorales: Also added it as a task

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,48 @@ def normalize_inspect_output(output)
...
+  path = File.join(Machinery::ROOT, "..", "alfred", "documents", "source", "integration_tests.yml")
+  matrix = YAML.load_file(path)
+
+  matrix.each do |command|
+    command_name, support_level = command
+    support_level.each do |_type, test_relations|
+      test_relations.each do |relation|
+        relation.each do |host, guests|
```
aduffeck: I think that if you make the arrays in the YAML matrix hashes instead it will easier to iterate the matrix, e.g.

```
  inspect:
    supported:
      "sles12:x86_64":
        "sles12:x86_64": full_test
        "sles12:s390x": full_test
        "sles12:power_le": full_test
        "sles11:x86_64": full_test
        "rhel5": full_test
        "rhel6": full_test
```

you should be able to iterate with 

```
test_relations.each do |host, guests|
``` 

directly.

mauromorales: Thanks, updated to use this form now

PR #1026, opened 2015-06-23
Fix integration tests

aduffeck:


aduffeck:
The difference is that in this case we need the output in a variable in order to do some processing on it (`find_command.stdout.split.first.chomp`). That's not possible with the `expect(...).to succeed.and include_stdout` syntax as in the other case.

mauromorales:
got it ... lgtm then

PR #1027, opened 2015-06-23
Add test matrix sources to machinery

mauromorales:
If the tests are running according to the test matrix we need to
separate alfred and machinery only data.


greygoo:
Lgtm

PR #1030, opened 2015-06-23
Show html webserver

aduffeck:
**NOTE:** This is not meant to be merged yet as it's not polished and there is documentation mission etc. It's merely about gathering some feedback about the general approach.


lib/show_task.rb
```diff
@@ -28,12 +28,25 @@ def show(description, scopes, filter, options = {})
...
       LocalSystem.validate_existence_of_package("xdg-utils")
-      Html.generate(description)
-      html_path = SystemDescriptionStore.new.html_path(description.name)
-      LoggedCheetah.run("xdg-open", html_path)
+
+      url = "http://#{options[:ip]}:#{options[:port]}/#{CGI.escape(description.name)}"
+
+      Machinery::Ui.use_pager = false
+      Machinery::Ui.puts <<EOF
+There is a webserver running, serving the description on #{url}.
```
tgoettlicher: Can you remove the period (`.`) to ease copy&paste (similar as in: https://github.com/SUSE/machinery/pull/1030/files#diff-92c36702dc702d09e7599c431e9ad5efR24 )?

lib/cli.rb
```diff
@@ -737,6 +728,26 @@ def self.parse_scopes(scope_string)
...
+  desc "Start a webserver serving an HTML view of a system description"
+  long_desc <<-LONGDESC
+    Starts a web server which serves an HTML view for the given system description.
+  LONGDESC
+  arg "NAME"
+  command "serve" do |c|
+    c.flag [:port, :p], type: Integer, required: false, default_value: 7585,
+      desc: "Listen on port PORT", arg_name: "PORT"
+    c.flag [:ip, :i], type: String, required: false, default_value: "localhost",
+      desc: "Listen on ip IP", arg_name: "IP"
```
tgoettlicher: `Listen on ip address IP`

lib/serve_html_task.rb
```diff
@@ -15,11 +15,19 @@
...
-    Hint.print(:share_html_contents, directory: output)
+class ServeHtmlTask
+  def serve(description, ip, port)
+    url = "http://#{ip}:#{port}/#{CGI.escape(description.name)}"
+
+    Machinery::Ui.use_pager = false
+    Machinery::Ui.puts <<EOF
+The description is now available at #{url}
+
+The webserver can be closed with Ctrl+c.
```
cornelius: It should probably be "web server", or maybe just "server".

cornelius:
Viewing config file diffs doesn't seem to work anymore.

cornelius:
When reloading the page after having used the navigation bar at the top it doesn't go to the same section again (this is an issue in particular because clicking the navigation bar again for the section which was displayed, doesn't do anything).

cornelius:
The general approach looks fine to me.

PR #1034, opened 2015-06-25
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150625_improve_compare_packages_output'

aduffeck:
  * e08356d Handle corner cases when comparing packages scopes
  * 7246e87 Add changelog entry for improved package comparison output
  * f8ca83b Add documentation for the comparison rendering process
  * 2e9b232 Add section to packages comparisons for same-but-different packages
  * 609d581 Extract high-level comparison structure into #do_render_comparison
  * f52f4fb Allow for customizing the rendering of comparison results


spec/unit/packages_renderer_spec.rb
```diff
@@ -51,4 +51,127 @@
...
+              "arch": "x86_64",
+              "vendor": "SUSE LLC <https://www.suse.com/>",
+              "checksum": "091d486ab725d7542933b74f9b3204e4"
+            }
+          ]
+        }
+      EOF
+    }
+
+    it "shows two 'only in x' and on 'in both but different' sections" do
```
cornelius: typo: s/on/one/

lib/renderer.rb
```diff
@@ -73,6 +77,14 @@ def for(scope)
     end
   end
 
+  def scope
+    # Return the un-camelcased name of the inspector,
+    # e.g. "foo_bar" for "FooBarInspector"
+    scope = self.class.name.match(/^(.*)Renderer$/)[1]
+    scope.gsub(/([^A-Z])([A-Z])/, "\\1_\\2").downcase
+  end
```
cornelius: Please add a unit test for this method.

lib/renderer.rb
```diff
@@ -135,13 +145,31 @@ def render_comparison(description1, description2, description_common, options =
...
 
-    render_comparison_only_in(description1, scope)
-    render_comparison_only_in(description2, scope)
-    render_comparison_common(description_common, scope) if options[:show_all]
+    do_render_comparison(description1, description2, description_common)
 
     @buffer
   end
 
+  def render_comparison_section(description, render_method)
```
cornelius: Could this be done with polymorphism instead of passing a method as argument?

cornelius:
The renderer is becoming hard to understand. I would suggest to move the comparison rendering to an own class. This would get rid of of the `comparison` prefixes in method names and better group methods together which are related.

PR #1035, opened 2015-06-25
Machinery helper

cornelius:
This is the prototype for using a helper binary to run an inspection. It only covers unmanaged files for now. It checks, if a helper binary exists for the architecture of the target system, and if one is there, it copies it to the target system and runs it there to do the inspection. If no helper binary is found it transparently falls back to the original method of inspection.

The first iteration of the prototype only covers pure inspection, no file extraction.

The code needs more tests and error handling before it can be merged to master. The basic architecture should be ok, though.

The same concept can be used for inspection of other scopes as well. This will need moving the code injecting and possibly running the binary to a more central place. The basic principle will stay the same, so this can be a later step.

### DIFF
--- a/lib/machinery_helper.rb
+++ b/lib/machinery_helper.rb
@@ -1,3 +1,20 @@
+# Copyright (c) 2015 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of version 3 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
 class MachineryHelper
   attr_accessor :local_helpers_path
 
--- a/lib/machinery_helper.rb
+++ b/lib/machinery_helper.rb
@@ -15,6 +15,18 @@
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
 
+# The MachineryHelper class handles the helper binaries Machinery can use to
+# do inspections. It provides methods to check, if a helper is available, to
+# inject it to the target machine, run it there, and clean up after it's done.
+#
+# The inspection checks, if a binary helper is available on the machine where
+# the inspection is started. It looks at the location
+#
+#    /usr/share/machinery/helpers/<arch>/machinery_helper
+#
+# where <arch> is the hardware architecture of the target system. Valid values
+# are x86_64, i586, s390x, and ppcle.
+
 class MachineryHelper
   attr_accessor :local_helpers_path
 
--- a/lib/machinery_helper.rb
+++ b/lib/machinery_helper.rb
@@ -1,0 +1,41 @@
+class MachineryHelper
+  attr_accessor :local_helpers_path
+
+  def initialize(s)
+    @system = s
+    @arch = nil
+
+    @local_helpers_path = "/usr/share/machinery/helpers"
+  end
+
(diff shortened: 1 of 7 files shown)


lib/machinery_helper.rb
```diff
@@ -0,0 +1,70 @@
...
+    @local_helpers_path = "/usr/share/machinery/helpers"
+  end
+
+  def local_helper_path
+    if !@arch
+      description = SystemDescription.new("", SystemDescriptionMemoryStore.new)
+      inspector = OsInspector.new(@system, description)
+      inspector.inspect(nil)
+
+      @arch = description.os.architecture
```
aduffeck: You can get the local architecture more easily using `LocalSystem.os.architecture`.

aduffeck: Ah, it's about the remote system, not the local one. Nevermind.

lib/machinery_helper.rb
```diff
@@ -0,0 +1,70 @@
...
+  def can_help?
+    File.exist?(local_helper_path)
+  end
+
+  def inject_helper
+    @system.inject_file(local_helper_path, ".")
+  end
+
+  def run_helper(scope)
+    json = @system.run_command("./machinery_helper", stdout: :capture)
```
aduffeck: I think it would be good to take a list of scopes that are inspected in one go (pun intented ;)) and pass it to the helper binary. That would allow the helper to only inspect what's needed and be smart about sharing data between different scopes where appropriate.

lib/remote_system.rb
```diff
@@ -140,4 +140,8 @@ def read_file(file)
       raise
     end
   end
+
+  def inject_file(local_file, remote_path)
+    system("scp #{local_file} root@#{host}:#{remote_path}")
```
aduffeck: Is there a reason for not using Cheetah (or LoggedCheetah) here? And this command and the `run_helper` one need to be changed to consider the `--remote-user` option.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -213,6 +213,18 @@ def inspect(filter, options = {})
 
     scope.scope_file_store = file_store_tmp
 
+    helper = MachineryHelper.new(@system)
+    if helper.can_help?
+      helper.inject_helper
+      helper.run_helper(scope)
+      scope.extracted = false
+      @description["unmanaged_files"] = scope
```
aduffeck: A `remove_helper` call is missing here, no?

Since the helper is ultimately not restricted to one scope I wonder if a scheme where the `System` object holds a reference to its helper would not be more suitable.

How I would imagine it without putting too much thought into it is that the `InspectTask` takes care of installing the helper when possible and removing it after the inspection. The inspectors get the system instance passed anyway, so they could just ask the system if it has a helper or not and decide what to do accordingly:

```ruby
if system.helper
  system.helper.run(scope)
  ...
else
  run_inspection(filter, options, do_extract, file_store_tmp, file_store_final, scope)
end
```

lib/machinery_helper.rb
aduffeck: After looking at the whole pull request I can see know that it's much easier to run one scope at a time, so we should probably defer that optimization to a later refactoring.

lib/machinery_helper.rb
```diff
@@ -0,0 +1,70 @@
...
+    File.join(@local_helpers_path, @arch, "machinery_helper")
+  end
+
+  # Returns true, if there is a helper binary matching the architecture of the
+  # inspected system. Return false, if not.
+  def can_help?
+    File.exist?(local_helper_path)
+  end
+
+  def inject_helper
```
thardeck: I would add a check if the file is already there to allow customers having their own binary there.

lib/remote_system.rb
cornelius: This method is just a stub for the sake of trying the prototype. You are right, that it needs to be implemented properly.

lib/machinery_helper.rb
cornelius: Why would a customer have their own binary there?

cornelius:
Closing this pull request, but keeping the branch around. We can use this as a base for the full implementation.

PR #1036, opened 2015-06-25
Retrieve the system of the host machinery runs on from the rspec metadata.

greygoo:
Adds a method to the spec halper that takes metadata[:description] as argument
and returns the system from it.


mauromorales:
lgtm

spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +64,10 @@ def normalize_inspect_output(output)
     gsub!(/(\r\033\[K.*?\r\033\[K).*\r\033\[K(.*)/, "\\1\\2") # strip all progress messages but two
 end
 
+def machinery_host(description)
+  description.sub!(/.*@/, "")
+end
```
cornelius: Could we add some unit tests for this? This would make it easier to see what is happening, because it would be explicitly expressed in the test, and it will also prevent things from breaking in the future.

greygoo:
closing as will be part of complete PR for matrix based test execution

PR #1038, opened 2015-06-26

### LINKED ISSUE #348: Man page not accessible when installing from gem

keberle94:
Fixes #348


keberle94:
Not ready for review

man/machinery-man.1.md
```diff
@@ -0,0 +1,10 @@
+
+## man — Shows man page
+
+### SYNOPSIS
+
+  `machinery [global options] man` 
+
+### DESCRIPTION
+
+The `man` command shows the machinery man page.
```
tgoettlicher: `Machinery` is written in other man pages with an upper-case `M` because Machinery is a name.

man/machinery-man.1.md
```diff
@@ -0,0 +1,10 @@
+
+## man — Shows man page
+
+### SYNOPSIS
+
+  `machinery [global options] man`
```
tgoettlicher: Which global options can be passed?

NEWS
```diff
@@ -3,6 +3,7 @@
 
 ## Version 1.9.1 - Tue Jun 23 16:57:35 CEST 2015 - thardeck@suse.de
 
+* Added `man` command to show the machinery man page
```
tgoettlicher: Please use present tense => `Add`.

lib/man_task.rb
```diff
@@ -0,0 +1,31 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+class ManTask
+  def man
+    begin
```
tgoettlicher: You could replace this construct by `LocalSystem.validate_existence_of_package("man")`.

man/machinery-man.1.md
```diff
@@ -0,0 +1,10 @@
+
+## man — Shows man page
```
tgoettlicher: I'd write it in upper-case because it's a headline.

keberle94:
Squashed by #1049

PR #1039, opened 2015-06-29
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150625_improve_compare_packages_output'

aduffeck:
  * 5937d08 Refactor Renderer API
  * f0eb2b0 Change PackagesScope#compare_with to consider changed packages
  * c468536 Add an (optional) field for changed elements to #compare_with
  * 9b917ab Fix typo
  * 5b0a26d Do not consider release attribute when the version has changed
  * 88397b9 Fix test, it was not testing what it said it tests
  * c92fdd9 Fix hound issues
  * a93f99d Expect if clauses spanning multiple lines to be indented, not aligned
  * e08356d Handle corner cases when comparing packages scopes
  * 7246e87 Add changelog entry for improved package comparison output
  * f8ca83b Add documentation for the comparison rendering process
  * 2e9b232 Add section to packages comparisons for same-but-different packages
  * 609d581 Extract high-level comparison structure into #do_render_comparison
  * f52f4fb Allow for customizing the rendering of comparison results

### DIFF
--- a/lib/compare_task.rb
+++ b/lib/compare_task.rb
@@ -74,17 +74,18 @@ class CompareTask
         partial_description_common = SystemDescription.new(
           "common",
           store,
-          scope => comparison[2]
+          scope => comparison[3]
         )
 
         output += Renderer.for(scope).render_comparison(
           partial_description1,
           partial_description2,
+          comparison[2],
           partial_description_common,
           options
         )
 
-        if partial_description1[scope] || partial_description2[scope]
+        if partial_description1[scope] || partial_description2[scope] || comparison[2]
           identical = false
         else
           identical_scopes << scope
--- a/lib/renderer.rb
+++ b/lib/renderer.rb
@@ -101,24 +101,32 @@ class Renderer
     @buffer
   end
 
-  def render_comparison_section(description)
+  def render_comparison_section(description, render_method)
     @system_description = description
-    indent { do_render }
+    indent { render_method.call }
     @buffer += "\n" unless @buffer.empty? || @buffer.end_with?("\n\n")
   end
 
   def render_comparison_only_in(description, scope)
     if description[scope]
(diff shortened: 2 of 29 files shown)


lib/compare_task.rb
```diff
@@ -74,17 +74,18 @@ def render_comparison(description1, description2, scopes, options = {})
...
           "common",
           store,
-          scope => comparison[2]
+          scope => comparison[3]
         )
 
         output += Renderer.for(scope).render_comparison(
           partial_description1,
           partial_description2,
+          comparison[2],
```
cornelius: This is inconsistent in terms of the API. It would be more logical to also wrap the changed elements in a description, so that the call would use the arguments `partial_description1, partial_description2, partial_description_changed, partial_description_common`.

lib/renderer.rb
```diff
@@ -94,54 +117,85 @@ def render(system_description, options = {})
...
-  end
-
-  def render_comparison(description1, description2, description_common, options = {})
+  # Renders the result of a comparison of two system descriptions. The arguments to that method
+  # are three SystemDescriptions:
+  #
+  # [description1] contains all elements that are only present in description1
+  # [description2] contains all elements that are only present in description2
+  # [description_common] contains all elements that are present in both descriptions
+  #
```
cornelius: The API documentation also needs to document the argument with the changed elements.

cornelius:
Maybe it would be good to change the return value of the `compare_with` method to be a dedicated object with named accessors to the different parts of the comparison result. Just having a plain array makes the calling code not very easy to read as you need to know on which position which part is.

cornelius:
I think introducing the new naming convention for the methods to be overridden by the specific renderers makes sense. There still might be ways how to improve that, but at least the basic structure is clear now.

PR #1040, opened 2015-06-29
Add package 'man' to machinery hosts

keberle94:
- Add 'man' package to openSUSE 13.2
- Add 'man' package to openSUSE Tumbleweed

- Package 'man' is needed on the machinery hosts to test the new machinery man function

PR #1041, opened 2015-06-29
Add package 'man' to machinery integration-test hosts

keberle94:
Superseeds #1040 

- Add 'man' package to openSUSE Tumbleweed

- Package 'man' is needed on the machinery integration-test hosts to test the new machinery man function


thardeck:
I would only test Tumbleweed and SLE12, so 13.2 needs to be removed

thardeck:
lgtm

PR #1046, opened 2015-06-30
Add tool SupportMatrix

mauromorales:
spreadsheet


cornelius:
In general this looks good to me. We need to fix the unit tests of course.

One thing: could we use a much smaller file for the tests of the tool itself? It should be enough to test the key cases, not the full data, shouldn't it?

mauromorales:
@cornelius good point, I'll shorten those data files

PR #1049, opened 2015-06-30

### LINKED ISSUE #348: Man page not accessible when installing from gem

keberle94:
Fixes #348 
Squashes #1038


spec/unit/man_task_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+      expect(LocalSystem).to receive(:validate_existence_of_package).with("man")
+      allow_any_instance_of(Kernel).to receive(:system).with("man", subject.man_path)
+
+      subject.man
+    end
+
+    it "calls the machinery man page" do
+      allow(LocalSystem).to receive(:validate_existence_of_package).with("man")
+      expect(subject).to receive(
+        "system").with("man", subject.man_path)
```
tgoettlicher: Wouldn't it read easier with the line break before `with`?
``expect(subject).to receive("system").
  with("man", subject.man_path)``

lib/cli.rb
```diff
@@ -566,6 +565,18 @@ def self.parse_scopes(scope_string)
...
+    Shows the man page of the machinery tool.
+
+  LONGDESC
+  command :man do |c|
+    c.action do
+      task = ManTask.new
+      task.man
+    end
+  end
+
```
tgoettlicher: Could you remove the superfluous blank lines?

man/machinery-man.1.md
```diff
@@ -0,0 +1,10 @@
+
+## Man — Shows man page
```
tgoettlicher: I wasn't precise enough what I meant when I said it's a head line:  `## man — Shows Man Page`.

keberle94: Sorry haven't looked properly into it.

keberle94:
Squashed by #1053

PR #1052, opened 2015-06-30
Run groups of tests defined in a integration test yaml file

greygoo:
The inspect, build, analyze, export and deploy test_examples are now run based on entries in alfred/documents/source/integration_tests.yml.
The tests to be executed can be selected via the shell environment variable TESTGROUP that needs to be set to "full", "acceptance" or "minimal". If the variable is empty, the full test set is executed to keep the default behaviour of the previous test execution method. Test groups are inclusive, full tests include acceptance and minimal tests and minimal tests include acceptance tests.


spec/integration/integration_spec_helper.rb
```diff
@@ -64,6 +65,54 @@ def normalize_inspect_output(output)
...
+def test_group_hierarchy 
+  {
+    "full" => %W(full_test minimal_test acceptance_test),
+    "minimal" => %W(minimal_test acceptance_test),
+    "acceptance" => %W(acceptance_test)
+  }
+end
+
+def include_examples_for_platform(currently_running_on)
+  path = File.join(Machinery::ROOT, "..", "alfred", "documents", "source", "integration_tests.yml")
```
cornelius: The tests in the machinery repository should not require or reference the ones from the alfred repository.

PR #1053, opened 2015-06-30

### LINKED ISSUE #348: Man page not accessible when installing from gem

keberle94:
Fixes #348 
Squashes #1049

### DIFF
--- a/man/machinery-man.1.md
+++ b/man/machinery-man.1.md
@@ -1,0 +1,10 @@
+
+## man — Shows Man Page
+
+### SYNOPSIS
+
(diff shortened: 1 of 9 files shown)


man/machinery-man.1.md
```diff
@@ -0,0 +1,10 @@
+
+## man — Shows Man page
```
tgoettlicher: I'd use an upper-case `P` in `page`.

tgoettlicher:
lgtm

gsanso:
Tested.

PR #1059, opened 2015-07-01
Show hint how to turn on hints again immediately after deactivation

### LINKED ISSUE #1057: Hints for `machinery config hint` is broken
Hints for `machinery config hint` is broken. 

See:
```
# machinery config hints on
hints = true
# machinery config hints off
hints = false
# machinery config hints off
hints = false
Hints can be switched on again by '/usr/bin/machinery config hints=on'.
```
It should show the hint if and only if the value of hints is toggled.

### LINKED ISSUE #1058: machinery config --help should mention `KEY=VALUE` syntax
`machinery config --help` prints:
```
NAME
    config - Show or change machinery's configuration

SYNOPSIS
    machinery [global options] config [KEY] [VALUE]

DESCRIPTION
    Show or change machinery's configuration.

    The value of a key is shown when no value argument is passed. If neither
    the key argument nor the value argument are specified a list of all keys and
    their values are shown. 
```

It should also mention that the syntax `KEY=VALUE` is also valid.

keberle94:
Fixes #1057

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -742,7 +742,7 @@ class Cli
       task = ConfigTask.new
       task.config(key, value)
 
-      if key == "hints" && !@config.hints
+      if key == "hints" && value == "false"
         Machinery::Ui.puts "Hints can be switched on again by '#{$0} config hints=on'."
       end
     end
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -742,7 +742,7 @@ class Cli
       task = ConfigTask.new
       task.config(key, value)
 
-      if key == "hints" && value == "false"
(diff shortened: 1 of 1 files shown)


lib/cli.rb
```diff
@@ -742,7 +742,7 @@ def self.parse_scopes(scope_string)
       task = ConfigTask.new
       task.config(key, value)
 
-      if key == "hints" && !@config.hints
+      if key == "hints" && value == "false"
```
tgoettlicher: Does this also work with `hints=off` instead of `hints=false`?

keberle94:
No you are right. I forgot about that way. Now it should.

tgoettlicher:
lgtm

PR #1060, opened 2015-07-01
Add tool SupportMatrix

mauromorales:
spreadsheet

Supperseeds #1048


thardeck:
`rake -T` fails even if I run bundle install because of missing rodf

thardeck:
The create ods should be also added to .gitignore

mauromorales:
I've added a rescue clause to deal with the problem in the Rakefile. Once the changes are merged to the original gem I can get rid of this and do it like with any other gem.

The created file is created in the .machinery folder so no need to add to the gitignore.

cornelius:
Could you rebase the pull request on master? At the moment it doesn't find the yaml files as they are not merged to the branch.

mauromorales:
@cornelius added them

cornelius:
Looks nice and works well.

The spreadsheet should have some legend to explain what the content of the cells mean, the colors and the "M", "F", "A" letters.

Rakefile
```diff
@@ -177,3 +182,23 @@ task :upgrade_test_descriptions do
...
+
+      You can solve this issue by:
+
+        1) running `bundle exec rake matrix:spreadsheet` or
+        2) by installing the gems binstubs `bundle install --binstubs`
+
+    EOF
+  end
+  sources = File.join(Machinery::ROOT, "spec", "definitions", "support")
+  file = SupportMatrix.new(sources, OdsFormatter.new).write(Machinery::DEFAULT_CONFIG_DIR)
```
cornelius: I wouldn't write the file to the config dir. It would make more sense to write it to the directory where the yaml files are stored as well. Maybe we should even check it in, so that it can be accessed when browsing the project in GitHub (and linked from the Wiki page).

cornelius:
Looks good to me now.

PR #1062, opened 2015-07-02
Filter RHEL5 non-deterministic unmanaged file

thardeck:
/var/log/mcelog was on some occurences there and sometimes not so we
filter it to prevent failing tests.


mauromorales:
lgtm

PR #1065, opened 2015-07-02
Hide data which is common in both descriptions in html view by default

keberle94:


keberle94:
Seems not to work for special scopes yet. Will open it again when it does.

PR #1071, opened 2015-07-02
[Review] Request from 'aduffeck' @ 'SUSE/machinery/show_html_webserver'

aduffeck:
  * b3ced13 Fix hound issues
  * 009ddab Add/update documentation for `machinery show --html` and `machinery serve`
  * 43e28dc Fix services title help on mouse-over
  * 4e7406b Fix typos
  * 6a7ac22 Wait for server to come up before opening the browser
  * 127e0b5 Fix config file diff links not working
  * 513dd0b Fix scope anchors not working on page reload
  * 236ef2c Warn when web server is reachable from the network
  * c9066e2 Fix tests
  * 6bbe79f Fix hound issues
  * 2670d29 Add CLI tests for "show --html" and "serve"
  * de25722 Implement "machinery serve NAME" command for serving HTML views
  * 0f1be2a Get rid of the generate-html task,it no longer works with the web server
  * 2410a04 Replace static HTML by a Sinatra webserver serving all descriptions


man/machinery-serve.1.md
```diff
@@ -0,0 +1,41 @@
...
+    Use specified system description.
+
+
+### OPTIONS
+
+  * `-p PORT`, `--port=PORT` (optional):
+    Specify the port on which the web server will serve the system description: Default: 7585
+
+  * `-i IP`, `--ip=IP` (optional):
+    Specify the IP address on which the web server will be made available. Default: 127.0.0.1
```
tgoettlicher: The documentation says `127.0.0.1` but the actual default value is `localhost`. That shouldn't make a difference if name resolution works the standard way. We should be consistent, though.

PR #1075, opened 2015-07-02
Run integration tests through test matrix

greygoo:


cornelius:
The code looks fine. I haven't run the tests to check, if they are covering the right things, though.

PR #1078, opened 2015-07-03
Add tool SupportMatrix

mauromorales:
spreadsheet

Superseeds #1060


mauromorales:
see lgtm in #1060

PR #1080, opened 2015-07-03
Add #binary to ScopeFileAccess which checks the file type

tgoettlicher:


lib/scope_file_access.rb
```diff
@@ -30,6 +30,16 @@ def file_content(filename)
 
     File.read(path)
   end
+
+  def binary?(filename)
```
cornelius: Shouldn't the method be the other way around and check, if a file is a text file? That is what is actually doing, so naming it `text_file?` might be more accurate.

cornelius: The parameter is a file name, but all the files which are checked are part of a system description. So analogous to the `write_file` method they could take a `SystemFile` object as parameter.

The same is true for the `file_content` methods. I didn't realize it when I reviewed them, but it would be good to make all methods in the scope file access API consistent.

cornelius:
As the two scope file access classes kepp getting longer, I would suggest to split them up and put them in two separate files. I would not do it as part of this pull request, though, but as a separate one afterwards.

tgoettlicher:
The method is called `binary?` because we will make use of it to decide if we display a file or offer it for download. When it comes to empty files it's easier to go with the following logic:
ASCII => show
empty => show
binary => download

That's why the function checks if a file is binary and returns `false` for empty files. 

Another option would be to check if a file is a text file and return `true` for empty files. I wouldn't call an empty file necessarily a text file, though.

tgoettlicher:
https://github.com/SUSE/machinery/pull/1086

PR #1081, opened 2015-07-03
Add rspec_junit_formatter to Gemfile

mauromorales:
currently we run this code and get this error in the unit tests

    rspec.ruby2.1 spec/unit -f RspecJunitFormatter ...

which returns an error for the gem that is installed with github

     require': cannot load such file -- odf/spreadsheet (LoadError)

running with bundle exec fixes the problem but the gem `rspec_junit_formatter` needs to be included in the Gemfile


cornelius:
lgtm

PR #1086, opened 2015-07-06

tgoettlicher:


lib/scope_file_access.rb
```diff
@@ -17,18 +17,18 @@ def file_path(system_file)
...
     if !extracted
       raise Machinery::Errors::FileUtilsError, "The requested file '#{filename}' is not available" \
         " because files for scope '#{scope_name}' were not extracted."
     end
 
-    path = File.join(scope_file_store.path, filename)
-    if !File.exist?(path)
-      raise Machinery::Errors::FileUtilsError, "The requested file '#{filename}' was not found."
-    end
+    File.read(system_file.scope.file_path(system_file))
```
aduffeck: Since the module is included directly in the `Scope` object this can also be `File.read(file_path(system_file))` instead.

lib/scope_file_access.rb
```diff
@@ -79,17 +79,23 @@ def tarball_path(system_file)
...
     rescue
-      raise Machinery::Errors::FileUtilsError, "The requested file '#{filename}' was not found."
+      raise Machinery::Errors::FileUtilsError,
+        "The requested file '#{system_file.name}' was not found."
     end
   end
+
+  def binary?(system_file)
+    output = Cheetah.run("file", "-", stdin: file_content(system_file), stdout: :capture)
+    !output.include?("ASCII") && !output.include?("no read permission")
```
aduffeck: Machinery should always be able to access its filestore, no? I think I'd rather treat the `no read permission` case as an error.

tgoettlicher:
Squashed: https://github.com/SUSE/machinery/pull/1090

PR #1087, opened 2015-07-06
Hide data which hasn't changed per scope in html view by default

keberle94:


html/assets/compare/machinery.js
```diff
@@ -50,4 +50,114 @@ $(document).ready(function () {
     $(this).hide();
     $("#collapse-all").show();
   });
+
+  // Show or hide elements which are common in scope
+  $("#show-common-elements-in-packages").click(function(){
```
mauromorales: Here I think you could have generic functions that are triggered by these buttons having certain classes, something in the lines of:

    $(".show-common-elements").click(function(){
        $(".scope_common_content").collapse("show");
        $(this).hide();
        $(".hide-common-elements").show();
      });

And not sure if it's the case here but maybe you could use jQuery toggle instead of having show/hide

html/assets/machinery.css
```diff
@@ -24,6 +24,12 @@ a.both_anchor {
   visibility: hidden;
 }
 
+.scope_content{
```
mauromorales: this could be done in one line like: border-bottom: 1px dotted #D58512;

keberle94:
Squashed in #1089

PR #1090, opened 2015-07-07

tgoettlicher:


tgoettlicher:
Squashed: https://github.com/SUSE/machinery/pull/1091

PR #1091, opened 2015-07-07

tgoettlicher:


aduffeck:
As discussed we need to handle cases where the content can not be delivered although the scope was extracted, e.g. because it's a link or a directory or because it is a deleted file. We can merge that pull request as it is and tackle those issues separately, though, from my pov.

PR #1095, opened 2015-07-07
export_task raises error if files aren't extracted

### LINKED ISSUE #1054: `machinery export-autoyast` should complain if files are not extracted
When `machinery export-autoyast` is run on a system description which doesn't contain extracted files it should complain.

vwallfahrer:
Fix #1054


lib/export_task.rb
```diff
@@ -24,6 +24,13 @@ def export(output_dir, options)
     @exporter.system_description.assert_scopes("os")
     @exporter.system_description.validate_export_compatibility
 
+    ["unmanaged_files", "changed_managed_files", "config_files"].each do |scope|
+      if @exporter.system_description[scope] &&
+        !@exporter.system_description.scope_extracted?(scope)
+          raise Machinery::Errors::ExportFailed.new("No extracted files in scope #{scope}")
```
aduffeck: There's already a `MissingExtractedFiles` exception class which shows some help about how to extract the missing files, maybe you could also use it here.

aduffeck:
lgtm in general.

I think Hound's right about the indentation, though, and maybe we can use the other exception class.

PR #1097, opened 2015-07-07
export_task raises error if files aren't extracted 

### LINKED ISSUE #1054: `machinery export-autoyast` should complain if files are not extracted

vwallfahrer:
Superseds #1095 
Fixes #1054


aduffeck:
lgtm

PR #1099, opened 2015-07-08
Comparison output is meaningful when one descriptions has excluded scopes

### LINKED ISSUE #1029: HTML-Output of the comparison is not meaningful when one descriptions has not extracted file
The HTML View contains no meaningful information when you are trying to compare a system description with another one, in which the files are not extracted. The output can be seen in the attached picture.

There should be information about the exact status of the files, on whether they are extracted or not, for each description. A hint that in this case the comparison is not possible would be also helpful for the User.

This issue came up during a usability test of the comparison views.

html-output_of_comparison_without_extracted_files_of_one_description

keberle94:
Fixes #1029


html/comparison.html.haml
```diff
@@ -15,6 +15,20 @@
...
+            %span.dismiss{:title => "Collapse/Expand"}
+        .row
+          .row
+            .col-xs-1
+            .col-xs-7
+              Couldn't compare <strong>'{{diff.meta.uninspected[diff.meta.description_a].join(", ")}}'</strong> because they are not present in description <strong>'{{diff.meta.description_a}}'</strong>
+          .row
+            .col-xs-1
+            .col-xs-7
+              Couldn't compare <strong>'{{diff.meta.uninspected[diff.meta.description_b].join(", ")}}'</strong> because they are not present in description <strong>'{{diff.meta.description_b}}'</strong>
```
aduffeck: The two sections should only be shown if there is something to, otherwise you can end up with something like this:

image

mauromorales: good catch :)

aduffeck: Maybe we can tweak the appearance a bit, e.g. like so:

image

html/comparison.html.haml
```diff
@@ -29,7 +43,7 @@
 
     %script#scope_os.partial{:type => "text/ng-template"}
       %a.scope_anchor{:id => "os"}
-      #os_container.scope
+      #os_container.scope{"ng-show" => "diff.os"}
```
aduffeck: Do we still need those show conditions now that the scopes are only included if they exist in the diff anyway?

html/comparison.html.haml
mauromorales: Sure thing

html/comparison.html.haml
mauromorales: I guess we don't ^__^

PR #1104, opened 2015-07-08
Remove curl from base images as it is only required in machinery images.

greygoo:
- The addition to the base images is not required by the serve html
  testcase and thus they are removed again to avoid sideffects with other
  testcases.


vwallfahrer:
lgtm

PR #1105, opened 2015-07-08
Expand common elements section in each scope on click of the 'both' link

keberle94:


mauromorales:
To me the auto scrolling when clicking on "both" is very unnatural. If I click there and the entire screen scrolls and moves, as a user I have no idea what just happened (specially if it hides the title of the section). In theory I think it's a good idea to jump to the anchor but when using it, it doesn't feel good.

html/assets/compare/machinery.js
```diff
@@ -67,4 +67,11 @@ $(document).ready(function () {
     $scope.find(".show-common-elements").show();
     return false;
   });
+
+  $(".both-show-common-elements").click(function(){
```
mauromorales: I would not use a "both-show-common-elements" but modify "show-common-elements" so it can handle both at the same.

keberle94:
Superseeded by #1123

PR #1110, opened 2015-07-09
Improve error message when export-autoyast/kiwi command options are missing

### LINKED ISSUE #1055: Improve error message when command options are missing
The error message shown when command options are missing should be improved.

Example: `# machinery export-kiwi foo` prints `kiwi-dir is required`. 

It should say `Option --kiwi-dir` is required.

vwallfahrer:
Superseds #1109 
Fixes #1055


thardeck:
lgtm

PR #1111, opened 2015-07-09
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150709_show_diffs_between_unmanaged_files'

aduffeck:
  * 9e54acc Add changelog entry for unmanaged files diff
  * f91a7b8 Only show "Diff files" link if there are files to diff
  * ecba4dd Make diffs of unmanaged files available in the HTML view
  * c9ea073 Port the 'compare' HTML view to the web server approach like 'show'
  * b36ef63 Raise exception if binary files are diffed
  * c720306 Add FileDiff class for diffing extracted files
  * f222113 Make byebug available in the tests

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -4,6 +4,7 @@
 * Improve display of package differences in compare command
 * Add `man` command to show the Machinery man page
 * Content of extracted files can be displayed in HTML view
+* Differences between unmanaged files can be viewed in the HTML comparison
 
 ## Version 1.9.1 - Tue Jun 23 16:57:35 CEST 2015 - thardeck@suse.de
 
--- a/lib/file_diff.rb
+++ b/lib/file_diff.rb
@@ -1,0 +1,28 @@
+# Copyright (c) 2013-2015 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of version 3 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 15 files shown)


lib/file_diff.rb
```diff
@@ -0,0 +1,32 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+class FileDiff
+  def self.diff(description1, description2, scope, path)
```
thardeck: Shouldn't we address this if this is a new method?

NEWS
```diff
@@ -4,6 +4,7 @@
 * Improve display of package differences in compare command
 * Add `man` command to show the Machinery man page
 * Content of extracted files can be displayed in HTML view
+* Differences between unmanaged files can be viewed in the HTML comparison
```
thardeck: this needs to be adapted to the new release

thardeck:
I can't really judge the HTML part but the rest looks good to me except of the comments. But we really need some tests for it in the future, it gets way to complex.

thardeck:
I like the Diff view, but the button to get might be not that easy to find, if you not explicitly search for it.

lib/file_diff.rb
aduffeck: The method is pretty short and straightforward, I don't really see how a refactoring could make it more readable or understandable. This metric probably is not that meaningful here because of the different prerequisite checks we need to do (scope extracted, file exists, file binary).

thardeck:
lgtm

Issue #1050, opened 2015-06-30
Inspection on Debian - Could not establish SSH connection to host


gsanso:
Machinery should show a warning at the beginning when it is run on a non supported platform. I inspected a Debian 8.1 box but the ssh connection couldn't be established although the ssh keys setup is correct.

```
peon@dhcp110:~/dev/machinery/bin> ./machinery inspect 10.100.98.86
Inspecting 10.100.98.86 for os, packages, patterns, repositories, users, groups, services, config-files, changed-managed-files, unmanaged-files...

Note: There are filters being applied during inspection. (Use `--verbose` option to show the filters)


Could not establish SSH connection to host '10.100.98.86'. Please make sure that you can connect non-interactively as root, e.g. using ssh-agent.

To copy your default ssh key to the machine run:
ssh-copy-id root@10.100.98.86



peon@dhcp110:~/dev/machinery/bin> ssh root@10.100.98.86

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
Last login: Thu Jun 18 16:41:02 2015 from 10.100.103.45

The programs included with the Debian GNU/Linux system are free software;
the exact distribution terms for each program are described in the
individual files in /usr/share/doc/*/copyright.

Debian GNU/Linux comes with ABSOLUTELY NO WARRANTY, to the extent
permitted by applicable law.
```

tgoettlicher:
Fix: https://github.com/SUSE/machinery/pull/1113

tgoettlicher:
Fix: https://github.com/SUSE/machinery/pull/1114

tgoettlicher:
Fixed.

PR #1114, opened 2015-07-09
Fix connection errors on debian inspections (gh#SUSE/machinery#1050)

tgoettlicher:


thardeck:
Can you add an explanation to the commit message?

PR #1116, opened 2015-07-09
Fix connection errors on debian inspections (gh#SUSE/machinery#1050)

tgoettlicher:
On debian systems the connection check failed with 'stdin is not a tty'.
To fix this we don't read from stdin but run the noop command(:).


thardeck:
lgtm

PR #1117, opened 2015-07-09
Fix connection errors caused by reading from stdin (gh#SUSE/machinery…

tgoettlicher:
…#1050)

On certain systems (e.g. debian) the connection check failed with 'stdin is not a tty'.
To fix this we don't read from stdin but run the noop command(:).


thardeck:
lgtm

PR #1118, opened 2015-07-10
Make global "--exclude" option a command option for show and inspect

### LINKED ISSUE #900: Global options are not shown in help for command
When looking for a global option to a command users do something like `machinery help show`. This doesn't list the global options such as `--exclude`. This can make users not find the global options at all.

aduffeck:
Fixes #900


mauromorales:
lgtm

PR #1121, opened 2015-07-10
Alert when missing scopes while comparing in html

mauromorales:
Superseeds #1099


html/comparison.html.haml
```diff
@@ -292,7 +303,7 @@
                   %a#diff-unmanaged-files{"data-toggle" => "modal", "data-target" => "#diff-unmanaged-files"}
                     Diff files
         .row.scope_content.collapse.in
-          .row
+          .row{"ng-show" => "diff.unmanaged_files.only_in1 && diff.unmanaged_files.only_in2"}
```
aduffeck: Why is that section only shown if there are file only in 1 and others only in 2? You'd also want to see it if there's only on e of those, no?

mauromorales:
@aduffeck yup, it should show when any of the two has changes

aduffeck:
lgtm

PR #1122, opened 2015-07-10
Add tumbleweed version check

thardeck:


spec/integration/machinery_opensuse_tumbleweed_spec.rb
```diff
@@ -136,5 +137,27 @@
         end
       end
     end
+
+    describe "check if the latest version is run" do
+      it "as base image" do
```
mauromorales: I personally would prefer to do the separation of these as context (ie 'when validating a base image', 'when validating tumbleweed') and then the it could say something like it matches the release number or sumething like that

mauromorales:
lgtm besides my comment

mauromorales:
Lgtm

PR #1125, opened 2015-07-13
Rearrange the exclude option in the inspect and build test.

greygoo:
- We made the exclude option a inspect specific option and not global
  option. Thus the inspect test needs to be adjusted.


aduffeck:
lgtm

PR #1126, opened 2015-07-13
Fix parsing of exclude option in inspect and show

aduffeck:


greygoo:
lgtm from a problem fixing pov, did checkout the branch and the excluded package is in fact excluded now.

aduffeck:
Weird. They run fine locally and also the prophet output doesn't show any errors. I'll fix the hound issues and see if the tests work then.

PR #1132, opened 2015-07-20
Check for necessary packages

### LINKED ISSUE #349: Show all packages needed to be installed for deploy in one go

vwallfahrer:
Fixes #349


spec/unit/build_task_spec.rb
```diff
@@ -44,6 +44,7 @@
 
   describe "#build" do
     it "stores the kiwi config to a temporary directory" do
+      allow(LocalSystem).to receive(:validate_existence_of_packages).and_return("")
```
cornelius: The result of the call is not checked, so it's not needed to return anything here. Same for the other calls.

lib/compare_task.rb
```diff
@@ -27,7 +27,7 @@ def compare(description1, description2, scopes, options = {})
   end
 
   def render_html_comparison(description1, description2, scopes, options)
-    LocalSystem.validate_existence_of_package("xdg-utils")
+    LocalSystem.validate_existence_of_packages(["xdg-utils"])
```
cornelius: It would be better to allow to just pass a name and not require to do that as array even if it's only one package.

The `validate_existence_of_packages` method could just internally convert the parameter to an array if it's passed as string.

lib/local_system.rb
```diff
@@ -29,17 +29,32 @@ def os
...
+                  raise(Machinery::Errors::MissingRequirement.new("You need the package '#{package}' from module '#{needed_module}'. You can install it as follows:\n" \
+                    "If you haven't selected the module '#{needed_module}' before, run `yast2 scc` and choose 'Select Extensions' and activate '#{needed_module}'.\nRun `zypper install #{package}` to install the package.\n\n"))
+                end
+              end
+            end
+          end
+
+      if !missing_packages.empty?
+        missing_packages.each do |missing_package|
+          error_string += "You need the package '#{missing_package}'. You can install it by running `zypper install #{missing_package}`\n"
```
cornelius: It would be better to put out one message for installation of all packages instead of a separate one for each package.

It also would be good to have a unit test, which checks the error message. This would also make it easier to see what the code does, when the unit test demonstrates how the expected output should look like.

PR #1139, opened 2015-07-20
Check for necessary packages

### LINKED ISSUE #349: Show all packages needed to be installed for deploy in one go

vwallfahrer:
Superseds #1132 
Fixes #349


cornelius:
What about keeping the option to pass a single package name as a parameter? Maybe it actually would be nice to keep the original method and just let it call the new one wrapping the package name in an array.

PR #1145, opened 2015-07-21
Allow copy and paste for config options

thardeck:
If you run `machinery config` and just copy the option the command
fails because there are spaces between the key and the value.


keberle94:
lgtm

keberle94:
lgtm

keberle94:
lgtm

PR #1146, opened 2015-07-21
Fix merge error introduced in bde551e

### LINKED ISSUE #1143: compare --html diff doesn't work anymore
If I start the "Diff files" overlay I directly get the message:
```
There was an unknown error downloading the file.
```

If I then select a file it hangs with the message
```
Generating diff. Please wait...
```

aduffeck:
Fixes #1143


keberle94:
lgtm

cornelius:
Looks like we do need better tests for the Javascript.

PR #1149, opened 2015-07-21
Check for necessary packages

### LINKED ISSUE #349: Show all packages needed to be installed for deploy in one go

vwallfahrer:
Fixes #349
Superseds #1148


cornelius:
`validate_existence_of_package` should call `validate_existence_of_packages` instead of duplicating the code.

PR #1160, opened 2015-07-22
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150722_fix_show_html'

### LINKED ISSUE #1150: Machinery show --html is not working when file-scopes are excluded.
When running ```machinery show DESCRIPTION --html``` on a description, which has excluded file-scopes,
bug-extracted-excluded-file-scopes
 the html-view shows no content at all.

aduffeck:
  * 2b7e66f Do not choke on anonymous Inspector subclasses
  * da9bf5e Add changelog entry for bug fix #1150
  * e67abfe Fix HTML view when some file scopes were not extracted
  * d77fbb1 Add tests for web server class
  * 3906391 Extract our sinatra app into its own class so that it can be tested
  * 87c401c Fix setting up the file structure of extracted scopes in the factory


thardeck:
lgtm besides my comments

aduffeck:
Discussed those comments with the team and we decided to merge it like it is.

PR #1171, opened 2015-07-27
Machinery serve does respect config port setting from config-file

keberle94:
Fixes #1170

PR #1174, opened 2015-07-28

keberle94:
Supersedes #1171


lib/cli.rb
```diff
@@ -769,7 +770,9 @@ def self.supports_filtering(command)
 
       description = SystemDescription.load(name, system_description_store)
       task = ServeHtmlTask.new
-      task.serve(description, options[:ip], options[:port])
+      task.serve(
+        description, options[:ip], options[:port] || Machinery::Config.new.http_server_port
```
aduffeck: Isn't `options[:port]` always set since we provide a `default_value`?

spec/integration/support/serve_examples.rb
```diff
@@ -62,7 +65,28 @@
...
+
+    it "makes the system description HTML available at the config-file port" do
+      @machinery.inject_directory(
+        system_description_dir,
+        machinery_config[:machinery_dir],
+        owner: machinery_config[:owner],
+        group: machinery_config[:group]
+      )
+
+      with_env("MACHINERY_CONFIG_FILE" => "/tmp/config") do
```
aduffeck: `with_env` sets the local env iirc, so that code only works if the test runs on the local machine, no?

spec/integration/support/serve_examples.rb
```diff
@@ -62,7 +65,28 @@
...
+        @machinery.run_command("#{machinery_command} config http-server-port=7500")
+      end
+
+      cmd = "#{machinery_command} serve opensuse131"
+      Thread.new do
+        @machinery.run_command(cmd)
```
aduffeck: And why isn't the `MACHINERY_CONFIG_FILE` environment variable necessary here?

spec/integration/support/serve_examples.rb
thardeck: this file should be cleaned up later on and it is better to use a more uncommon file

keberle94:
Superseded by #1176

PR #1176, opened 2015-07-29

keberle94:
Supersedes #1174


keberle94:
Superseded by #1177

PR #1177, opened 2015-07-29

keberle94:
Supersedes #1176


aduffeck:
The integration tests don't work for me. They just hang here:

```
local
  serve html
    makes the system description HTML and extracted files available at the specified port
    makes the system description HTML available at the config-file port
```

aduffeck:
lgtm (modulo squashing)

keberle94:
Squashed by #1178

PR #1179, opened 2015-07-29
Capture Ambiguous Option Exception and handle it properly

### LINKED ISSUE #1152: --extract leads to error
```
# machinery inspect --extract hostname
Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
/usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_block_parser.rb:66:in `do_parse': ambiguous option: --extract (OptionParser::AmbiguousOption)
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_block_parser.rb:33:in `parse!'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_parser.rb:136:in `parse!'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_parser.rb:17:in `block in parse_options'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_parser.rb:14:in `tap'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/gli_option_parser.rb:14:in `parse_options'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.11.0/lib/gli/app_support.rb:73:in `run'
        from ./bin/machinery:28:in `<main>
```

keberle94:
Fixes #1152


mauromorales:
lgtm

mauromorales:
lgtm just clean the hound issues

lib/cli.rb
```diff
@@ -83,6 +83,9 @@ def self.handle_error(e)
     Machinery::Ui.kill_pager
 
     case e
+    when OptionParser::AmbiguousOption
```
aduffeck: Couldn't this be part of the exception handling in lines 92-97? That way we'd also show the hint about the `--help` option.

mauromorales: Andre is right, it should be showing up so you can find out what those ambiguous options are ... sorry that I didn't think of it :(

spec/unit/cli_spec.rb
```diff
@@ -660,6 +660,23 @@
   end
 
   describe "#error_handling" do
```
mauromorales: Not related to your changes but seems this should actually be ".handle_error" ... maybe we can add it in an additional commit since it's such a small change

spec/unit/cli_spec.rb
```diff
@@ -660,6 +660,23 @@
...
+    context "OptionParser" do
```
mauromorales: I'd prefer this to be a bit more descriptive, if just OptionParser will be mentioned then I feel no need to have the context block. On the other hand if we say why we need to group these tests with something like "for options" (or something like that) then it gives more context:

    #error_handling
        shows stderr, stdout and the backtrace for unexpected errors
        OptionParser
          shows a proper error-message for [OptionParser::MissingArgument]

compared to:

    .handle_error
        shows stderr, stdout and the backtrace for unexpected errors
        for options
          shows a proper error-message for [OptionParser::MissingArgument]

mauromorales:
lgtm

keberle94:
Squashed and rebased by #1193

PR #1182, opened 2015-07-30
[Review] Request from 'aduffeck' @ 'SUSE/machinery/integrate_machinery_helper'

aduffeck:
  * 54457d0 Adapt and clean up tests
  * 648ec6a Fix hound issues
  * e48aeec Pipe STDERR output from the helper binary to the console
  * 046a283 Make sure to remove the helper binary after the inspection
  * eb8c76c Also apply the filters when using the helper
  * 346431e The helper returns the "unmanaged_files" content directly at the moment
  * b30a3b6 Only use the inspection helper if it's appropriate
  * b96be49 machinery_helper is now called machinery-helper
  * 1eb8a76 Make use of System#arch instead of figuring out the arch on our own
  * 3265c74 Implement #inject_file and #remove_file for the System classes
  * 7c7711d Move architecture detection from OsInspector to System
  * 759c6a6 Document scheme of helper binary lookup
  * 2eab081 Add license header
  * 5534b5e Never use machinery_helper in existing tests
  * d36465e Use MachineryHelper in unamanaged files inspection
  * 7eaec98 Add dummy machinery_helper for testing
  * 41bf955 Add method to inject a file to target system
  * b919416 Add class to handle helper binary for inspection
  * 713ed58 Merge pull request #1178 from SUSE/use_in_config_defined_port6
  * bd4da9b Ensure usage of config-path specified by the environment variable MACHINERY_CONFIG_FILE
  * cfe451f Get rid of FakeFS in config_spec
  * b240a20 Extraxt common elements to an higher level (describe-block)
  * ec35c8e Write seperate test for the creation of the right directory
  * f9d89c8 Convert define_entries to abstract method
  * 5ed9291 Allow config path change during initialization
  * febb3e0 Port independent 'show' unit test
  * bffa654 Use cofig-file-defined port when given instead of default
  * 883b0a6 Extract 'test basic html' functionality to a method


lib/machinery_helper.rb
```diff
@@ -0,0 +1,62 @@
...
+  end
+
+  # Returns true, if there is a helper binary matching the architecture of the
+  # inspected system. Return false, if not.
+  def can_help?
+    File.exist?(local_helper_path)
+  end
+
+  def inject_helper
+    @system.inject_file(local_helper_path, ".")
```
tgoettlicher: Is it on purpose that we don't copy the helper binary to a tmp directory but to '.'?

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -213,6 +212,57 @@ def inspect(filter, options = {})
...
+
+      scope.files.delete_if { |f| file_filter.matches?(f.name) }
+
+      @description["unmanaged_files"] = scope
+    else
+      run_inspection(file_filter, options, do_extract, file_store_tmp, file_store_final, scope)
+    end
+  end
+
+  def helper_usable?(helper, options)
```
tgoettlicher: `helper_usable?` looks like a method which only returns a boolean. It's not obvious that it also writes to stdout.

spec/integration/support/inspect_unmanaged_files_examples.rb
```diff
@@ -63,6 +63,7 @@ def parse_md5sums(output)
 
       expected = <<EOF
 Inspecting unmanaged-files...
+Note: Using traditional inspection because there is no helper binary for architecture 'x0_0' available.
```
tgoettlicher: `architecture 'x0_0'` is kind of irritating. Is there a way to improve `normalize_inspect_output` to not mangle the architecture string?

spec/unit/local_system_spec.rb
```diff
@@ -93,6 +93,31 @@
     end
   end
 
+  describe "#inject_file" do
+    it "works" do
```
tgoettlicher: `works` is too generic.

spec/unit/local_system_spec.rb
```diff
@@ -93,6 +93,31 @@
...
+
+      expect(File.exists?(expected_file)).to be(false)
+
+      local_system.inject_file(file, destination)
+      expect(File.exists?(expected_file)).to be(true)
+    end
+  end
+
+  describe "#remove_file" do
+    it "works" do
```
tgoettlicher: Same here.

spec/unit/machinery_helper_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    it "can not help if helper does not exist" do
+      helper = MachineryHelper.new(dummy_system)
+      helper.local_helpers_path = given_directory
+
+      expect(helper.can_help?).to be(false)
+    end
+  end
+
+  it "#inject_helper" do
```
tgoettlicher: I'd expect `describe '#inject_helper' do` here.

spec/unit/machinery_helper_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    helper.local_helpers_path = "ab/cd"
+    remote_path = "."
+
+    expect(dummy_system).to receive(:inject_file).with(local_helper_path,
+      remote_path)
+
+    helper.inject_helper
+  end
+
+  it "#remove_helper" do
```
tgoettlicher: Same here.

spec/unit/machinery_helper_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+    helper = MachineryHelper.new(dummy_system)
+
+    expect(dummy_system).to receive(:remove_file).with("./machinery-helper")
+
+    helper.remove_helper
+  end
+
+  it "#run_helper" do
```
tgoettlicher: And here.

plugins/unmanaged_files/unmanaged_files_inspector.rb
aduffeck: @tgoettlicher can you think of a better name?

spec/integration/support/inspect_unmanaged_files_examples.rb
aduffeck: I played around with the regex for a while but it became ugly to a degree where I think it's not worth it anymore. Are you fine with leaving it like it is?

lib/machinery_helper.rb
aduffeck: It was already like that in cornelius' original branch. @cornelius can you comment on that?

spec/integration/support/inspect_unmanaged_files_examples.rb
tgoettlicher: If it's too much work it's not worth the effort. A comment explaining that `normalize_inspect_output` adapts the text which leads to `x0_0` could help future developers when they'll read the tests.

plugins/unmanaged_files/unmanaged_files_inspector.rb
tgoettlicher: Unfortunately not. That's why the code wasn't in a separate method but part of the inspect method in my branch. If you also cannot come-up with a better name, I'm fine with it as it is right now.

tgoettlicher:
LGTM now.

PR #1184, opened 2015-07-30
Add parent directories of managed dirs to managed dir list

thardeck:
The original unmanaged-file inspector was changed to not mark
directories as unmanaged if they only consist managed sub directories.

### DIFF
--- a/plugins/unmanaged_files/unmanaged_files_inspector.rb
+++ b/plugins/unmanaged_files/unmanaged_files_inspector.rb
@@ -55,8 +55,8 @@ class UnmanagedFilesInspector < Inspector
     end
     # make sure that all parent directories of managed rpm directories are considered
     # managed
-    dirh.dup.each do |d, _e|
-      dir, _sep, _file = d.rpartition("/")
+    dirh.dup.keys.each do |d|
+      dir = d.rpartition("/").first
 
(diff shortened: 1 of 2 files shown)


cornelius:
I agree with the intention of the patch but I haven't checked the code or tested it.

tgoettlicher:
It looks good to me, especially because there are tests.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -53,6 +53,17 @@ def extract_rpm_database
         files[pair.first] = pair[1]
       end
     end
+    # make sure that all parent directories of managed rpm directories are considered
+    # managed
+    dirh.dup.each do |d, _e|
```
cornelius: I would do `dirh.dup.keys.each do |d|`, then you don't need an unused variable.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -53,6 +53,17 @@ def extract_rpm_database
...
+      dir, _sep, _file = d.rpartition("/")
```
cornelius: I would do `dir = d.rpartition("/").first`, then you don't need unused variables.

cornelius:
lgtm

PR #1194, opened 2015-08-03
Normalize architectures, so that tests can be used on different machines

aduffeck:


thardeck:
lgtm

thardeck:
lgtm but I would wait for the tests

PR #1195, opened 2015-08-03
Add /var/lock/ to filter file

thardeck:
The machinery_helper was filtering /var/lock but not machinery itself.
Since it does not make sense to inspect /var/lock we added it to the
filter list.


aduffeck:
lgtm

PR #1196, opened 2015-08-03
Machinery config --help mentions alternative `KEY=VALUE` syntax

keberle94:
Enhances #1058


lib/cli.rb
```diff
@@ -729,6 +729,10 @@ def self.supports_filtering(command)
 
     The value of a key is shown when no value argument is passed.
     If neither the key argument nor the value argument are specified a list of all keys and their values are shown.
+
+    ALTERNATIVE SYNOPSIS:
+
+    machinery [global options] config [KEY]=[VALUE]
```
tgoettlicher: Square bracket mean the attribute is optional. As there are no square brackets around the `=` is means the `=` is mandatory.
I'd say it need to read as follows:
`machinery [global options] config [KEY][=VALUE]`
What do you think. Is this understandable to the users?

keberle94: Yes you are right. That should be clearer for the user. I'll change so.

tgoettlicher:
LGTM

PR #1203, opened 2015-08-06
Separate examples that run on specific servers

mauromorales:
Make a different call for examples starting vagrant machines and from
the ones that connect to running servers


tgoettlicher:
LGTM

PR #1205, opened 2015-08-07
Show debug information for unexpected unmanaged file data

thardeck:
Related to https://bugzilla.suse.com/show_bug.cgi?id=939220 to get proper error output and make the code more robust.


tgoettlicher:
Why don't you add one line above accessiong the `os` object which checks if the `os` object is `nil`. In that case print the data (`file.inspect`) to the log. The next line will access `os.user` and make it crash. 
Using this approach we wouldn't have special handling of errors.

tgoettlicher:
Anyway, I could also live with your implementation.

Issue #1188, opened 2015-07-31
Document architecture requirements for `machinery build`


tgoettlicher:
Building a SLES11SP3 i586 fails on an openSUSE 13.2 x86_64 with the message which refers to the support matrix. The Matrix doesn't mention the architecture requirements but only the operating systems (e.g. SLES11SP3  is supported on openSUSE 13.2). This can lead to confusion. That's why we should be clear on the architecture requirements for `machinery build`.

PR #1207, opened 2015-08-10
Added help for building architectures

vwallfahrer:
#1188


lib/local_system.rb
```diff
@@ -87,7 +87,8 @@ def validate_build_compatibility(system_description)
         message = "Building '#{system_description.os.display_name}' is " \
           "not supported on this distribution.\n" \
           "Check the 'BUILD SUPPORT MATRIX' by running `#{$0} build --help` for " \
-          "further information which build targets are supported."
+          "further information which build targets are supported.\n" \
+          "You are only able to build the architecture you are running."
```
thardeck: I would output the current architecture so people know which one it is.

PR #1209, opened 2015-08-10
Added help for building architectures

### LINKED ISSUE #1188: Document architecture requirements for `machinery build`

vwallfahrer:
Superseds #1208 
Fixes #1188


thardeck:
lgtm

PR #1223, opened 2015-08-17
Mariadb container association

mauromorales:


lib/cli.rb
```diff
@@ -787,6 +787,25 @@ def self.supports_filtering(command)
     end
   end
 
+  if @config.experimental_features
+    desc "Dockerize a system description"
+    long_desc <<-LONGDESC
+      Detects workloads from a system description and creates docker containers.
```
cornelius: I would say "... and creates a recommendation for a corresponding docker container setup".

lib/workload_mapper.rb
```diff
@@ -0,0 +1,87 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+class WorkloadMapper
+  attr_reader :system, :workloads, :current_workload
```
cornelius: I would add a blank line to separate the `attr_reader` from the method definition.

lib/workload_mapper.rb
```diff
@@ -0,0 +1,87 @@
...
+    files.each do |file_path|
+      result.merge!(YAML::load(File.read(file_path)))
+    end
+    result
+  end
+
+  def load_clues
+    files = Dir["#{File.expand_path(workload_mapper_path)}/*.rb"]
+    files.each do |file_path|
+      instance_eval(File.read(file_path))
```
cornelius: I think I would separate the DSL from the mapper class, put the methods accessible from the clues into its own context.

spec/unit/workload_mapper_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+      allow(subject).to receive(:workloads).and_return(workloads)
+      allow(subject).to receive(:current_workload).and_return("mariadb")
+
+      subject.instance_eval do
+        parameter "port", "1234"
+      end
+
+      expect(subject.workloads["mariadb"]["parameters"]).to include("port" => "1234")
+    end
+  end
```
cornelius: It would be nice, if the tests for the DSL methods would be less dependent on the implementation. I would imagine the test to more look like the following sketch:

```ruby
  mapper = WorkloadMapper.new

  clue = 'identify "foo"'
  workload = mapper.check_clue(clue)

  expect(workload.name).to eq("foo")
```

lib/dockerize_task.rb
```diff
@@ -0,0 +1,23 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+class DockerizeTask
+  def dockerize(description, dir)
+    workload_mapper = WorkloadMapper.new(description)
+    workload_mapper.write_compose_file(dir)
```
cornelius: I would do that a bit more granular:

```
  mapper = WorkloadMapper.new
  mapper.write_compose_file(workloads)
```

Then we could also easily add some output about what happened and what workloads were identified here.

spec/data/docker/my.cnf
```diff
@@ -0,0 +1,103 @@
...
+
+# [mysqld6]
+# port       = 3309
+# datadir    = /var/lib/mysql-databases/mysqld6
+# pid-file   = /var/lib/mysql-databases/mysqld6/mysql.pid
+# socket     = /var/lib/mysql-databases/mysqld6/mysql.sock
+# user       = mysql
+
+!includedir /etc/my.cnf.d
+
```
cornelius: Do we need the full file here, or would a subset do as well?

lib/workload_mapper.rb
```diff
@@ -0,0 +1,76 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+class WorkloadMapper
+  attr_reader :workloads, :current_workload
+
+  def initialize(&block)
+    @workloads = {}
```
cornelius: Is there any reason to keep the workloads as a member variable. Returning it in `identify_workloads` and using it as an argument for `write_compose_file` would minimize local state. This would make `WorkloadMapper` more flexible and make the code easier to read as there are less implicit assumptions.

lib/workload_mapper.rb
```diff
@@ -0,0 +1,76 @@
...
+  def compose_service(workload, config)
+    name = config["service"]
+    service = {
+      name => compact(look_up_table[workload])
+    }
+    fill_in_template(service[name], config["parameters"])
+    service
+  end
+
+  def look_up_table
```
cornelius: Is there a better name for this method? It doesn't say what it looks up by what criteria, that makes it a bit hard to follow the code.

spec/unit/workload_mapper_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+    it "returns a valid compose node" do
+      allow(subject).to receive(:workloads).and_return(workloads)
+      allow(subject).to receive(:look_up_table).and_return(look_up_table)
+
+      expect(subject.compose_service(workload, config)).to eq(expected_node)
+    end
+  end
+
+  describe "#look_up_table" do
+    it "loads all the .rb files in the mapper path" do
```
cornelius: Does `look_up_table` load Ruby files?

lib/workload_mapper.rb
```diff
@@ -0,0 +1,76 @@
...
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+class WorkloadMapper
+  attr_reader :workloads, :current_workload
+
+  def initialize(&block)
+    @workloads = {}
+    @current_workload = nil
+    instance_eval(&block) if block_given?
```
cornelius: For what is this needed?

workload_mapper/mariadb.rb
```diff
@@ -0,0 +1,6 @@
+mariadb_config = "/etc/my.cnf"
+if system.runs_service?("mysql") && system.has_file?(mariadb_config)
+  identify "mariadb", "db"
```
cornelius: Shouldn't the `"db"` rather be in the `mariadb.yml` file? This is describing the role of the workload. Is this an attribute of the inspected system or is it an attribute of the workload itself?

mauromorales: just for the record: we decided to keep it this way so we can rename the services depending on what's going on for example mongodb can be a `db` service for a workload but considered a `queue` in a workload which uses postgres as it's `db` service

PR #1224, opened 2015-08-17
Changed the color of `both` link to orange

### LINKED ISSUE #1222: `both` link in `compare --html` view doesn't look like a link
The `both` link for similar details of the systems is not identifiable as link.

vwallfahrer:
Fixes #1222

### DIFF
--- a/html/comparison.html.haml
+++ b/html/comparison.html.haml
@@ -115,7 +115,7 @@
                   Changed
                   = ": {{diff.packages.changed.length}} packages"
                 %span.summary-part{"ng-show" => "diff.packages.common"}
-                  %a{:class=>"show-common-elements", href: "#packages_both"}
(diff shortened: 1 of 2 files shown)


html/comparison.html.haml
```diff
@@ -115,7 +115,7 @@
                   Changed
                   = ": {{diff.packages.changed.length}} packages"
                 %span.summary-part{"ng-show" => "diff.packages.common"}
-                  %a{:class=>"show-common-elements", href: "#packages_both"}
+                  %a.both_link{:class=>"show-common-elements", href: "#packages_both"}
```
mauromorales: why does the show-common-elements class needs to be inside the curly braces? couldn't we just use a.show-common-elements.both_link ?

PR #1225, opened 2015-08-18
Mariadb container association2

mauromorales:
Supersedes #1223


cornelius:
`spec/data/mapper/look_up_table.yml` is not needed anymore, isn't it?

Other than that the patch looks fine to me now.

PR #1230, opened 2015-08-20
Changed the color of `both` link to orange

### LINKED ISSUE #1222: `both` link in `compare --html` view doesn't look like a link

vwallfahrer:
Superseds #1224 
Fixes #1222

### DIFF
--- a/html/assets/machinery.css
+++ b/html/assets/machinery.css
@@ -28,6 +28,10 @@ a.both_anchor {
   visibility: hidden;
 }
 
+a.both_link {
+    color: #CC3300;
+}
+
 .scope_content{border-bottom: 1px dotted #d58512}
 
 .scope-summary {
@@ -119,7 +123,7 @@ h2 {
 }
 
 a, a:hover {
-  color: #7A7570;
+  color: #CC3300;
 }
 
 .pad-top {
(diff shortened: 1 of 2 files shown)


cornelius:
I don't think this is an improvement. It makes the links stand out too much. They are not the most important links.

What problem are you trying to solve here?

vwallfahrer:
It's not clear for the user that `both` is a link. What are your suggestions?

aduffeck:
@cornelius What do we do about this pull request?

cornelius:
@aduffeck Can we fix it?

suse-tests-pass:
Prophet reports failure.
( Failure: Merged 8e6ef38f4b0356565d81ba80dd059a6537319e43 into a218099f889fe953c02100d8a6e7bef64a21ec2e )

mschnitzer:
I think it does make much more sense here when we make the text-link to an image-link. We can use the background of https://github.com/SUSE/machinery/blob/master/html/assets/logo-os.png and replace `OS` with `both`. I think that'll look pretty good and fits to the design.

mschnitzer:
As discussed with @cornelius: we'll close that issue. Reason: The change of the color doesn't makes it really better. We'll postpone that.

PR #1233, opened 2015-08-20
Adapt unmanaged-file integration test expectation

thardeck:
The unmanaged-file message has changed because of the newly bundled
helper.


keberle94:
lgtm

PR #1234, opened 2015-08-20
Mariadb workload dockerization

mauromorales:
Integration of the Dockerfiles into our workload mapper


workload_mapper/mariadb/container/scripts/start.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
tgoettlicher: Can you please add `set -e` to make the script fail in cause of an error?

cornelius:
The changes look good to me, but the integration tests need to be fixed before we can merge this.

PR #1235, opened 2015-08-20
Handle case when no workload is detected

tgoettlicher:


tgoettlicher:
Closing because it's already covered by a different commit.

PR #1236, opened 2015-08-21
Added progress and result output to `containerize` command

vwallfahrer:


lib/containerize_task.rb
```diff
@@ -20,5 +20,11 @@ def containerize(description, dir)
     mapper = WorkloadMapper.new
     workloads = mapper.identify_workloads(description)
     mapper.write_compose_file(workloads, dir)
+    if workloads.empty?
+      Machinery::Ui.puts "No workloads detected."
+    else
+      Machinery::Ui.print "\nFound #{workloads.count} #{Machinery::pluralize(workloads.count, "workload")}."
```
cornelius: It should put out the list of workloads it found, not the count. This is much more useful information.

lib/workload_mapper.rb
```diff
@@ -44,7 +44,10 @@ def identify_workloads(system_description)
     Dir["#{File.expand_path(workload_mapper_path)}/*.rb"].each do |file_path|
       mapper = WorkloadMapperDSL.new(system_description)
       workload = mapper.check_clue(File.read(file_path))
-      workloads.merge!(workload.to_h)
+      if workload.to_h
+        Machinery::Ui.puts "Detected workload '#{workload.name}'."
+        workloads.merge!(workload.to_h)
+      end
```
cornelius: There should not be output in this class. This can be handle in the `ContainerizeTask` class.

spec/unit/containerize_task_spec.rb
```diff
@@ -56,5 +57,16 @@
...
+    it "shows the number of detected workloads" do
+      expect_any_instance_of(WorkloadMapper).
+        to receive(:identify_workloads).with(system_description).and_return(workloads)
+      expect_any_instance_of(WorkloadMapper).
+        to receive(:write_compose_file).with(workloads, output_path)
+      containerize_task.containerize(system_description, output_path)
+      expect(captured_machinery_output).to include(
+        "\nFound 1 workload. Wrote to"
+      )
+    end
```
cornelius: For the tests I would rather have one test, which shows how the output looks like and checks that the actual output matches that.

PR #1237, opened 2015-08-21
Added progress and result output to `containerize` command

vwallfahrer:
Superseds #1236


lib/containerize_task.rb
```diff
@@ -20,5 +20,15 @@ def containerize(description, dir)
...
     workloads = mapper.identify_workloads(description)
     mapper.write_compose_file(workloads, dir)
+    if workloads.empty?
+      Machinery::Ui.puts "No workloads detected."
+    else
+      workloads.each do |workload|
+        Machinery::Ui.puts "Detected workload '#{workload[0]}'."
+      end
+      Machinery::Ui.print "\nFound #{workloads.count}" \
+        " #{Machinery::pluralize(workloads.count, "workload")}."
```
cornelius: I don't think this is helpful output. The number of workloads will always be pretty small.

spec/unit/containerize_task_spec.rb
```diff
@@ -56,5 +57,16 @@
...
       containerize_task.containerize(system_description, output_path)
     end
+
+    it "shows the number of detected workloads" do
+      expect_any_instance_of(WorkloadMapper).
+        to receive(:identify_workloads).with(system_description).and_return(workloads)
+      expect_any_instance_of(WorkloadMapper).
+        to receive(:write_compose_file).with(workloads, output_path)
+      containerize_task.containerize(system_description, output_path)
+      expect(captured_machinery_output).to include(
```
cornelius: I would not check that the string is included, but compare against the full output.

Specifying the output with the heredoc syntax makes it easier to read, because it than looks exactly like it will look like to the user.

spec/unit/workload_mapper_spec.rb
```diff
@@ -92,6 +92,7 @@
   end
 
   describe "#identify_workloads" do
+    capture_machinery_output
```
cornelius: This shouldn't be there anymore, should it?

lib/workload_mapper_dsl.rb
```diff
@@ -32,6 +32,7 @@ def parameter(name, value)
   end
 
   def to_h
+    return nil unless service
```
cornelius: Wouldn't it be better to return an empty hash here, so that there are no checks needed when dealing with the results?

There also should be a unit test for a change like this as this is very fundamental API, which should be fully covered by tests.

PR #1246, opened 2015-08-21
Clean up before each package build

thardeck:
To prevent issues with the machinery-helper `osc build` is always run
with the option "--clean".
The reason for the issues were overlapping pathes between the helper rpm
and machinery.


keberle94:
LGTM

keberle94:
LGTM - Tumbleweed passed

PR #1255, opened 2015-08-27
Add call to write readme file in containerize task

mauromorales:


keberle94:
lgtm, besides squashing

PR #1257, opened 2015-08-28
Fix code fragment layout of containerize-readme

keberle94:
With this layout the user would be able to copy and paste codeblocks. But I'm not sure if that's the right layout to provide this code.


mauromorales:
I think it is: http://daringfireball.net/projects/markdown/syntax#precode so ... lgtm

PR #1258, opened 2015-08-28
WorkloadMapper.identify_workload validates extracted scopes

mauromorales:


keberle94:
Is it possible not to create the output directory if we're not able to containerize?

keberle94:
LGTM

PR #1262, opened 2015-08-31
Add `machinery inspect --docker-container` command

vwallfahrer:


plugins/users/users_inspector.rb
```diff
@@ -25,8 +25,10 @@ def initialize(system, description)
 
   def inspect(_filter, _options = {})
     passwd = @system.read_file("/etc/passwd")
-    shadow = @system.read_file("/etc/shadow")
-
+    begin
+      shadow = @system.read_file("/etc/shadow")
+    rescue => e
```
cornelius: Don't silently ignore exceptions.

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+      end
+    end
+  end
+
+  def requires_root?
+    false
+  end
+
+  def run_command(*args)
+    #debugger
```
cornelius: This should not be here.

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+
+    # There are three valid ways how to call Cheetah.run, whose interface this
+    # method mimics. The following code ensures that the "commands" variable
+    # consistently (in all three cases) contains an array of arrays specifying
+    # commands and their arguments.
+    #
+    # See comment in Cheetah.build_commands for more detailed explanation:
+    #
+    #   https://github.com/openSUSE/cheetah/blob/0cd3f88c1210305e87dfc4852bb83040e82d783f/lib/cheetah.rb#L395
+    #
```
cornelius: What's the point of having this comment here?

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+      cheetah_class = LoggedCheetah
+    end
+    with_utf8_locale do
+      cheetah_class.run(*["docker", "exec", "-i", host, "bash", "-c", piped_args.compact.flatten.join(" "), options])
+
+    end
+  end
+
+  # Retrieves files specified in filelist from the local system and raises an
+  # Machinery::Errors::RsyncFailed exception when it's not successful. Destination is
```
cornelius: The part with the exception is not true, is it?

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+  def read_file(file)
+    run_command("cat", file, stdout: :capture)
+  rescue Errno::ENOENT
+    # File not found, return nil
+    return
+  end
+
+  # Copies a file to the local system
+  def inject_file(source, destination)
+    FileUtils.copy(source, destination)
```
cornelius: Does this work with a container?

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+  rescue => e
+    raise Machinery::Errors::InjectFileFailed.new(
+      "Could not inject file '#{source}' to local system.\n" \
```
cornelius: Why "local system"?

lib/docker_system.rb
```diff
@@ -0,0 +1,148 @@
...
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+require "cheetah"
+class DockerSystem < System
+  attr_accessor :host
+
+  def initialize(host)
+    @host = host
```
cornelius: It's not really a host, it's a container.

lib/cli.rb
```diff
@@ -542,6 +548,9 @@ def self.supports_filtering(command)
         Machinery::Ui.puts filter.to_array.join("\n") + "\n\n"
       else
         show_filter_note(scope_list, filter)
+        if options["docker-container"] == true
```
cornelius: The `== true` part is redundant.

lib/cli.rb
```diff
@@ -533,6 +536,9 @@ def self.supports_filtering(command)
       if options["extract-files"] || options["extract-unmanaged-files"]
         inspect_options[:extract_unmanaged_files] = true
       end
+      if options["docker-container"]
+        inspect_options[:docker_container] = true
```
cornelius: Why is the option treated as a boolean option here, when it is defined as a string option before?

man/machinery-inspect.1.md
```diff
@@ -34,6 +34,9 @@ trigger errors.
   * `-n NAME`, `--name=NAME` (optional):
     Store the system description under the specified name.
 
+  * `-d CONTAINER_ID`, `--docker-container CONTAINER_ID` (optional):
```
cornelius: When an id is passed here, how is the `HOSTNAME` argument interpreted?

PR #1265, opened 2015-08-31
Add check if services which are intended to be linked are detected

keberle94:
- #link_compose_services checks if the services, which are intended to
  be linked are detected and throws a proper error-message which linking
  failed.


thardeck:
s/which/when/

keberle94:
Superseded by #1266

PR #1266, opened 2015-08-31
Add check if services which are intended to be linked are detected

keberle94:
Supersedes #1265 

- #link_compose_services checks if the services, which are intended to
  be linked are detected and throws a proper error-message when linking
  failed.

### DIFF
--- a/lib/workload_mapper.rb
+++ b/lib/workload_mapper.rb
@@ -32,8 +32,14 @@ class WorkloadMapper
     services.each do |service, config|
       config.fetch("links", {}).each do |linked_service|
         services[service]["environment"] ||= {}
-        vars = services[linked_service].fetch("environment", {})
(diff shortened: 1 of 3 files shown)


lib/workload_mapper.rb
```diff
@@ -32,8 +32,14 @@ def link_compose_services(services)
     services.each do |service, config|
       config.fetch("links", {}).each do |linked_service|
         services[service]["environment"] ||= {}
-        vars = services[linked_service].fetch("environment", {})
-        services[service]["environment"].merge!(vars)
+        if services[linked_service]
+          vars = services[linked_service].fetch("environment", {})
```
thardeck: Shouldn't this be "[]" instead of "{}"?

thardeck:
lgtm

PR #1275, opened 2015-09-02
Update data consistency document

cornelius:


docs/Internal-Consistency-of-System-Descriptions.md
```diff
@@ -173,20 +173,13 @@ filters:
...
 ```
 
 ### Scopes as plugins
 
 When it's clear how the data of scopes is used and it's not a problem if additional possibly duplicated data is present, it makes sense to make it as easy as possible to add scopes just to get more information about a system on inspection. This can be solved by providing new scopes as plugins.
 
-With the scope being the central unit, the files the scope consists of should be structured in a more central way. The easiest solution would be to have a directory per scope in the `scopes` directory, which at least contains a scope definition file as described above. Additional files such as a model, an inspector, etc can be added and either be automatically recognized based on naming convention or explicitly referenced by the `files` section proposed for the scope definition file.
+With the scope being the central unit, the files the scope consists of are structured in a central way. There is a directory per scope in the `scopes` directory, which at least contains a scope definition file as described above. Additional files such as a model, an inspector, etc can be added and are automatically recognized based on naming convention.
```
mauromorales: I find the first sentence to be a bit hard to read:

`With the scope being the central unit, the files the scope consists of are structured in a central way`

Maybe something shorter like the following sentence could be a bit clearer

`The scope is the central unit and its files are structured in such a way`

mauromorales:
lgtm

PR #1277, opened 2015-09-02

mschnitzer:
Squashes #1276 

- ssh brute-force or

### DIFF
--- a/lib/remote_system.rb
+++ b/lib/remote_system.rb
@@ -89,20 +89,22 @@ class RemoteSystem < System
     end
   end
 
-  # Tries to run the noop-command(:) on the remote system as root (without a password or passphrase)
-  # and raises an Machinery::Errors::SshConnectionFailed exception when it's not successful.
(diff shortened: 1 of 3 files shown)


tgoettlicher:
Have you verified that it works with servers running old versions of sshd?

mschnitzer:
@tgoettlicher That's a good point, thanks. I'll double check that tomorrow.

lib/remote_system.rb
```diff
@@ -19,6 +19,9 @@ class RemoteSystem < System
   attr_accessor :host
   attr_accessor :remote_user
 
+  SSH_BASIC_OPTIONS = ["-o", "ControlMaster=auto", "-o",
+    "ControlPath=~/.ssh/socket%r@%h-%p", "-o", "ControlPersist=600"]
```
tgoettlicher: `ssh` will fail if the `~/.ssh` directory doesn't exists.

mschnitzer:
@cornelius we should have a discussion in a meeting about that.

PR #1279, opened 2015-09-03
Refactor #assert_scopes to be more accurate in error messages

keberle94:

PR #1280, opened 2015-09-03
Refactor #assert_scopes to be more accurate in error messages

keberle94:
Squashes #1279
Fixes spelling of commitmessages


thardeck:
lgtm

lib/system_description.rb
```diff
@@ -218,11 +218,22 @@ def scopes
...
-    unless missing.empty?
+    if !missing_scopes.empty?
       raise Machinery::Errors::SystemDescriptionError.new(
-        "The system description misses the following section(s): #{missing.join(", ")}."
+        "The system description misses the following" \
+          " #{Machinery.pluralize(missing_scopes.size, "scope")}: #{missing_scopes.join(", ")}."
+      )
+    end
+
+    empty_scopes = scopes.select { |scope| !self[scope] || self[scope].empty? }
```
aduffeck: The `!self[scope]` case is already covered in line 221 so it doesn't have any effect here anymore. I'd remove it from the clause.

keberle94:
Closing for now, due to the discussion whether the check for empty scopes is sufficient at this place.

PR #1283, opened 2015-09-04
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150904_enable_helper_inspection_with_file_extraction'

aduffeck:
  * a1b83a2 Adapt changelog entry for helper inspection
  * 767af34 Remove dead code
  * 9b1ac1a Do not exclude files with broken UTF-8 filenames anymore
  * ed71365 Enable helper to inspect unmanaged-files with extraction

### DIFF
--- a/plugins/unmanaged_files/unmanaged_files_inspector.rb
+++ b/plugins/unmanaged_files/unmanaged_files_inspector.rb
@@ -271,7 +271,6 @@ class UnmanagedFilesInspector < Inspector
       helper.remove_helper
     end
 
-    scope.extracted = !!do_extract
     scope.files.delete_if { |f| filter.matches?(f.name) }
 
     if do_extract
@@ -291,8 +290,10 @@ class UnmanagedFilesInspector < Inspector
       file_store_final.remove
       file_store_tmp.rename(file_store_final.store_name)
       scope.scope_file_store = file_store_final
+      scope.extracted = true
     else
       file_store_final.remove
+      scope.extracted = false
     end
 
     @description["unmanaged_files"] = scope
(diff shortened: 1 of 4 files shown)


keberle94:
LGTM

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -246,10 +244,31 @@ def inspect(filter, options = {})
       ensure
         helper.remove_helper
       end
-      scope.extracted = false
 
+      scope.extracted = !!do_extract
```
cornelius: This is a confusing statement. Maybe it would be clearer to put a `scope.extracted = true` at the end of the extraction code and a `scope.extracted = false` at the end of the code, which doesn't extract.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -246,10 +244,31 @@ def inspect(filter, options = {})
...
+
+        scope.retrieve_trees_from_system_as_archive(@system,
+          scope.files.select(&:directory?).map(&:name), []) do |count|
+          show_extraction_progress(files.count + count)
+        end
+
+        scope.files = extract_tar_metadata(scope.files, file_store_tmp.path)
+        file_store_final.remove
+        file_store_tmp.rename(file_store_final.store_name)
+        scope.scope_file_store = file_store_final
```
cornelius: It would be good to have this code in an own method.

plugins/unmanaged_files/unmanaged_files_inspector.rb
aduffeck: Is it? The double-bang is relatively common operator imo, I've seen it in several places before.

plugins/unmanaged_files/unmanaged_files_inspector.rb
aduffeck: Agreed, I'll extract it.

cornelius:
I think this code needs some more refactoring, but it is good enough for now.

PR #1285, opened 2015-09-04
Disable containerize command

mauromorales:


thardeck:
lgtm but we probably also need to update the Wiki

mauromorales:
removed from wiki

PR #1288, opened 2015-09-04
Containerize portus

mauromorales:


lib/containerize_task.rb
```diff
@@ -43,4 +44,16 @@ def write_readme_file(dir)
       File.join(dir, "README.md")
     )
   end
+
+  private
+
+  def copy_workload_setup_files(description, workloads, services, path)
```
mauromorales: They are being used by the erb template

lib/containerize_task.rb
```diff
@@ -43,4 +44,16 @@ def write_readme_file(dir)
...
+    workloads.each do |workload, workload_config|
```
mauromorales: It's being used by the erb template

export_helpers/containerize_readme.md
```diff
@@ -20,6 +20,13 @@ Install on openSUSE 13.2:
 
 Log out and in again to refresh the user's group.
 
+## Set Up
+
+We've included a rake task to setup your new containerized application. Please
+run it before starting docker-compose.
+
+    rake setup
```
cornelius: Why is this a rake task? Wouldn't a plain script be more simple?

lib/containerize_task.rb
cornelius: Wouldn't it be nicer to make this somehow explicit? E.g. by encapsulating what is available to the template in a separate class?

export_helpers/containerize_readme.md
mauromorales: @cornelius I liked the idea of having more control through the API so it was possible to check when a container was running or not. Here is a version with the shell script, definitely more simple but I still don't like to have to wait for the db on a loop, specially when it cannot connect for some issue.

https://github.com/SUSE/machinery/pull/1294

PR #1291, opened 2015-09-07
Move scopes_empty? check from assert scopes and make it a empty_repo check in KiwiConfig

keberle94:
Fixes Merge conflict
Supersedes #1287 
Fixes #1268

PR #1292, opened 2015-09-07
Move scopes_empty? check from assert scopes and make it a empty_repo check in KiwiConfig

### LINKED ISSUE #1268: export-autoyast option fails when target has no repositories.
image

keberle94:
Squashes #1291
Fixes #1268


keberle94:
See lgtm #1291

PR #1296, opened 2015-09-07

mschnitzer:
Supersedes #1277 

- ssh brute-force or


mschnitzer:
Needs rebasing.

PR #1297, opened 2015-09-07

mschnitzer:
Supersedes #1296 

- ssh brute-force or


mschnitzer:
WIP: Please do not merge yet!

keberle94:
Because of an Bug in SSH, which is blocking Cheetah we drop this PR for now.

PR #1298, opened 2015-09-07
Containerize portus 2

mauromorales:
supersedes #1288


workload_mapper/mariadb/container/scripts/config_mariadb.sh
```diff
@@ -46,4 +46,6 @@ if [ -z "$DB_FILES" ]; then
 fi
 
 # Don't run this again.
-rm -f /scripts/config_mariadb.sh
+# rm -f /scripts/config_mariadb.sh
```
cornelius: Is this commented out intentionally? If yes, I think it should rather be removed.

workload_mapper/rails/setup/setup.rb.erb
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env ruby
+# Copyright (c) 2013-2015 SUSE LLC
```
cornelius: I suppose this should rather be just 2015, as it's a new piece of code.

export_helpers/containerize_readme.md
```diff
@@ -20,6 +20,13 @@ Install on openSUSE 13.2:
 
 Log out and in again to refresh the user's group.
 
+## Set Up
+
+When neccesarry, we've included a `setup.rb` script in your new containerized
+application. Please run it before starting docker-compose.
+
+    ./setup.rb
```
cornelius: We could go one step further and wrap running setup and docker-compose in one `start` script, so that it's really just one step for the user to get the containers up and running. But I'm ok with the solution for now. We can refine it later, if needed.

cornelius:
Besides the comments about minor issues the change looks good to me.

PR #1299, opened 2015-09-07
Disabling bundling for openSUSE Tumbleweed

thardeck:
We wanted to submit machinery to openSUSE Factory where bundling of gems
isn't allowed.

### DIFF
--- a/RPM_CHANGES
+++ b/RPM_CHANGES
@@ -1,5 +1,6 @@
 # Machinery RPM Changelog
 
+* disable gem bundling for openSUSE Tumbleweed
 
 ## Version 1.12.0 - Fri Sep 04 16:51:10 CEST 2015 - cschum@suse.de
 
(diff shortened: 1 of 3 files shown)


bmwiedemann:
looks good

suse-tests-pass:
Prophet reports failure.
( Failure: Merged 2d2527d02b4b2bd4683b29de8495734a49e44eb0 into a218099f889fe953c02100d8a6e7bef64a21ec2e )

RPM_CHANGES
```diff
@@ -1,5 +1,6 @@
 # Machinery RPM Changelog
 
+* disable gem bundling for openSUSE Tumbleweed and Leap
```
cornelius: We are only disabling it for Tumbleweed, don't we?

cornelius:
The rest of the patch looks good to me.

RPM_CHANGES
thardeck: True, I forgot to change the changelog entry.

cornelius:
lgtm

PR #1302, opened 2015-09-07
Don't print path to binary if not needed

cornelius:


suse-tests-pass:
Prophet reports failure.
( Failure: Merged 7f6aaf8e837466fd542e41629a772f0f98dfae34 into a218099f889fe953c02100d8a6e7bef64a21ec2e )

cornelius:
@thardeck In which versions is which in which packages?

thardeck:
At least in 13.2 it is in the `which` package.
```
rpm -q --whatprovides /usr/bin/which 
which-2.20-4.1.2.x86_64
```

thardeck:
I don't like the which approach. Like I mentioned if you have the machinery package installed and use another machinery version - like in the git tree - it still would show only machinery instead of the git machinery path.

Why not just match if the path is `/usr/bin/machinery` and only in this case do not show the path?

cornelius:
On Thursday 10 September 2015 00:52:19 Tim Hardeck wrote:
> I don't like the which approach. Like I mentioned if you have the machinery
> package installed and use another machinery version - like in the git tree
> - it still would show only machinery instead of the git machinery path.
> 
> Why not just match if the path is `/usr/bin/machinery` and only in this case
> do not show the path?

Well, this is what the patch is doing. Doesn't it work for you?

-- 
Cornelius Schumacher <cschum@suse.de>

thardeck:
You are right, I just meant we don't need `which` for this but if you prefer it than an static entry, fine with me.

thardeck:
lgtm

thardeck:
http://ruby-doc.org/stdlib-2.0.0/libdoc/mkmf/rdoc/MakeMakefile.html#method-i-find_executable might be a better solution than which.

cornelius:
The `find_executable` is not really meant for use in applications, it e.g. writes a log file by default. So I'm going with the `which` solution for now. If that causes problems we can still look for something else.

PR #1304, opened 2015-09-08
Improve output format of `machinery config`

keberle94:

PR #1313, opened 2015-09-09
Improve output format of `machinery config`

keberle94:
Supersedes #1304

### DIFF
--- a/spec/unit/config_task_spec.rb
+++ b/spec/unit/config_task_spec.rb
@@ -23,6 +23,7 @@ describe ConfigTask do
   include FakeFS::SpecHelpers
   let(:config_task) { ConfigTask.new }
   let(:key) { "hints" }
+  let(:long_key) { "remote-user-configuration-change-long" }
 
   describe "#config" do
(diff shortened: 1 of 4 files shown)


spec/unit/config_task_spec.rb
```diff
@@ -59,9 +60,14 @@
...
-      expect(captured_machinery_output).to include("#{key}=")
-      expect(captured_machinery_output).to include("remote-user=")
+      config = Machinery::Config.new
+      config.entry(key, default: true, description: "configtext")
+      config.entry(long_key, default: "root", description: "configtext")
+      @config_task = ConfigTask.new(config)
+
+      @config_task.config
+      expect(captured_machinery_output).to match(/#{key}\s{33}=/)
+      expect(captured_machinery_output).to match(/#{long_key}\s=/)
```
thardeck: `\s` should be a real space I suppose

thardeck:
I have tested it and there seems to be an unexpected line break

thardeck:
lgtm

PR #1319, opened 2015-09-10
Set status context to allow triggered prophet job to run actual tests.

greygoo:
When running the trigger job in the integration context, it already adds
a comment to the pull request which then indicates to the triggered actual
test run that tests ran already, making it skip the test execution. By
setting a own context called trigger we avoid this. The backdraw of this
approach however is an additional status entry for each pull request called
prophet trigger.


cornelius:
lgtm

PR #1320, opened 2015-09-10
Unbundle machinery helper

thardeck:


thardeck:
So I have talked to Cornelius about it and there is the issue that the old machinery-helper package will get removed as soon as there is a new in OBS so I try to fix this.

We would also need a runtime check for the version because of the gem machinery version.

thardeck:
We need to make a decision how we want to go forward first.

PR #1327, opened 2015-09-11
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_150911_fix_extraction_with_helper'

aduffeck:
  * 7ff0b31 Filter special and remote mounts when using helper inspection
  * 4a14bde Switch to root inspection for most of the tests


thardeck:
lgtm

PR #1331, opened 2015-09-11
Improve consistency in cli.rb

mschnitzer:
I improved the consistency in our cli.rb for each sub command. Currently we have sometimes sub command sections with a leading : in front of a command and sometimes we quoted our sub command. Quoting is necessary when we want to use a sub command with a dash in it.

Inconsistency-Examples:
```ruby
command :analyze do |c|
```
```ruby
command "show" do |c|
```
I just quoted all sub commands and removed the leading : from some sub commands.

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -254,7 +254,7 @@ class Cli
       the package for the modified config files
   LONGDESC
   arg "NAME"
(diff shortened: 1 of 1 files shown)


cornelius:
Looks good to me.

PR #1310, opened 2015-09-08
Handle socket errors in `-i` option

mschnitzer:
Supersedes #1185

PR #1335, opened 2015-09-11
Allow only ports in a range between 2-65535

mschnitzer:
Supersedes #1310


lib/cli.rb
```diff
@@ -348,6 +360,11 @@ def self.supports_filtering(command)
 
       name1 = shift_arg(args, "NAME1")
       name2 = shift_arg(args, "NAME2")
+
+      if options[:html]
+        check_port_validity(options[:port])
+      end
```
aduffeck: This could be simplified to one line like that: 

```ruby
check_port_validity(options[:port]) if options[:html]
```

lib/cli.rb
```diff
@@ -667,6 +684,11 @@ def self.supports_filtering(command)
       if name == "localhost" && !CurrentUser.new.is_root?
         Machinery::Ui.puts "You need root rights to access the system description of your locally inspected system."
       end
+
+      if options[:html]
+        check_port_validity(options[:port])
+      end
```
aduffeck: Same here.

NEWS
```diff
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Fix: Allow only ports (for command show and serve) in a range between
+  2 and 65535 (gh#SUSE/machinery#1316)
```
aduffeck: Maybe change that to 

```
* Validate port option in the `show` and `serve` commands (gh#SUSE/machinery#1316)
```

mschnitzer:
Superseded by #1337

PR #1336, opened 2015-09-11
Always use `Hint.program_name`

cornelius:


bin/machinery
```diff
@@ -22,7 +22,7 @@ require_relative '../lib/machinery'
 begin
   LocalSystem.validate_machinery_compatibility
   Machinery.initialize_logger(ENV["MACHINERY_LOG_FILE"] || Machinery::DEFAULT_LOG_FILE)
-  command_log = "Executing (Version #{Machinery::VERSION}) '#{$0} #{ARGV.join(" ")}'"
+  command_log = "Executing (Version #{Machinery::VERSION}) '#{$PROGRAM_NAME} #{ARGV.join(" ")}'"
```
mschnitzer: Did you forget to use Hint.program_name or is that not possible here?

mschnitzer:
What should `#{Hint.program_name}` actually do when a user uses an alias? For example: I have an alias called `gm` for the GitHub binary located in `~/GitHub/machinery/bin/machinery`.

When I run for example `gm compare osuse131_not_extracted osuse132_not_extracted`, the log file has the following entry:
```
I, [2015-09-14T10:16:26.887834 #29540]  INFO -- : Executing (Version 1.12.0) '/home/manuel/GitHub/machinery/bin/machinery compare osuse131_not_extracted osuse132_not_extracted'
```

Shouldn't it be `gm compare osuse131_not_extracted osuse132_not_extracted`, should it? Or is this more a shell related thing? For example: The shell replaces the alias with its content and executes that command.

Anyway, when I create a file in /usr/bin called `machinery` and place the following content in it:
```ruby
require_relative '/home/manuel/GitHub/machinery/lib/machinery'

begin
  LocalSystem.validate_machinery_compatibility
  Machinery.initialize_logger(ENV["MACHINERY_LOG_FILE"] || Machinery::DEFAULT_LOG_FILE)
  command_log = "Executing (Version #{Machinery::VERSION}) '#{$PROGRAM_NAME} #{ARGV.join(" ")}'"
  command_log += " (store: #{ENV["MACHINERY_DIR"]})" if ENV["MACHINERY_DIR"]
  Machinery.logger.info command_log
  Cli.run(ARGV)
rescue Machinery::Errors::IncompatibleHost => e
  puts e
```
When I now call `machinery` compare osuse131_not_extracted osuse132_not_extracted`

cornelius:
The log file is not the right place to look for the output of `Hint.program_name`. It's not used there. This still uses `$PROGRAM_NAME` (which is the same as `$0`). In the log file it's more useful to have the full path, because this makes it clear and unambiguous, what binary actually was called.

In the user facing output as part of the hints or other output containing machinery call examples, it's more useful to have what the user has typed. You can see this, when you look at the output of code changes in this pull request or (probably most simple) `machinery help`.

Unfortunately it's not entirely possible to always find out how the user called the program, so some cases will not be covered.

Anyway, we now have a central method to improve this, if we can, but this is orthogonal to this pull request as this is only about actually using the central method.

mschnitzer:
@cornelius OK, got it. Could you please fix the Hound issues?

mschnitzer:
@cornelius lgtm

PR #1342, opened 2015-09-14
Do not overwrite the locale setting in LoggedCheetah#run

aduffeck:
**NOTE**: This patch is not finished, see my mail for details.


thardeck:
I'll will look into it

PR #1347, opened 2015-09-14
Require to run commands as root in container readme

mauromorales:


greygoo:
running 
  setup.rb as user
  docker-compose build as root
  docker-compose up as user 
did work here fine. But for now I'd be fine with running everything as root.

greygoo:
LGTM

PR #1349, opened 2015-09-14
[Review] Request from 'greygoo' @ 'SUSE/machinery/review_150914_update_news_file_to_include_all_changes_since_last_release_'

greygoo:
  * 4ca4883 Add changes from git log that are missing in NEWS.


NEWS
```diff
@@ -12,6 +12,10 @@
   which have empty repository scopes (gh#SUSE/machinery#1268)
 * Inspection of unmanaged-files is now also using the faster machinery-helper
   when the files are extracted
+* Introduce format version 5 which adds the "environment" scope
```
cornelius: Maybe better 'internal "environment" scope' to make it clear that it's not supposed to be visible to users.

NEWS
```diff
@@ -12,6 +12,10 @@
...
+* Only print binary path if it's not first in PATH(gh#SUSE/machinery#1284)
+* Add documentation about the concept of inspection
```
cornelius: This is not part of the release, so let's leave it out from the NEWS.

NEWS
```diff
@@ -12,6 +12,10 @@
   which have empty repository scopes (gh#SUSE/machinery#1268)
 * Inspection of unmanaged-files is now also using the faster machinery-helper
   when the files are extracted
+* Introduce format version 5 which adds the "environment" scope,
+  "environment" being a hidden scope only used internaly has no impact on users
+* Only print binary path if it's not first in PATH(gh#SUSE/machinery#1284)
```
cornelius: There probably should be a space after "PATH"

NEWS
```diff
@@ -12,6 +12,10 @@
   which have empty repository scopes (gh#SUSE/machinery#1268)
 * Inspection of unmanaged-files is now also using the faster machinery-helper
   when the files are extracted
+* Introduce format version 5 which adds the "environment" scope,
+  "environment" being a hidden scope only used internaly has no impact on users
```
cornelius: Maybe better two sentences? "Introduce format version 5 which adds the "environment" scope. This scope is a hidden scope which is used internally by Machinery."

"has no impact on users" is not completely true, because it requires the format upgrade. That's why we are mentioning it here.

NEWS
mschnitzer: @greygoo I would also suggest to use backticks instead of quotes. It seems that we're using more backticks than quotes. Am I right?

thardeck: Nearly all other format changes look like this:
`Introduce system description format version 3 (see https://github.com/SUSE/machinery/wiki/System-Description-Format#version-3)`

NEWS
```diff
@@ -12,6 +15,8 @@
   which have empty repository scopes (gh#SUSE/machinery#1268)
 * Inspection of unmanaged-files is now also using the faster machinery-helper
   when the files are extracted
+* Only print binary path if it's not first in PATH(gh#SUSE/machinery#1284)
```
cornelius: There is a space missing after "PATH"

NEWS
```diff
@@ -12,6 +12,11 @@
   which have empty repository scopes (gh#SUSE/machinery#1268)
 * Inspection of unmanaged-files is now also using the faster machinery-helper
   when the files are extracted
+* Introduce format version 5 which adds the "environment" scope (see
+  https://github.com/SUSE/machinery/blob/master/docs/System-Description-Format.md#version-5)
```
cornelius: There needs to be a period at the end of the sentence. Otherwise the two sentences are not separated.

NEWS
```diff
@@ -12,6 +12,11 @@
...
+  This scope is a hidden scope which is used internally by Machinery
```
cornelius: I would also add a period at the end here, as this entry is not using the short form where we leave out the period, but consists of full sentences.

cornelius:
lgtm

PR #1358, opened 2015-09-15
Fix UTF-8 filtering on 13.1

thardeck:
In Ruby 2.0 the encode method is a no-op when encoding a string to its
current encoding. This prevented our method from filtering invalid
UTF-8 characters on openSUSE 13.1.

That's why we have to convert to another charset first.


cornelius:
lgtm

PR #1361, opened 2015-09-16
Update man page to latest release

thardeck:


cornelius:
Looks fine. :shipit: 

But we need to also implement a mechanism to update the man page with each release.

PR #1363, opened 2015-09-17
Disable syntax checking during build

thardeck:
Checking the syntax is not needed for build. It will be run for release
though.


keberle94:
tested and even on my system build was faster, lgtm.

PR #1367, opened 2015-09-17
Containerize wordpress

mauromorales:


mauromorales:
@cornelius I've moved the main functionality to the ScopeFileAccessArchive and just loop over the scopes inside SystemDescription. Definitely feels better to have this separated :) ... let me know if I missed anything

lib/scope_file_access_archive.rb
```diff
@@ -32,6 +32,16 @@ def export_files_as_tarballs(destination)
     end
   end
 
+  def has_file?(name)
+    found = files.any? { |file| file.name == name }
+    return true if found
```
cornelius: The intermediate `found` variable is not really needed, is it?

lib/system_description.rb
```diff
@@ -276,19 +276,28 @@ def runs_service?(name)
   end
 
   def has_file?(name)
-    self["config_files"].files.any? { |file| file.name == name } ||
-      self["unmanaged_files"].files.any? { |file| file.name == name }
+    EXTRACTABLE_SCOPES.each do |scope|
+      found = self[scope].has_file?(name)
+      return true if found
```
cornelius: Same here. `return true if self[scope].has_file?(name)` would do the trick as well.

lib/system_description.rb
```diff
@@ -276,19 +276,28 @@ def runs_service?(name)
...
+      end
     end
+    # if scope_extracted?("config_files")
+    #   file = self["config_files"].files.find { |f| f.name == path }
+    #   return parse_variable_assignment(file.content, key) if file
+    # end
+    # if scope_extracted?("unmanaged_files")
+    #   file = self["unmanaged_files"].files.find { |f| f.name == path }
+    #   return parse_variable_assignment(file.content, key) if file
+    # end
```
cornelius: The commented code should be removed.

spec/unit/system_description_spec.rb
```diff
@@ -555,26 +555,6 @@ class SystemDescriptionFooConfig < Machinery::Object; end
...
-              }
-            ]
-          }
-        }
-      EOF
-    }
-    it "returns true when found" do
-      expect(system_description.has_file?("/etc/my.cnf")).to be_truthy
-    end
-  end
```
cornelius: Why do you remove the test? Wouldn't it still be useful?

mauromorales: @cornelius wouldn't the test in scope_file_access_archive_spec.rb be enough in this case?

cornelius: The current test actually only tests what is in the scope_file_access_archive, that's right. But it still would be good to have a test for the code in `SystemDescription.has_file?`, wouldn't it?

cornelius:
There is a warning when running docker-compose:

```
Warning: the mapping "mariadb/data:/var/lib/mysql" in the volumes config for service "db" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./mariadb/data:/var/lib/mysql"
```

This should be fixed as we want to demonstrate a clean example setup.

cornelius:
When accessing the containerized application I get an error message from the web server: "Error establishing a database connection"

mauromorales:
@cornelius added a `setup.rb` script that should get rid of the issue that causes the "Error establishing a database connection" ... the warning was taken care in #1395 which is merged so once I rebase will be included

workload_mapper/wordpress/setup/setup.rb.erb
```diff
@@ -0,0 +1,67 @@
...
+
+db_image = "#{app_name}_db"
+build_image("./#{db_workload}", db_image)
+
+volumes = ["#{Dir.pwd}/#{db_workload}/data:/var/lib/mysql"]
+db = Docker::Container.create("Image" => db_image, "name" => "#{db_image}_0", "Env" => db_env_vars, "Binds" => volumes, "Tty" => true, "Stream" => true)
+db.tap(&:start).attach(tty: true) { |output| print output }
+
+web_image = "#{app_name}_web"
+build_image("./rails", web_image)
```
cornelius: This should be "./wordpress", shouldn't it?

cornelius:
With the change in the setup script it looks good to me.

PR #1368, opened 2015-09-17
Add rake tasks for integration test groups acceptance, minimal and full.

greygoo:


Rakefile
```diff
@@ -37,6 +37,24 @@ RSpec::Core::RakeTask.new("spec:integration") do |t|
   t.pattern = "spec/integration/**/*_spec.rb"
 end
 
+desc "Run minimal RSpec code examples in spec/integration"
```
aduffeck: The description (`minimal`) doesn't match the selected tests.

Rakefile
```diff
@@ -37,6 +37,24 @@ RSpec::Core::RakeTask.new("spec:integration") do |t|
...
+  RSpec::Core::RakeTask.new("spec:integration:acceptance") do |t|
```
aduffeck: The indentation is wrong here, it should be aligned with the beginning of the line.

keberle94:
At first I was a little bit confused by the fact that minimal runs more tests than the acceptance tests. But after a look at the spreadsheet it behaves exactly the expected way. LGTM

PR #1369, opened 2015-09-17
Remove obsolete helper Note

### LINKED ISSUE #1356: Change note about machinery helper
Display the note about the machinery helper only when the old helper is running

thardeck:
the helper should be the default so no need for an extra note.

Fixes #1356


mauromorales:
lgtm

PR #1371, opened 2015-09-18
Adapt man page to Red Hat support

thardeck:
Machinery does also allow to use `yum` instead of `zypper` so `zypper`
is no exclusive requirement.


mschnitzer:
lgtm

PR #1377, opened 2015-09-18
Mention man page and wiki page in `help` command

### LINKED ISSUE #1373: Add pointers to documentation to `machinery help`
The help output of `machinery help` should mention the man page as central place of documentation  and have a link to the Wiki page for supplementary user documentation.

mschnitzer:
This fixes #1373

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -43,6 +43,9 @@ class Cli
   post do |global_options,command,options,args|
     if command.is_a?(GLI::Commands::Help) && !global_options[:version] && ARGV == ["help"]
 
+      Machinery::Ui.puts "\nFor more detailed information, open the man page or visit our " \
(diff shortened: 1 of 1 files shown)


lib/cli.rb
```diff
@@ -43,6 +43,9 @@ class Cli
   post do |global_options,command,options,args|
     if command.is_a?(GLI::Commands::Help) && !global_options[:version] && ARGV == ["help"]
 
+      Machinery::Ui.puts "\nFor more detailed information, open the man page or visit our " \
+        "wiki page: https://github.com/SUSE/machinery/wiki"
```
cornelius: It should say how the man page is opened and it should say that the primary documentation is there and that secondary information is on the Wiki (while "primary" and "secondary" are not the right terms to describe it to users).

mschnitzer:
Superseded by #1385

PR #1380, opened 2015-09-18
Add all developer guides to docs/ directory

mschnitzer:
Related to #1372

Issue #1372, opened 2015-09-18
Separate user and developer documentation


cornelius:
There is some mix between documentation targeted at users and developers of Machinery itself. It would be clearer, if they would be clearly separated.

I would suggest to do the following to fix that:

* [x] Put all documentation targeted at developers of Machinery itself into the [docs directory](https://github.com/SUSE/machinery/tree/master/docs) in the machinery git repository. So developers have everything available in one place when they check out the git repo.
* [x] Add a table of content to the [README.md](https://github.com/SUSE/machinery/blob/master/docs/README.md) in the docs directory giving an overview and some structure to the documentation.
* [x] Remove information only targeted at people who already are developers of Machinery from the Machinery Wiki. Keep the information about how to make the transition from user to developer. Also keep the information for developers working with Machinery, but not on Machinery.

greygoo:
QA PASSED

PR #1382, opened 2015-09-18
Add all developer guides to docs/ directory

mschnitzer:
Related to #1372
Supersedes #1380


cornelius:
This looks good.

There is one more page, which should be moved from the Wiki to docs/, the one about gem licenses.

mschnitzer:
Superseded by #1387

PR #1378, opened 2015-09-18
Update README and add documentation hints to it

mschnitzer:
This fixes #1374

PR #1383, opened 2015-09-18
Update README and add documentation hints to it

### LINKED ISSUE #1374: Clearly link documentation from README.md
The [Machinery README](https://github.com/SUSE/machinery/blob/master/README.md) should have clear links to documentation. It should link the end user documentation (the rendered man page) (and not link to the sources of the man page like it is doing now), and it should link the developer documentation (the docs directory).

I would suggest to adapt the structure of the README a little bit, so that it has the following sections:

* Installation
* Usage
* Documentation
* Development
* Contact

mschnitzer:
This fixes #1374 
Supersedes #1378


README.md
```diff
@@ -91,3 +97,9 @@ You can subscribe to our
 discuss using or contributing to Machinery. If you have any questions or
 feedback please feel free to
 [send them to the mailing list](mailto:machinery@lists.suse.com) as well.
+
+## Contribute
+
+Now that you have Machinery running from git on your machine you are ready to hack. If you would like to get some overview of architecture and design of Machinery have a look at our [Developer Documentation](https://github.com/SUSE/machinery/tree/master/docs).
+
+We are happy if you share your changes with us as pull requests. Read the [Contribution Guidelines](https://github.com/SUSE/machinery/blob/master/CONTRIBUTING.md#machinery-contribution-guidelines) for details how to do that.
```
cornelius: I would actually really put this section under point 6 in the Developement section, because it is the whole point of getting it running from git.

"Contribute" is also a too limiting title here, because it is not only about contributing, but also about just taking the code and hacking on it for yourself without sharing it with us.

mschnitzer:
Superseded by #1386

PR #1390, opened 2015-09-21
Build docker containers faster

mauromorales:
It's not necessary to install docker with the docker-registry so build gets faster


cornelius:
lgtm

PR #1389, opened 2015-09-21
Add User and Developer Documentation link

mschnitzer:

PR #1392, opened 2015-09-21
Add User and Developer Documentation link

mschnitzer:
Supersedes #1389


cornelius:
I don't think we should add the developer documentation here. This is accessible from the GitHub repo, which is the primary place for developers.

This page will be changed anyway, when we integrate the new information.

But let's merge the link to the user documentation for now.

mschnitzer:
Superseded by #1399

PR #1393, opened 2015-09-21
Rails dockerfile include cmd

mauromorales:


workload_mapper/rails/container/Dockerfile
```diff
@@ -43,6 +41,4 @@ ADD apache2/listen.conf /etc/apache2/listen.conf
 ADD apache2/rails_app_vhost.conf /etc/apache2/vhosts.d/
 RUN cat /etc/apache2/httpd.conf.local >> /etc/apache2/httpd.conf
 
-ENV COMPOSE=1
-
-EXPOSE 3000
```
cornelius: Why did you remove the EXPOSE statement?

cornelius: As dicsussed in person right now: The EXPOSE statement should be there, because it describes an attribute defined by the container, which is important to know when working with the container.

cornelius:
lgtm

PR #1395, opened 2015-09-22
Remove ambiguous path in compose templates

### LINKED ISSUE #1394: Warnings when running docker-compose
When running `docker-compose up` in an exported portus setup I get the following warnings:

```
Warning: the mapping "mariadb/data:/var/lib/mysql" in the volumes config for service "db" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./mariadb/data:/var/lib/mysql"
Warning: the mapping "registry:/registry" in the volumes config for service "registry" is ambiguous. In a future version of Docker, it will designate a "named" volume (see https://github.com/docker/docker/pull/14242). To prevent unexpected behaviour, change it to "./registry:/registry"
```

This is supposed to be a reference setup, so there should be no warnings.

This is with docker 1.8.2

mauromorales:
Fixes #1394

### DIFF
--- a/spec/unit/workload_mapper_spec.rb
+++ b/spec/unit/workload_mapper_spec.rb
@@ -110,7 +110,7 @@ describe WorkloadMapper do
       {
         "db" => {
           "build" => "./mariadb",
-          "volumes" => ["mariadb/data:/var/lib/mysql"],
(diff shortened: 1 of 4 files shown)


cornelius:
:shipit:

PR #1401, opened 2015-09-23
[Review] Request from 'aduffeck' @ 'SUSE/machinery/inspect_container2'

aduffeck:
  * 43f75fe Add changelog entry for docker container inspection
  * becb2b2 Only remove the helper binary after doing the file extraction
  * 833cd12 Implement "machinery inspect-container" command
  * 022ad06 Add DockerSystem and DockerContainer for working with docker containers
  * 5e43b33 Store the system type in the environment scope
  * 182875b Pass a System object to InspectTask instead of a hostname


lib/cli.rb
```diff
@@ -587,7 +591,83 @@ def self.check_port_validity(port)
...
+      end
+
+
+      Hint.print(:show_data, name: name)
+
+      if !options["extract-files"] || Inspector.all_scopes.count != scope_list.count
+        Hint.print(:do_complete_inspection, name: name, docker_container: image)
+      end
+    end
+  end
```
cornelius: There is a lot of duplication between the code handling the "inspect" command and the code doing "inspect-container". We should dry that up.

lib/machinery.rb
```diff
@@ -37,6 +37,9 @@
 require "diffy"
 require "sinatra/base"
 require "mimemagic"
+require "docker"
+require "rubygems"
+require "rubygems/package"
```
cornelius: Why are the "rubygems" requires needed?

lib/docker_container.rb
```diff
@@ -0,0 +1,119 @@
...
+    LoggedCheetah.run("docker", "exec", "-i", @container.id, *args)
+  end
+
+  # Copies a file to the container
+  def inject_file(source, destination)
+    LoggedCheetah.run("docker", "cp", source, "#{@container.id}:#{destination}")
+  end
+
+  # Retrieves files specified in filelist from the remote system and raises an
+  # Machinery::Errors::RsyncFailed exception when it's not successful. Destination is
```
cornelius: The part about the exception is not true.

lib/docker_container.rb
```diff
@@ -0,0 +1,119 @@
...
+  # the directory where to put the files.
+  def retrieve_files(filelist, destination)
+    filelist.each do |file|
+      destination_dir = File.dirname(File.join(destination, file))
+
+      tar = StringIO.new
+      @container.copy(file) { |chunk| tar.write(chunk) }
```
cornelius: Do I understand it correctly that `@container.copy` returns a tar archive? If yes, how is this different from what we do in the `create_archive` method, and couldn't it be the same?

lib/docker_system.rb
```diff
@@ -0,0 +1,92 @@
...
+
+  # Retrieves files specified in file_list from the container and creates an archive.
+  def create_archive(file_list, archive, exclude = [])
+    @container.create_archive(file_list, archive, exclude)
+  end
+
+  def requires_root?
+    false
+  end
+end
```
cornelius: `DockerSystem` does not much more than forwarding calls to `DockerContainer`. Is it really useful to have two classes here? Couldn't this just be done in `DockerSystem`?

lib/docker_system.rb
```diff
@@ -0,0 +1,92 @@
...
+    @container.inject_file(source, destination)
+  end
+
+  # Retrieves files specified in file_list from the container
+  def retrieve_files(file_list, destination)
+    @container.retrieve_files(file_list, destination)
+  end
+
+  # Retrieves files specified in file_list from the container and creates an archive.
+  def create_archive(file_list, archive, exclude = [])
```
cornelius: It's not necessarily related to this patch, but the name of this method is a bit obscure. Maybe `retrieve_files_as_archive` would be better.

lib/docker_system.rb
aduffeck: I agree, but from the discussion we had I thought it was your idea to abstract away the actual docker access (which partly happens through the API and partly throught the docker command) from the DockerSystem, but maybe I misunderstood.

I'm perfectly fine with merging those two classes together.

lib/docker_container.rb
aduffeck: There are a few subtle differences, e.g. the docker tar file doesn't contain the owner/group names of the files and the copy method doesn't allow for excluding files. It's also not possible to combine several files into one tarball. 

In order to be compatible with the `System` API we have to roll our own solution I think.

lib/machinery.rb
aduffeck: In order to gain access to the `Gem::Package::TarReader` class which we use for unpacking the tarballs.

PR #1402, opened 2015-09-23
Update release script for new `machinery-helper` version

mschnitzer:
Since the `machinery-helper` script stores the version in a version.go
file, it's necessary to adjust the release script.

### DIFF
--- a/tools/release.rb
+++ b/tools/release.rb
@@ -24,8 +24,8 @@ require_relative "spec_template"
 
 class Release
   def self.version
-    if File.exist?("VERSION")
-      File.read("VERSION").chomp
+    if File.exist?("version.go")
+      File.read("version.go")[/(\d+\.\d+\.\d+)/]
     else
       Machinery::VERSION
     end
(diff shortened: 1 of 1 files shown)


tools/release.rb
```diff
@@ -24,8 +24,8 @@
 
 class Release
   def self.version
-    if File.exist?("VERSION")
-      File.read("VERSION").chomp
+    if File.exist?("version.go")
+      File.read("version.go").chomp.match /(\d+)\.(\d+)\.(\d+)/
```
thardeck: Why is there still a chomp if a regex is missing?

tools/release.rb
```diff
@@ -24,8 +24,8 @@
 
 class Release
   def self.version
-    if File.exist?("VERSION")
-      File.read("VERSION").chomp
+    if File.exist?("version.go")
+      File.read("version.go").match /(\d+)\.(\d+)\.(\d+)/
```
thardeck: you don't return a string but an array

thardeck:
lgtm

PR #1404, opened 2015-09-23
Add move command for machinery

### LINKED ISSUE #1397: Add a command to rename descriptions
We have `machinery copy` and we have `machinery remove`. But we don't have a command to rename a description.

Following the scheme of naming commands after the UNIX tools, this command should probably called `machinery move`.

keberle94:
Fixes  #1397
The `machinery move` command moves a systemdescription to the specified
name.


NEWS
```diff
@@ -1,5 +1,6 @@
 # Machinery Release Notes
 
+* Add `move` command to be able to rename system descriptions (gh#SUSE/machinery#1397)
```
thardeck: I would prefer 
```
Add `move` command to rename system descriptions (gh#SUSE/machinery#1397)
```

man/machinery-move.1.md
```diff
@@ -0,0 +1,30 @@
...
+
+`machinery move`
+    FROM_NAME TO_NAME
+
+`machinery` help move
+
+
+### DESCRIPTION
+
+The `move` command moves a stored system description. It moves a new
```
thardeck: I would call it `renames` and the second sentence is confusing..

man/machinery-move.1.md
```diff
@@ -0,0 +1,30 @@
...
+description named TO_NAME as the description FROM_NAME.
+
+
+### ARGUMENTS
+  * `FROM_NAME` (required):
+    Name of the source system description.
```
thardeck: I would prefer something like `Current name of the system description`

man/machinery-move.1.md
```diff
@@ -0,0 +1,30 @@
...
+
+  * `TO_NAME` (required):
+    Name of the target system description.
```
thardeck: I would prefer something like `New name of the system description`

spec/unit/move_task_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+describe MoveTask do
+  let(:move_task) { MoveTask.new }
+
+  describe "#move" do
+    it "moves the system description directory" do
+      store = double
+      from = "description1"
+      to = "description2"
+      expect(store).to receive(:move).with(from, to)
+      move_task.move(store, from, to)
```
thardeck: Do we have tests for the methods which check for the [non]existance of system descriptions?

Anyway I would at least add one test for a conflict I think.

spec/unit/system_description_store_spec.rb
```diff
@@ -176,6 +176,35 @@
...
+      create_machinery_dir
+    end
+
+    it "moves an existing SystemDescription" do
+      expect(store.list).to eq([test_name])
+      store.move(test_name, new_name)
+      expect(store.list).to eq([new_name])
+    end
+
+    it "throws an error when the to be copied SystemDescription does not exist" do
```
thardeck: the title

thardeck:
I would rename the command `move` to `rename`

cornelius:
Why would "rename" be better than "move". The reasoning behind "move" is that it would be the equivalent of `mv` just as "list" is the equivalent of `ls` and "copy" is the equivalent of `cp`.

thardeck:
Because move sounds like we would move a system description to another place, like a different system description store and not just change the name.

thardeck:
And we also don't use the standard unix names like `cp` and `mv`

keberle94:
I have no strong

PR #1173, opened 2015-07-28
Raise error if port is already in use

keberle94:
Squashes #1169
Fixes #1153

PR #1410, opened 2015-09-23
Check for command xdg open instead of package xdg utils

tgoettlicher:


tgoettlicher:
Fixes #1173

lib/local_system.rb
```diff
@@ -67,6 +67,16 @@ def validate_existence_of_packages(packages)
       end
     end
 
+    def validate_existence_of_command(command, package)
+      Cheetah.run("which", command)
```
thardeck: The problem is that `which` is not always installed, at least it could be since it is part of the "which" package in openSUSE.

lib/local_system.rb
```diff
@@ -67,6 +67,16 @@ def validate_existence_of_packages(packages)
...
+    rescue Cheetah::ExecutionFailed
+      output = <<-EOF
+You need the command '#{command}' from package '#{package}'.
+You can install it by running `zypper install #{package}`.
```
thardeck: We shouldn't recommend using zypper if this check is mainly targeted for other distributions.

spec/unit/local_system_spec.rb
```diff
@@ -186,6 +186,26 @@
     end
   end
 
+  describe ".validate_existence_of_command" do
+    it "raises an Machinery::Errors::MissingRequirementsError error if the command isn't found" do
```
thardeck: should be `Machinery::Errors::MissingRequirement`

thardeck:
What about the `lib/compare_task.rb` which also checks for the xdg-utils.

lib/local_system.rb
tgoettlicher: `which` is required by the RPM's spec file.

lib/local_system.rb
tgoettlicher: Users of other distributions will know how to install packages on their distributions. Let's leave it as it is for now.

thardeck:
lgtm

PR #1408, opened 2015-09-23
Add move command for machinery

keberle94:
Fixes #1397
The machinery move command moves a systemdescription to the specified
name.

PR #1411, opened 2015-09-23
Add move command for machinery

### LINKED ISSUE #1397: Add a command to rename descriptions

keberle94:
Fixes #1397
Squashes #1408
The machinery move command moves a system-description to the specified
name.


thardeck:
lgtm

PR #1413, opened 2015-09-25
Log extracted files

thardeck:
The plan was to log the files which are extracted and not the commands since they only provided the destination.

### DIFF
--- a/lib/system.rb
+++ b/lib/system.rb
@@ -68,6 +68,9 @@ class System
   # To be able to deal with arbitrary filenames we use zero-terminated
   # filelist and the --null option of tar
   def create_archive(file_list, archive, exclude = [])
+    Machinery.logger.info(
+      "The following files are packaged in #{archive}: " + Array(file_list).join(", ")
+    )
     created = !File.exists?(archive)
     out = File.open(archive, "w")
     begin
--- a/lib/system.rb
+++ b/lib/system.rb
@@ -76,7 +76,8 @@ class System
         *exclude.flat_map { |f| ["--exclude", f]},
         stdout: out,
         stdin: Array(file_list).join("\0"),
-        privileged: true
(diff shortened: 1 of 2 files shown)


mauromorales:
In cases where there is a list of files I think it looks good to have a line break

    I, [2015-09-25T10:07:16.913573 #15208]  INFO -- : The following files are extracted:
    /etc/.pwd.lock
    /etc/alternatives/jsondiff
    /etc/alternatives/jsonpatch
    /etc/alternatives/jsonpointer

But in cases where there is just one file, I think it looks weird

    I, [2015-09-25T10:07:31.373905 #15208]  INFO -- : The following files are extracted:
    /var/lib/cloud/seed/
    I, [2015-09-25T10:07:31.614012 #15208]  INFO -- : The following files are extracted:
    /var/lib/cloud/sem/
    I, [2015-09-25T10:07:31.842499 #15208]  INFO -- : The following files are extracted:
    /var/lib/docker/

if we remove for those cases it feels much better

    I, [2015-09-25T10:07:31.373905 #15208]  INFO -- : The following files are extracted: /var/lib/cloud/seed/
    I, [2015-09-25T10:07:31.614012 #15208]  INFO -- : The following files are extracted: /var/lib/cloud/sem/
    I, [2015-09-25T10:07:31.842499 #15208]  INFO -- : The following files are extracted: /var/lib/docker/

thardeck:
I used your first option at the beginning but we are talking about thousands of files so we would increase the log size quite a bit. And leaving files out is no option either so I though using one line should be Ok, at least for debugging purposes.

mauromorales:
@thardeck right but couldn't we put the line for when having multiple files and remove it when we have just one?

mauromorales:
lgtm

PR #1415, opened 2015-09-25
Expand list command to take optional multiple Name parameters

### LINKED ISSUE #1398: Add command to show overview of description
It's hard to get an overview of a description, e.g. what scopes it contains, if the scopes are extracted, when it was last modified, from which host it was inspected, etc.

`machinery list` shows at least a good part of this information, but only for all descriptions at once. Adding `machinery list NAME` to show only the overview for the given description would solve the problem.

keberle94:
Fixes #1398


keberle94:
Closing until Integration test is fixed.

man/machinery-list.1.md
```diff
@@ -4,13 +4,15 @@
 ### SYNOPSIS
 
 `machinery list`
+    NAME
```
tgoettlicher: This doesn't say that you can pass multiple arguments.

NEWS
```diff
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Expand `list` command be able to take system descriptions which should be shown
```
thardeck: I would prefer something like
```
Allow limiting the `list` command output to certain system descriptions by passing them along as argument
```

man/machinery-list.1.md
```diff
@@ -26,6 +28,10 @@ scopes for each system.
 
 ### EXAMPLES
 
+  * Lists the two specified system descriptions `a` and `b`
```
tgoettlicher: Colon (`:`) is missing.

lib/cli.rb
```diff
@@ -608,11 +608,15 @@ def self.check_port_validity(port)
 
   desc "List system descriptions"
   long_desc <<-LONGDESC
-    List system descriptions and their stored scopes.
+    List system descriptions and their stored scopes, when no NAME parameter is specified.
```
tgoettlicher: `List all system descriptions...`

man/machinery-list.1.md
keberle94: I just looked that up in the description for our remove command. What would be the right approach here [NAME , NAME] ?

keberle94:
Squashed by #1418

PR #1417, opened 2015-09-25
Don't write to local directory

cornelius:
The test was writing to the local directory. Write to a temporary
directory instead and clean up when the test is done.


mauromorales:
lgtm

PR #1418, opened 2015-09-25
Expand list command to take optional multiple Name parameters

### LINKED ISSUE #1398: Add command to show overview of description

keberle94:
Squashes #1415 
Fixes #1398

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Allow limiting the `list` command output to certain system descriptions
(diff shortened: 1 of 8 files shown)


tgoettlicher:
LGTM

PR #1419, opened 2015-09-25
Remove `-i` and `-p` from `show` and add a `--public` option to `serve`

### LINKED ISSUE #1341: Clean up binding the server for the HTML view to IP addresses
At the moment the `show` and `serve` command have an `-i` option which takes an IP address (or a host name resolving to an IP address), the server serving the HTML view binds to, so that the view can be made available on a publicly reachable IP address.

The option is duplicated across the two commands, and the only valid values are IP addresses of the machine where the server is running. This makes it quite confusing for users.

I would suggest to remove the `-i` and the `-p` option from the `show` command. This command is by its nature intended to serve pages locally. So it can just always bind to a fixed port on localhost, and the browser will immediately started with a URL pointing to this.

I would also suggest to replace the `-i` option of the `serve` command by a `--public` option, which makes the server listen to all IP addresses of the machine, where it's running on, so that it is accessible from outside of the machine.

mschnitzer:
This closes #1341


lib/cli.rb
```diff
@@ -827,19 +837,36 @@ def self.check_port_validity(port)
...
+        raise Machinery::Errors::InvalidCommandLine.new("You have to specify a valid server " \
+          "port either in the 'http_server_port' section of the configuration file or " \
+          "via the --port option. The specified port '#{options[:port]}' is not valid. A valid " \
+          "port can be in a range between 2 and 65535.")
+      rescue Machinery::Errors::ServerNeedRootPrivileges
+        raise Machinery::Errors::InvalidCommandLine.new("You specified the port " \
+          "'#{options[:port]}' either in the 'http_server_port' section of the configuration " \
+          "file or via the --port option. To start the server, you need root privileges. If " \
+          "you don't want to start the server as root user, you must choose a port between " \
+          "1024 and 65535.")
```
cornelius: The error messages are duplicated here. There should only be one place where they are defined.

mschnitzer:
I'll refactor some things and will open a new PR then.

PR #1426, opened 2015-09-29
Fix wrong web image name in wordpress setup script

mauromorales:


greygoo:
LGTM, setup now finishes

PR #1429, opened 2015-09-30
Allow gli-gem between the versions `2.11.0` and `2.13.2`

keberle94:


cornelius:
Why between these versions?

keberle94:
We still need version `2.11.0` because it's the latest version for SLES, but within the new leap and tumbleweed repositories there is only version `2.13.2` available. That is the reason why we tested this newer version, which worked fine.

cornelius:
Shouldn't it just be `>= 2.11.0`?

keberle94:
Then we would allow all higher available versions without testing, wouldn't we ?

keberle94:
Andre suggested to have a <=2.14

cornelius:
For many gems we assume that newer version work as well. Why would that be different for GLI?

Looking at the gem dependencies I'm actually not sure, if we have a consistent convention there. Maybe we should define one.

aduffeck:
@keberle94 @cornelius I think it would be best to only allow patch updates in general in order to prevent issues. Those problems most likely wouldn't be caught by us our test suites but only by the users installing machinery as a gem.

cornelius:
@aduffeck Agreed

keberle94:
Superseded by #1464

PR #1431, opened 2015-10-01
Add one-click-installer file

tgoettlicher:


cornelius:
The file works on Tumbleweed. We need to call it, though. Can you change the buttons, so that there is only one "Install on openSUSE" button?

cornelius:
lgtm

PR #1433, opened 2015-10-01
Add a basic feature test for the HTML version of "machinery show"

aduffeck:
The test starts a machinery web server using "machinery serve" and opens
the according webpage in a headless browser. It then verifies that the
page can be loaded, the scopes are rendered and the collapes/expand
toggles work.

The tools that are being used for that are:

  * Capybara (https://github.com/jnicklas/capybara)

    Test framework for web applications that allows to describe the
    actions on the web page that the user would do.

  * Poltergeist (https://github.com/teampoltergeist/poltergeist)
    and Phantomjs (http://phantomjs.org/)

    Phantomjs is a headless webkit browser which is used to render the
    pages and Poltergeist is the according capybara driver

  * RSpec-Steps (https://github.com/LRDesign/rspec-steps)

    RSpec addon that allows to chain examples in a non-isolated fashion.
    That is particularly helpful when testing web apps because you can
    open a page once and have run examples on it.


cornelius:
The code looks great and works very smoothly. The warnings about phantomjs missing in the path are quite distracting, though, especially as this is in the tests, which should be green without any issues. Can we find a way how to get rid of the warnings?

cornelius:
lgtm

aduffeck:
merged

PR #1436, opened 2015-10-01
Incorporate machinery helper

thardeck:


tools/release.rb
```diff
@@ -243,46 +199,49 @@ def generate_changelog
...
+    Cheetah.run "git", "tag", "-a", @tag, "-m", "Tag version #{@release_version}"
     Cheetah.run "git", "push"
     Cheetah.run "git", "push", "--tags"
   end
 
+  def generate_development_version
+    # The development version RPMs have the following version number scheme:
+    # <base version>.<timestamp><os>git<short git hash>
+    timestamp = Time.now.strftime("%Y%m%dT%H%M%SZ")
+    commit_id = Cheetah.run("git", "rev-parse", "--short", "HEAD", :stdout => :capture).chomp
```
mschnitzer: I'm not sure if that is mistake or not (from hound)

bin/machinery
```diff
@@ -20,6 +20,12 @@
 require_relative '../lib/machinery'
 
 begin
+  if Dir.exist?(File.join(Machinery::ROOT, ".git"))
+    Dir.chdir(File.join(Machinery::ROOT, "machinery-helper")) do
+      Cheetah.run("rake")
```
aduffeck: During the discussion yesterday we said that it would be cleaner to extract that login from the Rakefile into a separate class that could be used from both the Rakefile and machinery itself. Can we still do it like that?

machinery-helper/.gitignore
```diff
@@ -0,0 +1,2 @@
+machinery-helper
```
aduffeck: Didn't we say we put the helper in `helper`?

lib/machinery_helper.rb
```diff
@@ -34,17 +31,25 @@ def initialize(s)
     @system = s
     @arch = @system.arch
 
-    @local_helpers_path = "/usr/share/machinery/helpers"
+    @local_helpers_path = File.join(Machinery::ROOT, "machinery-helper")
+    @remote_helpers_path = File.join(Machinery::REMOTE_HELPERS_PATH, "machinery-helper")
   end
 
   def local_helper_path
-    File.join(@local_helpers_path, @arch, "machinery-helper")
```
aduffeck: I wonder if it makes sense to keep the arch in the path, even though we only use one for now. That would make it clearer that it's architecture dependent and would also be more future-proof.

machinery-helper/Rakefile
```diff
@@ -0,0 +1,55 @@
...
+dir = File.expand_path(File.dirname(__FILE__))
+
+go_files = Dir.glob(File.join(dir, "*.go"))
+
+task :check_git_head do
+  git_head_file = File.join(dir, ".git_head")
+  current_head = `git rev-parse HEAD`.chomp if system("which git > /dev/null 2>&1")
+  old_head = File.read(git_head_file) if File.exist?(git_head_file)
+  if current_head != old_head
+    File.write(git_head_file, current_head)
```
aduffeck: I guess it would be safer to only write the file after the helper was successfully built.

machinery-helper/Rakefile
```diff
@@ -0,0 +1,55 @@
...
+    if ["x86_64"].include?(arch)
+      Dir.chdir(dir) do
+        puts("Building machinery-helper binary.")
+        system("go build")
+      end
+    else
+      puts("The hardware architecture #{arch} is not yet supported by the machinery-helper.")
+    end
+  else
+    puts("The Go compiler is not available on this system. Skipping building the machinery-helper.")
```
aduffeck: I think we should explain to the user what it means to them when the machinery-helper is not built and what they can do to fix it (can we tell if this rake task is being run as part of the gem installation?).

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -264,7 +264,14 @@ def helper_usable?(helper)
   def run_helper_inspection(helper, filter, do_extract, file_store_tmp, file_store_final, scope)
     begin
       helper.inject_helper
-      helper.run_helper(scope)
+      if !helper.is_supported?
+        raise Machinery::Errors::UnsupportedHelperVersion.new(
+          "Error: machinery-helper is not compatible with the Machinery version" \
+          " #{Machinery::VERSION}. The versions need to match."
+        )
+      else
```
aduffeck: The `else` branch is not needed. It would look cleaner like that imo:

```ruby
helper.inject_helper

if !helper.is_supported?
  raise Machinery::Errors::UnsupportedHelperVersion.new(
    "Error: machinery-helper is not compatible with the Machinery version" \
    " #{Machinery::VERSION}. The versions need to match."
  )
end

helper.run_helper(scope)
```

tools/release.rb
aduffeck: It's a valid complaint, it should be `stdout: :capture` instead of `:stdout => :capture`.

lib/machinery_helper.rb
```diff
@@ -22,10 +22,7 @@
 # The inspection checks, if a binary helper is available on the machine where
 # the inspection is started. It looks at the location
 #
-#    /usr/share/machinery/helpers/<arch>/machinery-helper
-#
-# where <arch> is the hardware architecture of the target system. Valid values
-# are x86_64, i586, s390x, and ppcle.
+#    <machinery-dir>/machinery-helper/machinery-helper
```
cornelius: What is `<machinery-dir>`? That's not clear from the comment.

lib/machinery_helper.rb
```diff
@@ -34,17 +31,25 @@ def initialize(s)
     @system = s
     @arch = @system.arch
 
-    @local_helpers_path = "/usr/share/machinery/helpers"
+    @local_helpers_path = File.join(Machinery::ROOT, "machinery-helper")
+    @remote_helpers_path = File.join(Machinery::REMOTE_HELPERS_PATH, "machinery-helper")
```
cornelius: It should be `@remote_helper_path` (singular), as there always is only one helper installed on the remote machine.

lib/machinery_helper.rb
```diff
@@ -34,17 +31,25 @@ def initialize(s)
...
   def can_help?
-    File.exist?(local_helper_path)
+    if File.exist?(local_helper_path)
+      LocalSystem.validate_architecture(@arch)
+      true
+    else
+      false
+    end
+  rescue Machinery::Errors::UnsupportedArchitecture
+    false
```
cornelius: You should not use exceptions to control program flow.

lib/machinery_helper.rb
```diff
@@ -52,15 +57,19 @@ def inject_helper
   end
 
   def run_helper(scope)
-    json = @system.run_command(
-      File.join(
-        Machinery::REMOTE_HELPERS_PATH, "machinery-helper"
-      ), stdout: :capture, stderr: STDERR
-    )
+    json = @system.run_command(@remote_helpers_path, stdout: :capture, stderr: STDERR)
```
cornelius: The code would be a little bit clearer and robust, if it would not use the member variable directly, but a method to access the remote helper path.

lib/machinery_helper.rb
```diff
@@ -52,15 +57,19 @@ def inject_helper
...
     scope.set_attributes(JSON.parse(json))
   end
 
   def remove_helper
-    @system.remove_file(File.join(Machinery::REMOTE_HELPERS_PATH, "machinery-helper"))
+    @system.remove_file(@remote_helpers_path)
+  end
+
+  def is_supported?
```
cornelius: `is_supported?` is a too generic name. The method name should make it clear what it does, especially when thinking from the point of view of the calling code.

docs/Inspection-Concept.md
```diff
@@ -63,9 +63,8 @@ self-contained binary without external dependencies. It does the gathering of
 information and most of the processing on the inspected machine and then returns
 the result to Machinery.
 
-For each platform there is a helper binary. The binary is bundled with the
-Machinery package and is installed to
-`/usr/share/machinery/helpers/<arch>/machinery-helper`. On the target system it
+There is a helper binary which is bundled with the Machinery package and is installed
+to `<machinery-dir>/machinery-helper/machinery-helper`. On the target system it
```
cornelius: Please explain what `<machinery-dir>` is.

Rakefile
```diff
@@ -67,67 +67,55 @@ RSpec::Core::RakeTask.new("spec:tools") do |t|
...
+  conf.version = Machinery::VERSION
+
+  #lets ignore license check for now
+  conf.skip_license_check << /.*/
 end
 
 namespace :man_pages do
   desc 'Build man page(s)'
   task :build do
-    if Dir.exist?("man")
```
cornelius: Why should the directory not exist?

Rakefile
```diff
@@ -67,67 +67,55 @@ RSpec::Core::RakeTask.new("spec:tools") do |t|
...
+    # Build man page for website (manual.html)
+    system "ronn -f man/generated/machinery.1.md"
+    system "man/generate_man"
   end
 end
 
+# Disable packaging_tasks' tarball task. We package a gem, so we don't have to
+# put the sources into IBS. Instead we build the gem in the tarball task
+Rake::Task[:tarball].clear
+task :tarball => ["man_pages:build"] do
```
cornelius: Why is the task called `:tarball`, if it actually builds a gem?

machinery-helper/Rakefile
```diff
@@ -0,0 +1,55 @@
...
+task :build_helper do
+  if system("which go > /dev/null 2>&1")
+    arch = `uname -p`.chomp
+    if ["x86_64"].include?(arch)
+      Dir.chdir(dir) do
+        puts("Building machinery-helper binary.")
+        system("go build")
+      end
+    else
+      puts("The hardware architecture #{arch} is not yet supported by the machinery-helper.")
```
cornelius: This and the other message should be recognizable as warnings and be put out on STDERR.

Rakefile
thardeck: Because it is part of packaging tasks and normally nearly all obs packages work with tarballs.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -264,7 +264,14 @@ def helper_usable?(helper)
   def run_helper_inspection(helper, filter, do_extract, file_store_tmp, file_store_final, scope)
     begin
       helper.inject_helper
-      helper.run_helper(scope)
+      if !helper.is_supported?
+        raise Machinery::Errors::UnsupportedHelperVersion.new(
+          "Error: machinery-helper is not compatible with the Machinery version" \
+          " #{Machinery::VERSION}. The versions need to match."
```
cornelius: The message should say what version of the helper was found.

tools/release.rb
```diff
@@ -23,96 +23,68 @@
...
-    @release_checks  = ReleaseChecks.new(@tag, @options[:jenkins_name])
-    @changelog_files = [@options[:changes_file], @options[:rpm_changes_file]]
-    @package_changes_path = "package/#{@options[:package_name]}.changes"
-    set_version
-    add_default_rpm_changes_entry
-  end
-
-  def check
-    @release_checks.check
+    @release_version = @options[:version]
```
cornelius: As you store the version in the `@release_version` variable, you should use that all over the place and not access @options[:version] in other places of the code.

It also would be better to have a method to access it, rather than using the member variable directly. This is clearer and more robust.

bin/machinery
thardeck: The check for file changes it quite powerful, it is only run on developer machines and it only takes 0,1 seconds to run.

Running rake also prevents all the methods getting parsed and kept in memory while regular machinery is run.

machinery-helper/.gitignore
thardeck: We decided to use machinery-helper instead so "go build" creates the correct file name without parameters.

lib/machinery_helper.rb
thardeck: This would make the logic more complicated without much gain for now and we decided to not share helpers between different architecture so there would be no use for it.

cornelius:
There are still two places, where the version is defined, right? The versions file in the go and the one in the ruby code.

plugins/unmanaged_files/unmanaged_files_inspector.rb
thardeck: Heh, I thought you wanted a trailing `if` if there is no else and that would have been ugly.

aduffeck: No, a trailing if should only be used if it fits on one line.

cornelius:
I think it might be worth to clean up this pull request a bit and separate the different parts to make review easier and the history clear. This is a big step, so the history should be meaningful.

It would be good to have one separate commit for the move of the helper (including the style fixes), and some additional commits to add the build system, adapt the release script, do the version check, ...

Rakefile
thardeck: The helper didn't had a man page directory before and the method should have worked on both.

thardeck:
```
There are still two places, where the version is defined, right? The versions file in the go and the one in the ruby code.
```
Yes, this is on purpose, because everything else we did come up with was a hack.

keberle94:
Started squashing on branch `incorporate_machinery_helper10`

PR #1442, opened 2015-10-05
Incorporate machinery helper

mschnitzer:
Supersedes #1441


docs/Inspection-Concept.md
```diff
@@ -63,10 +63,9 @@ self-contained binary without external dependencies. It does the gathering of
 information and most of the processing on the inspected machine and then returns
 the result to Machinery.
 
-For each platform there is a helper binary. The binary is bundled with the
-Machinery package and is installed to
-`/usr/share/machinery/helpers/<arch>/machinery-helper`. On the target system it
-is copied to the home directory of the root user. It is removed when the
+There is a helper binary which is bundled with the Machinery package and is installed
+to `<machinery-installation-path>/machinery-helper/machinery-helper`. On the target
```
cornelius: What is `<machinery-installation-path>`? This should be explained.

lib/local_system.rb
```diff
@@ -105,8 +105,12 @@ def validate_build_compatibility(system_description)
       end
     end
 
+    def matching_architecture?(arch)
```
cornelius: Wouldn't `matches_architecture?` be the better name? Then code using the methods would read naturally as `if system.matches_architecture?`.

lib/machinery_helper.rb
```diff
@@ -34,33 +31,42 @@ def initialize(s)
     @system = s
     @arch = @system.arch
 
-    @local_helpers_path = "/usr/share/machinery/helpers"
+    @local_helpers_path = File.join(Machinery::ROOT, "machinery-helper")
+    @remote_helper_path = File.join(Machinery::REMOTE_HELPER_PATH, "machinery-helper")
```
cornelius: There should be methods to access both paths and they should be used at all other places in this file instead of accessing the member variable directly.

lib/machinery_helper.rb
```diff
@@ -34,33 +31,42 @@ def initialize(s)
...
+    if File.exist?(local_helper_path) && LocalSystem.matching_architecture?(@arch)
+      true
+    else
+      false
+    end
   end
 
   def inject_helper
-    @system.inject_file(local_helper_path, Machinery::REMOTE_HELPERS_PATH)
+    @system.inject_file(local_helper_path, Machinery::REMOTE_HELPER_PATH)
```
cornelius: There is some confusion here about what a remote helper path is. Either it is the path to the binary (as in @remote_helper_path`) or it is the path to the directory where the helper is (as in `Machinery::REMOTE_HELPER_PATH`).

These should be two clearly distinguished names and they should use the same level of encapsulation (if one of them is encapsulated by a method, the other one should be as well).

lib/machinery_helper.rb
```diff
@@ -34,33 +31,42 @@ def initialize(s)
...
 
   # Returns true, if there is a helper binary matching the architecture of the
   # inspected system. Return false, if not.
   def can_help?
-    File.exist?(local_helper_path)
+    if File.exist?(local_helper_path) && LocalSystem.matching_architecture?(@arch)
+      true
+    else
+      false
+    end
```
cornelius: The if/else clause is not needed. It can just be `File.exist?(local_helper_path) && LocalSystem.matching_architecture?(@arch)`.

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -264,6 +264,13 @@ def helper_usable?(helper)
   def run_helper_inspection(helper, filter, do_extract, file_store_tmp, file_store_final, scope)
     begin
       helper.inject_helper
+      if !helper.helper_version_supported?
```
cornelius: This reads weird. Maybe better `if !helper.has_compatible_version?`?

cornelius:
This doesn't include the versioning check as we discussed it to simply take the git revision for checking compatibility of the helper and machinery. I think it would make sense to add it here before we merge this pull request.

bin/machinery
```diff
@@ -20,6 +20,12 @@
 require_relative '../lib/machinery'
 
 begin
+  if Dir.exist?(File.join(Machinery::ROOT, ".git"))
+    Dir.chdir(File.join(Machinery::ROOT, "machinery-helper")) do
+      Cheetah.run("rake")
```
cornelius: It might be more robus to explicitly call the build task.

thardeck:
The command output might change so I wanted to make sure.

machinery-helper/Rakefile
```diff
@@ -0,0 +1,123 @@
...
+  end
+  true
+end
+
+desc "Build the helper if needed"
+task :build do
+  raise "Error: Build of Machinery helper failed" if !run_build
+end
+
+desc "Don't fail on building errors"
```
cornelius: This needs to be "Don't fail in build errors". We are not building errors ;-)

plugins/unmanaged_files/unmanaged_files_inspector.rb
```diff
@@ -264,6 +264,13 @@ def helper_usable?(helper)
   def run_helper_inspection(helper, filter, do_extract, file_store_tmp, file_store_final, scope)
     begin
       helper.inject_helper
+      if !helper.has_compatible_version?
+        raise Machinery::Errors::UnsupportedHelperVersion.new(
+          "Error: machinery-helper is not compatible with the Machinery version" \
+          " #{Machinery::VERSION}. The versions need to match."
```
cornelius: The error message should not put out the Machinery version, but the helper version and what version was expected.

machinery-helper/Makefile
```diff
@@ -0,0 +1,4 @@
+all: build
+
+build:
+	go build
```
cornelius: The Makefile is not needed anymore.

cornelius:
Modulo the typo in the :try_build task description the pull request looks good to me now.

PR #1443, opened 2015-10-05
Fix wrong file type detection

### LINKED ISSUE #1154: xml files are not shown but only offered for download html view

mschnitzer:
This closes #1154


mschnitzer:
Superseded by #1446

PR #1448, opened 2015-10-06
Adapt os-inspection to new tumbleweed syntax

keberle94:


plugins/os/os_inspector.rb
```diff
@@ -99,6 +99,11 @@ def get_os_from_os_release
     # name instead of an abbreviation
     os = Os.for(result["pretty_name"] || result["name"])
     os.version = result["version"]
+
+    if os.version == "Tumbleweed"
```
thardeck: I would add an explanation as comment

thardeck:
lgtm

PR #1449, opened 2015-10-06
Fix wrong file type detection

### LINKED ISSUE #1154: xml files are not shown but only offered for download html view

mschnitzer:
This closes #1154 
Supersedes #1446


cornelius:
I'm not sure, if this patch works as expected. I tried with some example and it didn't work. I extracted the example to a unit test. Could you have a look at what's going on there?

cornelius:
Pushed the test again as it was deleted by force push: https://github.com/SUSE/machinery/tree/fix_wrong_filetype_detection_test

lib/helper.rb
```diff
@@ -20,6 +20,25 @@ def self.is_int?(string)
     (string =~ /^\d+$/) != nil
   end
 
+  def self.file_is_binary?(path)
+    # Code by http://www.thecodingforums.com/threads/test-if-file-is-binary.843447/#post-4572282
+    # Modified by SUSE Linux GmbH
+    ascii = control = binary = 0
+
+    File.open(path, "rb") { |io| io.read(1024) }.each_byte do |bt|
```
cornelius: What is `bt`? Please don't use abbrevations in variable names.

cornelius:
I inspected a system where I put an XML file, but it was shown as binary. The content of the file is this:

```xml
<?xml version='1.0' encoding='us-ascii'?>

<!--  A SAMPLE set of slides  -->

<slideshow 
    title="Sample Slide Show"
    date="Date of publication"
    author="Yours Truly"
    >

    <!-- TITLE SLIDE -->
    <slide type="all">
      <title>Wake up to WonderWidgets!</title>
    </slide>

    <!-- OVERVIEW -->
    <slide type="all">
        <title>Overview</title>
        <item>Why <em>WonderWidgets</em> are great</item>
        <item/>
        <item>Who <em>buys</em> WonderWidgets</item>
    </slide>

</slideshow>
```

lib/helper.rb
```diff
@@ -20,6 +20,25 @@ def self.is_int?(string)
     (string =~ /^\d+$/) != nil
   end
 
+  def self.file_is_binary?(path)
+    # Code by http://www.thecodingforums.com/threads/test-if-file-is-binary.843447/#post-4572282
+    # Modified by SUSE Linux GmbH
+    ascii = control = binary = 0
```
cornelius: Don't do parallel assignments. See https://github.com/bbatsov/ruby-style-guide#syntax

spec/unit/helper_spec.rb
```diff
@@ -68,6 +70,23 @@
     end
   end
 
+  describe ".content_is_binary?" do
+    use_given_filesystem
+
+    it "detects xml content as non-binary" do
+      expect(Machinery::content_is_binary?(File.read(given_file("xml_file")))).to be(false)
+      expect(Machinery::content_is_binary?(File.read(given_file("xml_file_2")))).to be(false)
```
cornelius: In which way is the test with `xml_file_2` different from the one with `xml_file`?

mschnitzer: I implemented that because we thought first, that there is an issue in our binary detect method. But it turned out that this ain't actually true.

cornelius:
lgtm

PR #1453, opened 2015-10-06
Speed up comparing unmanaged-files from O(n²) to O(n)

aduffeck:
Results from my machine, comparing a 580-file description
with a 2800-file one:

                  user     system      total        real
    new:       0.040000   0.000000   0.040000 (  0.036712)
    original: 43.340000   0.020000  43.360000 ( 43.745186)


keberle94:
lgtm

Issue #1155, opened 2015-07-22
Issues in HTML view


tgoettlicher:
Issues in the HTML view:
* some columns are cut-off and not resizeable 
* checksum is shown in `compare --html` view
* file content view doesn't scroll back to top when a new file is shown

eotchi:
Tracked in 
- https://trello.com/c/YxnQa7qH
- https://trello.com/c/NEMoY85W
- https://trello.com/c/OjlkqoCC

mauromorales:
All issues have been fixed at this point. Couldn't find the PR for the checksum but I've verified that it's in master.

tgoettlicher:
There were layout issues in the users section of the HTML view at the review meeting.

cornelius:
Let's close this and open more specific issues where needed.

PR #1454, opened 2015-10-06
File content view doesn't scroll back to top when a new file is shown

mschnitzer:
Related to #1155


aduffeck:
Works fine, lgtm.

PR #1459, opened 2015-10-06
Fix release_checks regressions

thardeck:
it looks like the checks were messed up during the release script revert.


mschnitzer:
lgtm - Please fix the Hound issues first

PR #1460, opened 2015-10-06
Fix architecture check in machinery-helper Rakefile

thardeck:
Make sure the architecture check is run first to prevent failures.
Fix missing variable usage.


keberle94:
lgtm

PR #1464, opened 2015-10-07
Limit gem updates to only patch-level and update supported gli version

keberle94:
Supersedes #1429


keberle94:
Will refactor commits to separate operations

PR #1468, opened 2015-10-07
Add documentation of packaging

cornelius:


docs/Packaging.md
```diff
@@ -0,0 +1,16 @@
+# Packaging
+
+For installation on user's machines Machinery is packaged as gem and as RPM. The
```
mschnitzer: Shouldn't it be `GEM` rather than `gem` ?

docs/Packaging.md
```diff
@@ -0,0 +1,16 @@
...
+systems, which have Ruby installed.
+
+The gem is published on RubyGems as
+[machinery-tool](https://rubygems.org/gems/machinery-tool). It includes the
+sources of the machinery helper, which are compiled on installation of the gem
+if there is a Go toolchain available.
+
+The RPM is maintained in the Open Build Service in the
+[systemsmamnagement:machinery](https://build.opensuse.org/project/show/systemsmanagement:machinery)
+project. The RPM bundles the gems it needs as well as the helper binary, so that
```
mschnitzer: I would say `The needed gems` or `The required gems`

docs/Packaging.md
cornelius: "gem" is not an acronym, so it should be ok as it is.

mschnitzer:
lgtm

PR #1473, opened 2015-10-08
Add information about possible warnings with local installed gems

keberle94:
We add additional information about warnings of possible conflicts from the rpm bundled gems
with the local installed gems to the packaging documentation.


docs/Packaging.md
```diff
@@ -14,3 +14,7 @@ The RPM is maintained in the Open Build Service in the
 project. The RPM bundles the required gems as well as the helper binary, so that
 there is one self-contained package, which can be easily installed, and doesn't
 expose any dependencies to the user.
+
+In a few cases this can cause warnings with your local installed gems.
+If you want to get rid of these warnings you just have to uninstall your local installed gem version.
+An example for these warnings would be the gem Nokogiri because of the two different compiled libraries.
```
cornelius: This is lacking some context and it is not very specific. We should clearly document what is happening there, and for the Nokogiri case be explicit about under which circumstances the issue happens.

cornelius:
lgtm

PR #1475, opened 2015-10-08
Add tests for helper rakefile

thardeck:


machinery-helper/Rakefile
```diff
@@ -15,90 +15,23 @@
...
-  if File.exist?(GIT_REVISION_FILE)
-    old_revision = File.read(GIT_REVISION_FILE)
-  else
-    old_revision = "unknown"
-  end
-  git_revision != old_revision
-end
 
 def run_build
+  helper_build = HelperBlueprint.new(HELPER_DIR)
```
cornelius: I don't get the naming here. What is the responsibility of `HelperBlueprint` and why is the variable called `helper_build`?

cornelius:
The file with the `HelperBuilder` class should be named `helper_builder.rb`.

cornelius:
Other than the file name the patch looks good to me.

PR #1481, opened 2015-10-09
Remove `-i` and `-p` from `show` and add a `--public` option to `serve`

### LINKED ISSUE #1341: Clean up binding the server for the HTML view to IP addresses

mschnitzer:
For cleaning up some network binding commands, I removed the `-i` and `-p` function from `show` and `compare`. Both functions will now only listen on `127.0.0.1` because they are not intended to be an alias for `serve`. Since no port can be specified, the port will be set by the configuration in the `http_server_port` section. For the `serve` command, I replaced the `-i` option with the new option `--public` which makes the server listen on all configured IP addresses on the local machine.

The man page is up to date. I didn't mention my changes in the `compare` man page because the network features are still experimental.

This fixes #1341

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Fix: Clean up binding the server for HTML view to IP addresses
+  (gh#SUSE/machinery#1341)
 * Fix: Make links to sections with common elements clearer in HTML comparison
   view
 * Fix: Scrolling issue of file view (gh#SUSE/machinery#1155)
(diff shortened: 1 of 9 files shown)


cornelius:
lgtm

PR #1483, opened 2015-10-09
Publish gem with release script

thardeck:


tools/release.rb
```diff
@@ -117,6 +117,18 @@ def publish_man_page
     Cheetah.run("git", "checkout", "master")
   end
 
+  def publish_gem
+    Dir.chdir(File.join(Machinery::ROOT, "package")) do
+      gem = Dir.glob("*.gem").first
```
cornelius: Does this always give the right gem?

Maybe it would be better to explicitly look for the version which is released.

cornelius:
Looks good to me.

PR #1487, opened 2015-10-12
Add missing helper requirement to gemspec

thardeck:
tools/build_helper.rb is required for the machinery-helper Rake task.


mauromorales:
LGTM

PR #1503, opened 2015-10-15
[Review] Request from 'aduffeck' @ 'SUSE/machinery/server_side_rendering_for_compare2'

aduffeck:
  * a871485 Add changelog entry for the improved comparison
  * 9215a7d Consider changed elements when comparing users
  * 1968e7c Consider changed elements when comparing unmanaged-files
  * 125bc12 Consider changed elements when comparing services
  * 45bef5b Consider changed elements when comparing repositories
  * b59f870 Consider changed elements when comparing patterns
  * f4c3fdd Consider changed elements when comparing groups
  * 43536be Consider changed elements when comparing config-files
  * d585fb9 Consider changed elements when comparing changed-managed-files
  * 4fc459e Also consider changed elements when comparing filescopes
  * 3874e16 Parameterize changed_elements so that it can be used for all scopes
  * 66e3de7 Extract PackagesModel#extract_changed_elements into Scope
  * 953664f Make Array#to_s return a meaningful representation of the content


cornelius:
#1506

cornelius:
I have a comparison here, which shows "0 changed" services. This should be hidden instead. Does this work for you?

thardeck:
So I have tested it a little bit and besides the already fixed bug it seems to work fine. There is quite some redundancy but this is supposed to be addressed in a later refactoring.

I would also improve the output of the different attributes part but this shouldn't be part of this pr.

html/partials/compare/repositories.html.haml
```diff
@@ -18,6 +18,11 @@
             - if @diff["repositories"].only_in2
               %span.summary-part
                 #{@description_b.name}: #{@diff["repositories"].only_in2.length} repositories
+            - if @diff["repositories"].changed
+              %span.summary-part
+                %a{ href: "#repositories_changed" }
+                  changed
+                = ": #{@diff["repositories"].changed.length} repositories"
```
cornelius: The text has the wrong form, when only one repository is changed. Then it says "1 repositories", but it should say "1 repository"

aduffeck: You're right, but I'd propose to tackle the pluralization in a separate pull request. That  one is quite big already and there are many other places also in the current code where this needs to be fixed.

Fine with you?

cornelius:
lgtm

PR #1505, opened 2015-10-16
Navigation for excluded scopes 2

### LINKED ISSUE #1199: Don't show navigation icons for non-existing scope in html view

mauromorales:
Supersedes #1504 fixes #1199


cornelius:
I think the look of the navigation was better before.

They were the same as the ones we use at the left side of the content, so there is a connection for the user between the navigation button and the target.

The buttons also were of equal size, which gives a more balanced look.

And their colors picked up the color scheme from the web site.

Generating the buttons from HTML instead of using a picture is a good idea, but I think we should recreate the look which was there.

mauromorales:
This is what it should actually look like:

navbar

Could we debug in your machine and see if machinery-base.css is properly loading and if it has the same values for `.btn-default` or if it's taking them from a cached or different file?

mauromorales:
didn't get the issue on another vm but on my laptop, so will debug there ^__^

PR #1511, opened 2015-10-16
Navigation for excluded scopes 4

### LINKED ISSUE #1199: Don't show navigation icons for non-existing scope in html view

mauromorales:
Supersedes #1509 fixes #1199


cornelius:
The alignment of the controls doesn't seem to work yet. See screenshot:

rendered1

mauromorales:
my bad, only  tested with full screen

cornelius:
Also the navigation buttons are slightly lower than the reset button, even when the browser is wide enough.

PR #1512, opened 2015-10-16
Navigation for excluded scopes 5

### LINKED ISSUE #1199: Don't show navigation icons for non-existing scope in html view

mauromorales:
Supersedes #1511 fixes #1199


mauromorales:
@cornelius I think this time should be correct. The input will grow/shrink depending on the window size and verified the position of the buttons with a grid

aligned

cornelius:
lgtm

PR #1514, opened 2015-10-16
Parse YAML files during man page generation

thardeck:
The plugin md files were replaced by a YAML file. The man page
generation wasn't adapted before.


tgoettlicher:
LGTM

PR #1516, opened 2015-10-19
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_151019_fix_pluralization_in_html_view'

aduffeck:
  * 4d7f4f0 Only show summary sections if there is according content
  * 4019918 DRY up the scope summary section. Fix pluralization.
  * dec2a05 Make `scope.length` also work for non-array scopes

### DIFF
--- a/html/partials/compare/repositories.html.haml
+++ b/html/partials/compare/repositories.html.haml
@@ -11,23 +11,10 @@
       .col-xs-11
         %h2
           Repositories
-          .scope-summary
-            - if @diff["repositories"].only_in1
-              %span.summary-part
-                #{@description_a.name}: #{@diff["repositories"].only_in1.length} repositories
-            - if @diff["repositories"].only_in2
-              %span.summary-part
-                #{@description_b.name}: #{@diff["repositories"].only_in2.length} repositories
-            - if @diff["repositories"].changed
-              %span.summary-part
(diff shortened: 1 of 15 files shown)


html/partials/compare/repositories.html.haml
```diff
@@ -11,23 +11,10 @@
...
-                = ": #{@diff["repositories"].changed.length} repositories"
-            - if @diff["repositories"].common
-              %span.summary-part
-                %a.show-common-elements{ href: "#repositories_both" }<
-                  both
-                = ": #{@diff["repositories"].common.length} repositories"
+          = render_partial "compare/summary",
+            scope: "repositories",
+            singular: "repository",
+            plural: "repositories"
```
mauromorales: Would it make sense to add the plural to the plugin yml file and extract it from there?

mauromorales:
lgtm besides my question

html/partials/compare/repositories.html.haml
aduffeck: Hm, to me the inflection doesn't really belong to the plugin definition file. 

What I have on my list for some time, though, is to make our pluralize method a bit smarter so that it can handle the `repository` -> `repositories` case. That would also simplify that code a bit.

PR #1517, opened 2015-10-19
Add package building to prophet integration test

thardeck:
We had two cases of broken integration tests because the package
building was broken in master.
This change should prevent it for most parts.


aduffeck:
lgtm

PR #1519, opened 2015-10-19
Html buttons

mauromorales:


spec/features/navigation_bar_spec.rb
```diff
@@ -20,6 +20,8 @@
   initialize_system_description_factory_store
 
   let(:store) { system_description_factory_store }
+  let(:nav_scopes) { all(".scope-navigation a.btn-sm:not(.disabled) span").map(&:text) }
+  let(:content_scopes) { all("#content_container a.btn-lg span").map(&:text) }
```
aduffeck: Those `let` declarations are a bit confusing syntactically imo. Usually `let` blocks define some fixture data to work with in the tests, but here you're processing the current state during the test when the `let` method is called first, so the value depends on where it is used.

I think it would be cleaner to change that to

```ruby
  def nav_scopes
    all(".scope-navigation a.btn-sm:not(.disabled) span").map(&:text)
  end

  def content_scopes
    all("#content_container a.btn-lg span").map(&:text)
  end
```

instead.

aduffeck:
LGTM besides the syntactic issue.

PR #1523, opened 2015-10-19
Uncollapse elements when 'both' link is used

### LINKED ISSUE #1469: in compare --html: when collapsing all scopes - the "both" link is still visible but does not work anymore
Howto reproduce:

o run `machinery compare --html system1 system2`
=> see a browser displaying the differences and common items
o click on 'collapse all'
=> see all scopes collapse their content
=> see each scope still have a 'both' link beside them
o click the 'both' link
=> nothing happens

thardeck:
Fixes #1469


mauromorales:
The "both" link works fine but the "arrow down" image should change to "arrow up"

See example for the packages scope:

mauromorales:
lgtm

PR #1527, opened 2015-10-20
Fix wrong alert when a scope is missing from both descriptions

### LINKED ISSUE #1508: There is an empty alert shown in the HTML comparison
See screenshot:

rendered1

It's comparing two descriptions which each hav a package, repositories, and an extracted unmanaged files scope.

aduffeck:
Fixes #1508

### DIFF
--- a/lib/server.rb
+++ b/lib/server.rb
@@ -231,7 +231,7 @@ class Server < Sinatra::Base
     Inspector.all_scopes.each do |scope|
       if @description_a[scope] && @description_b[scope]
         @diff[scope] = Comparison.compare_scope(@description_a, @description_b, scope)
-      else
+      elsif @description_a[scope] || @description_b[scope]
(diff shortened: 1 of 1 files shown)


lib/server.rb
```diff
@@ -231,7 +231,7 @@ def diff_to_object(diff)
     Inspector.all_scopes.each do |scope|
       if @description_a[scope] && @description_b[scope]
         @diff[scope] = Comparison.compare_scope(@description_a, @description_b, scope)
-      else
+      elsif @description_a[scope] || @description_b[scope]
         @meta[:uninspected] ||= Hash.new
 
         if !@description_a[scope] && @description_b[scope]
```
thardeck: Couldn't the later be now removed for both?

thardeck:
lgtm

PR #1528, opened 2015-10-20
Fix overlapping text and video on low-resolution screens

tgoettlicher:


tgoettlicher:
Asked Manuele for review via email.

manuelecarlini:
I tried on my machine and make a test on my server to access via mobile and it looks good

PR #1532, opened 2015-10-20
[Review] Request from 'greygoo' @ 'SUSE/machinery/review_151020_escape_file_download_urls'

greygoo:
  * 0365419 Escape file download URL so special characters don't cause errors.


html/assets/show/machinery.js
```diff
@@ -84,7 +84,7 @@ $(document).ready(function () {
     var file = $(this);
     var scope = file.parents(".scope").data("scope");
     var description = $("body").data("description");
-    var url = "/descriptions/" + description + "/files/" + scope + file.text().trim();
+    var url = escape("/descriptions/" + description + "/files/" + scope + file.text().trim());
```
aduffeck: `escape` was deprecated some time ago. You should use `encodeURI` instead.

PR #1533, opened 2015-10-20
Improve layout for show --verbose

mauromorales:


spec/unit/cli_spec.rb
```diff
@@ -451,9 +451,9 @@
           ])
 
           expect(captured_machinery_output).
-            to include("The following filters were applied before showing the description:")
+            to match(/  The following filters were applied before showing the description:/)
```
thardeck: Why using match instead of include?

thardeck:
lgtm

spec/unit/cli_spec.rb
mauromorales: I went with it just to try to get the format/style correct, does it make sense?

PR #1537, opened 2015-10-21
Add Changelog entry for gh#SUSE/machinery/1497.

greygoo:


NEWS
```diff
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Offer files containing '%' and other special characters for download
```
thardeck: "Offer files" sounds misleading, I would write something like:
"Files with special characters are now downloadable (..)" or
"File names are now escaped in HTML view (...)"

PR #1539, opened 2015-10-21
Add Changelog entry for gh#SUSE/machinery/1497.

greygoo:
Add Changelog entry for issue 1497


NEWS
```diff
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Files containing special characters can now be downloaded.
```
thardeck: we don't use dots at the end

PR #1544, opened 2015-10-22
Fix popover

### LINKED ISSUE #1542: Popup for Config File Diffs doesn't appear
Commit 7b26cc4243e4fee7ca3773da9a6de46ea2c0b98a broke the popup for config file diffs.

mauromorales:
fixes #1542


tgoettlicher:
Can you please provide a changelog entry, as this fixes an user-visible bug?

tgoettlicher:
As we haven't released a package after we introduced the bug, I don't insist on a changelog entry.

tgoettlicher:
LGTM

PR #1545, opened 2015-10-22
Don't fail if config file diffs don't exist

tgoettlicher:


mauromorales:
code lgtm but I wasn't able to reproduce, if I delete the file it still works and if I leave the file but remove it's contents it still breaks. I'm probably not looking at the right thing but wanted to point it out.

tgoettlicher:
I created a new pull request: https://github.com/SUSE/machinery/pull/1567
The commits contains an explanation in which scenarios the bug occurred.

You can reproduce it by deleting the file in `~/.machinery/hostname/analyze/file.diff`.

PR #1547, opened 2015-10-23
Fix big buttons not moving along when scrolling in compare view

mauromorales:
supersedes #1546

### DIFF
--- a/spec/features/feature_spec_helper.rb
+++ b/spec/features/feature_spec_helper.rb
@@ -48,13 +48,12 @@ Capybara.configure do |config|
   config.default_driver = :poltergeist
 end
 
-def get_position(selector)
-  pos = page.driver.evaluate_script <<-EOS
+def get_coordinates(selector)
+  page.driver.evaluate_script <<-EOS
     function() {
-      var elem1 = document.querySelector('#{selector}');
-      var rect1 = elem1.getBoundingClientRect();
-      return [rect1.left, rect1.top];
+      var found_element = document.querySelector('#{selector}');
+      var element_position = found_element.getBoundingClientRect();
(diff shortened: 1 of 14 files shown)


spec/features/feature_spec_helper.rb
```diff
@@ -47,3 +47,14 @@ def self.system_phantomjs_path
 
   config.default_driver = :poltergeist
 end
+
+def get_position(selector)
+  pos = page.driver.evaluate_script <<-EOS
+    function() {
+      var elem1 = document.querySelector('#{selector}');
```
aduffeck: It's just a cosmetic issue, but these variable names are not very nice.

spec/features/scope_buttons_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+      create_test_description(
+        scopes:        ["os", "packages", "repositories", "services"],
+        name:          "name",
+        store:         store,
+        store_on_disk: true
+      )
+    }
+
+    it "also move down" do
+      visit("/name")
```
aduffeck: The issue was about the compare view, no?

aduffeck:
The code looks good to me and works. Once those two small issues are tackled this can be merged from my pov.

spec/features/scope_buttons_spec.rb
mauromorales: that was the one that broke test ... but this one is because there was a regression in which the scope logos were not scrolling along when the user scrolled down

mauromorales: ahhh right ... I'll add one for the compare view

aduffeck:
lgtm

PR #1551, opened 2015-10-26
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_151026_make_comparison_view_full_width'

aduffeck:
  * 2588ca3 Highlight nav button of the current scope
  * 097371d Make the comparison view make use of all the horizontal space


html/assets/show/machinery.js
```diff
@@ -1,3 +1,23 @@
...
+function setCurrentScopeState(anchor) {
+  var header_height =  $("#nav-bar").height() + 20;
+  var pos = anchor.offset();
+  var top_pos = $(this).scrollTop() + header_height;
+  if(top_pos >= pos.top) {
+    anchor.addClass("over-the-top");
+  } else {
+    anchor.removeClass("over-the-top");
+  }
+}
```
thardeck: Isn't this code identical with the part of the other machinery.js? In general we might look into minimize redundancy between both files if possible.

thardeck:
I think two tables look worse in the show html view. One is packages because the size of name is way to big even if not needed and the other is groups were the users list is too big. For groups I would make all columns equal I think.

thardeck:
Except of my comments it looks good to me.

thardeck:
lgtm

PR #1557, opened 2015-10-27
Update machinery-helper README to match our current building process

mauromorales:


thardeck:
It is Ok, but the .git_revision file is also generated by the Rakefile so it sounds a bit misleading.

PR #1559, opened 2015-10-27
Create version.go even if .git_revision does already exist

thardeck:
If `rake build` is run the first time in machinery-helper the file
.git_revision and machinery-helper/version.go is created.
If version.go is removed but .git_revision stays building of
machinery-helper fails. This is fixed by this additional check.


mauromorales:
lgtm

PR #1567, opened 2015-10-28
Don't fail if config file diffs don't exist

tgoettlicher:
If the corresponding package for a changed config file cannot
be downloaded it's not possible to generate a diff for the config file.
In this case we cannot show the diff in the html view.


aduffeck:
I'm not sure if what you wrote in the commit message is the actual problem, because looking at the `server.rb` the diff only gets set if there is an according diff file available from the disk:

```ruby
file.diff = diff_to_object(File.read(path)) if File.exists?(path)
```

PR #1569, opened 2015-10-28
Add current outgoing ip to Autoyast Readme.md

### LINKED ISSUE #1543: README.md for autoyast2 generated profile is picking wrong IP address
On a system with many network interface, machinery seems to be picking one interface among many to fill the README.md example to use the autoyast profile.

It might be better to use the interface which is used as default route.

thardeck:
Fixes #1543


lib/autoyast.rb
```diff
@@ -101,6 +100,12 @@ def profile
     builder.to_xml
   end
 
+  def outgoing_ip
+    output = Cheetah.run("ip", "route", "ls", stdout: :capture)
```
tgoettlicher: There is a straight forward way of getting the outgoing ip address: 
`ip r get 1 | awk '{print $NF; exit}' `
or if you don't want to use `awk`
`ip r get 1` and use the last word of the first line only.

spec/unit/autoyast_spec.rb
```diff
@@ -136,4 +145,24 @@
...
+EOF
+    }
+
+    it "returns the current outgoing ip" do
+      expect(Cheetah).to receive(:run).with(
+        "ip", "route", "ls", stdout: :capture
+      ).and_return(ip_route_ls)
+      expect(autoyast.outgoing_ip).to eq("10.100.2.35")
+    end
+  end
```
tgoettlicher: Can you please add a test which handles the situation if no default route is set?

lib/autoyast.rb
thardeck: At least online I saw an output of `ip route` that didn't show the default route in the first line so I am not sure if it will always work. That's why I have created the regex.

tgoettlicher:
The code looks good to me. I wonder if it wouldn't make sense to also replace `/path/to/autoyast_export` by the actual path in `README.md`. I wouldn't insist on that, though.

tgoettlicher:
LGTM

PR #1574, opened 2015-10-29
Fix 1543 2

### LINKED ISSUE #1543: README.md for autoyast2 generated profile is picking wrong IP address

thardeck:
Squashed #1569 and Fixes  #1543


tgoettlicher:
LGTM

PR #1575, opened 2015-10-29
Fix help with default for analyze operation

cornelius:
The command line help message contained a confusing "(default: none)"
part. By letting the default value be set by GLI instead of doing it
in the code the message is less confusing now. Behavior is identical.


mauromorales:
lgtm

PR #1577, opened 2015-10-29
Remove experimental flag for HTML comparison

thardeck:


cornelius:
We should do this change only after we have fixed the bugs and made sure that there are no new blockers from the results of the usability test.

PR #1578, opened 2015-10-29
Fix html binary recognition

### LINKED ISSUE #1154: xml files are not shown but only offered for download html view

thardeck:
Fixes #1154


lib/scope_file_access_flat.rb
```diff
@@ -30,7 +30,7 @@ def binary?(system_file)
     path = system_file.scope.file_path(system_file)
     return false if File.zero?(path)
 
-    Machinery.content_is_binary?(File.read(path))
+    Machinery.content_is_binary?(File.read(path).slice(0, 1024))
```
aduffeck: Instead of reading the whole file into memory and passing on the small slice you can also just read the amount of data you want (e.g. `File.read(path, 1024)`).

lib/helper.rb
```diff
@@ -21,24 +21,18 @@ def self.is_int?(string)
...
-      when 0...32
-        control += 1
-      when 32...128
+      if (32..128).include?(byte)
         ascii += 1
       else
-        binary += 1
+        # count binary but exclude newlines and tabs
+        binary += 1 unless [9, 10, 13].include?(byte)
       end
```
cornelius: Why would you treat newlines and tabs differently than other printable characters? I would go with a simple

```ruby
printable = 0

content.each_byte do |byte|
  printable += 1 if [9,10,13].include?(byte) || (32..128).include?(byte)
end

printable.to_f / content.size < 0.9
```

or something like that.

cornelius:
When you fix the Hound issue this looks good to me.

PR #1579, opened 2015-10-30
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_151029_speed_up_config_files_and_changed_managed_files_inspection'

aduffeck:
  * 3863e64 Adapt sudo configuration to work with the changed RPM call
  * 3d81f72 Move file attribute detection from inspectors to RpmDatabase
  * 32d9aff Change changed-rpm-file detection to only call rpm -V once


lib/rpm_database.rb
```diff
@@ -37,25 +37,38 @@ def changed_files(&block)
...
-            package_version: package_version,
-            status:          "changed",
-            changes:         changes
-          )
-        end
-      end.compact
-    end.uniq
+    result = out.each_line.map do |line|
+      line.chomp!
+      if !line.match(/^[^ ]+[ ]+. \/.*$/)
```
mauromorales: could this also be a `next unless ...` ?

aduffeck: Good point, I'll change that.

mauromorales:
lgtm

PR #1581, opened 2015-10-30
Fix html binary recognition2

### LINKED ISSUE #1154: xml files are not shown but only offered for download html view

thardeck:
Squashed #1578 and fixes #1154


cornelius:
lgtm

PR #1583, opened 2015-10-30
Use file to recognize binary files

thardeck:


lib/helper.rb
```diff
@@ -21,13 +21,7 @@ def self.is_int?(string)
...
 
   def self.content_is_binary?(content)
-    printable = 0
-
-    content.each_byte do |byte|
-      printable += 1 if [9, 10, 13].include?(byte) || (32..128).include?(byte)
-    end
-
-    printable.to_f / content.size < 0.9
+    !Cheetah.run("file", "-", stdout: :capture, stdin: content).include?(" text")
```
aduffeck: I'd call `file` with the `-b` option which suppresses the filename in the output in order to not match a binary file with ` text` in its name.

aduffeck:
lgtm

Issue #1580, opened 2015-10-30
Machinery inspect  crash


jasper167:
When running "machinery inspect --extract-files HOSTNAME" the following out is given.

Inspecting os...                                                                                                                                                                                 
 -> Found operating system 'SUSE Linux Enterprise Server 11 SP4' version '11.4'.                                                                                                                 
Inspecting packages...                                                                                                                                                                           
 -> Found 967 packages.                                                                                                                                                                          
Inspecting patterns...                                                                                                                                                                           
 -> Found 0 patterns.                                                                                                                                                                            
Inspecting repositories...                                                                                                                                                                       
 -> Found 3 repositories.                                                                                                                                                                        
Inspecting users...                                                                                                                                                                              
 -> Found 32 users.                                                                                                                                                                              
Inspecting groups...                                                                                                                                                                             
 -> Found 52 groups.                                                                                                                                                                             
Inspecting services...                                                                                                                                                                           
 -> Found 149 services.                                                                                                                                                                          
Inspecting config-files...                                                                                                                                                                       
 -> Extracted 34 changed configuration files.
Inspecting changed-managed-files...                                                                                                                                                              
 -> Extracted 10 changed files.
Inspecting unmanaged-files...                                                                                                                                                                    
 -> Inspection failed!
                                                                                                                                                                                                 

Errors while inspecting unmanaged-files:
 -> The inspection failed because of the unexpected input data:
{:path=>"root/\\\\", :type=>:file, :size=>168463, :mode=>"644", :user=>"root", :group=>"root"}

Please file a bug report at: https://github.com/SUSE/machinery/issues/new

mschnitzer:
@jasper167 OK, I found the problem. It's about the directory name in `root/`. There is a directory called `\\` or something like that. I'll work on that.

PR #1589, opened 2015-11-03
Fix: Crash with special characters in folder name

mschnitzer:
Related to #1580


NEWS
```diff
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Fix: Crash during inspection of a folder which contains special characters
```
thardeck: please don't use colons

thardeck:
The inspection is still failing if I create a `\t\t\\` directory in `/root`. The command was `mkdir "\\t\\t\\"`

mschnitzer:
Superseded by #1594

PR #1591, opened 2015-11-03
Fix 1490

### LINKED ISSUE #1490: Add documentation for inspect-container when image name has a slash
Container images are commonly named as USERNAME/PROJECT. In order to use this name it's necessary to add the --name option and rename to something that doesn't have a slash. Otherwise machinery will say the system description name has invalid characters.

    $ machinery inspect-container xorilog/telegram
    Inspecting localhost:5000/myopensuse for os, packages, patterns, repositories, users, groups, services, config-files, changed-managed-files, unmanaged-files...
    
    Note: There are filters being applied during inspection. (Use `--verbose` option to show the filters)
    
    
    Inspecting os...
    System description name 'xorilog/telegram' is invalid. Only 'a-zA-Z0-9_:.-' are valid characters.

thardeck:
The cli tests are really bad to test raises. I have moved the raise to a separate method to be able to test it.

Fixes #1490


lib/cli.rb
```diff
@@ -154,6 +154,14 @@ def self.handle_error(e)
     true
   end
 
+  def self.check_container_name!(image, name)
+    if image.include?("/") && name.include?("/")
+      raise Machinery::Errors::InvalidCommandLine.new(
+        "Error: Slashes are not allowed in machinery descriptions. Please provide a" \
+          " name for the description with the parameter `--name=NAME`."
```
mauromorales: I'd state here what is the currently generated description name. We can reuse the message for invalid description names:

```
System description name 'mamorales/opensuse' is invalid. Only 'a-zA-Z0-9_:.-' are valid characters.
Please provide a valid name for the description with the parameter `--name=NAME`.
```

mauromorales:
lgtm

PR #1592, opened 2015-11-04
Fix 1490 2

### LINKED ISSUE #1490: Add documentation for inspect-container when image name has a slash

thardeck:
Squashed #1591 Fixes #1490

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -154,6 +154,16 @@ class Cli
     true
   end
 
+  def self.check_container_name!(image, name)
+    if image.include?("/") && name.include?("/")
+      raise Machinery::Errors::InvalidCommandLine.new(
+        "Error: System description name '#{name}' is invalid. By default Machinery" \
+          " uses the image name as description name if the parameter `--name` is not" \
+          " provided.\nIf the image name consist a slash the `--name=NAME` parameter" \
+          " is mandatory. Valid characters are 'a-zA-Z0-9_:.-'."
+      )
+    end
+  end
 
   on_error do |e|
     Cli.handle_error(e)
@@ -636,6 +646,8 @@ class Cli
       inspector_task = InspectTask.new
 
       name, scope_list, inspect_options, filter = parse_inspect_command_options(image, options)
+
+      check_container_name!(image, name)
 
       begin
         system.start
--- a/spec/unit/cli_spec.rb
+++ b/spec/unit/cli_spec.rb
@@ -75,6 +75,44 @@ describe Cli do
 
       run_command(["inspect-container", "docker_image_foo"])
     end
+
+    it "calls .check_container_name and raises when the image and name consist slashes" do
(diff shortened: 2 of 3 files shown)


lib/cli.rb
```diff
@@ -154,6 +154,16 @@ def self.handle_error(e)
     true
   end
 
+  def self.check_container_name!(image, name)
+    if image.include?("/") && name.include?("/")
```
aduffeck: @thardeck Wouldn't it make sense to validate the name using `SystemDescription.validate_name` instead of special casing the `/` case?

lib/cli.rb
```diff
@@ -154,6 +154,16 @@ def self.handle_error(e)
...
+      raise Machinery::Errors::InvalidCommandLine.new(
+        "Error: System description name '#{name}' is invalid. By default Machinery" \
+          " uses the image name as description name if the parameter `--name` is not" \
+          " provided.\nIf the image name consist a slash the `--name=NAME` parameter" \
```
aduffeck: There's a typo, I think `consist` should be `contains`?

spec/unit/cli_spec.rb
```diff
@@ -75,6 +75,44 @@
 
       run_command(["inspect-container", "docker_image_foo"])
     end
+
+    it "calls .check_container_name and raises when the image and name consist slashes" do
+      expect(Cli).to receive(:check_container_name!).with("docker/foo", "docker/foo")
+      # unlike the real world raise doesn't stop the program flow in this CLI tests and
+      # it is not possible to check for a raise with `run_command`
+      # expect_any_instance_of(InspectTask).not_to receive(:inspect_system)
```
aduffeck: Shouldn't this line be uncommented?

spec/unit/cli_spec.rb
```diff
@@ -75,6 +75,44 @@
...
+
+      run_command(["inspect-container", "docker/foo"])
+    end
+
+    it "calls .check_container and continues if only the image consists slashes" do
+      expect(Cli).to receive(:check_container_name!).with("docker/foo", "docker_foo")
+      expect_any_instance_of(InspectTask).to receive(:inspect_system).
+        with(
```
aduffeck: Since this test is only about verifying that `inspect_system` is called I'd omit the expectations on the arguments.

PR #1593, opened 2015-11-04
Fix misleading `port already in use` message

### LINKED ISSUE #1541: Machinery recommends to use `--port` option which does not exist
```
$ machinery show host --html 
Trying to start a web server for serving the description on http://127.0.0.1:7585/host.

The server can be closed with Ctrl+C.

Port 7585 is already in use.
Stop the already running server on port 7585 or specify a new port by using --port option.
$ machinery show --port 4567 host --html  
Unknown option --port

Run 'machinery show --help' for more information.
```

mschnitzer:
It's not possible to get the information which command was executed (show, serve, or compare) since the message gets printed in a thread. I decided to write an error message which can be used in both cases.

This fixes #1541


lib/html.rb
```diff
@@ -51,7 +51,7 @@ def self.run_server(system_description_store, opts, &block)
         rescue Errno::EADDRINUSE
           servefailed_error = <<-EOF.chomp
 Port #{Server.settings.port} is already in use.
-Stop the already running server on port #{Server.settings.port} or specify a new port by using --port option.
+Stop the already running server on port #{Server.settings.port} or specify a new port in the machinery configuration file by modifying `http_server_port`. For the `serve` command is also a `--port` option available.
```
mschnitzer: It's a HEREDOC so it should be fine in this case.

mauromorales:
I'm not sure if telling the user to reconfigure http-server-port is the right thing. If the user is used to looking at his/her system descriptions on localhost:7585 and starts a new browser tab s/he will be suggested to use the most common url which now will be wrong.

In my opinion, once we have a server running we should just point the user to the right url. So if the user has `http://localhost:7585/opensuse` and now wants to compare opensuse and sles he just needs to go to `http://localhost:7585/compare/opensuse/sles` there is no need to start another server.

mauromorales:
lgtm just needs to be squashed

mschnitzer:
Superseded #1601

PR #1599, opened 2015-11-05
Split line which is too long (over 100 characters)

mschnitzer:
It's just a style issue.

PR #1600, opened 2015-11-05
Split line which is too long (over 100 characters)

mschnitzer:
It's just a style issue.

Supersedes & Squashes #1599


thardeck:
lgtm

PR #1617, opened 2015-11-09
Fix 1612

### LINKED ISSUE #1612: config-files inspector is sometimes unable to set a type
I have inspected the Orthos host Beethoven which runs SLES11SP4.
The config-files inspector seems to be unable to set the type for the files `/etc/postfix/sender_canonical` and `/etc/postfix/virtual` which results in a validation error.

The result is the same with 1.14.2 and 1.16.0.

`machinery validate beethoven`
```
In scope config_files: The property #14 (files/changes) of type Hash did not match any of the required schemas.
In scope config_files: The property #15 (files/changes) of type Hash did not match any of the required schemas.
```

The according json entries are:
```
      {
        "name": "/etc/postfix/sender_canonical",
        "package_name": "postfix",
        "package_version": "2.9.4",
        "status": "changed",
        "changes": [
          "size",
          "md5"
        ],
        "mode": "644",
        "user": "root",
        "group": "root",
        "type": null
      },
      {
        "name": "/etc/postfix/virtual",
        "package_name": "postfix",
        "package_version": "2.9.4",
        "status": "changed",
        "changes": [
          "size",
          "md5"
        ],
        "mode": "644",
        "user": "root",
        "group": "root",
        "type": null
      },
```

thardeck:
Fixes #1612


spec/unit/rpm_database_spec.rb
```diff
@@ -220,6 +220,28 @@
...
+            mode:  "644",
+            user:  "root",
+            group: "root",
+            type:  "file"
+          }
+        ]
+      )
+    end
+
+    it "raises if the type is unknown" do
```
cornelius: Should be "raises error if ..."

lib/rpm_database.rb
```diff
@@ -132,11 +133,15 @@ def parse_stat_line(line)
       "dir"
     when "symbolic link"
       "link"
-    when "regular file"
+    when / file$/
```
cornelius: I think I would match the style check by extracting the translation of the type output to the string we use in the JSON to a separate method.

thardeck: We already have two methods just for stat line so I would prefer not to change it.

cornelius: Instead of regular expression matching I would just cover the two cases we know about explicitly. But maybe that's too strict. When going for a fuzzy match it might make more sense to only match for `file$`, not for the space as well.

cornelius: @thardeck In which way is having less methods better here?

thardeck:
It will be squashed in the new pr

PR #1620, opened 2015-11-09
Prevent identical scopes message in case of identical descriptions

### LINKED ISSUE #1613: Comparing the same system descriptions shows obsolete message
When comparing the same system descriptions it shows the following message.
```
Compared descriptions are identical.
Following scopes are identical in both descriptions: os,packages,patterns,repositories,users,groups,services,config_files,changed_managed_files,unmanaged_files
```
The first line is enough, the second is not needed if they are identical.

thardeck:
Fixes #1613

### DIFF
--- a/lib/compare_task.rb
+++ b/lib/compare_task.rb
@@ -71,7 +71,8 @@ EOF
       end
     end
     if identical && common_scopes
-      output = "Compared descriptions are identical.\n" + output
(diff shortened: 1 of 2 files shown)


mschnitzer:
There is an empty line left.
```
Compared descriptions are identical.

lines 1-2/2 (END)
```

Besides that, it looks good to me. (change log entry is missing)

PR #1623, opened 2015-11-10
Fix1615 2

### LINKED ISSUE #1615: empty but existing scopes do not show in the serve view
When a system is inspected and a scope that is not excluded has no results in it, this scope is not shown in the web view at all.
E.g. if you have no changed managed files but did run the inspection with -x, you will have an enabled icon called 'UMF' in the webpage header - but licking it will not do anything as the corresponding section below is missing completely. In contrast - on the commandline output that section is listed, it is empty beside the information "Files extracted: yes".
I think also in the web view there should be an empty section listed - or the icon on top be greyed out.

mauromorales:
fixes #1615 supersedes #1619


aduffeck:
LGTM and works.

But what about the other scopes, should we handle it for all of them? Or at least all where it could happen that the list is empty?

mauromorales:
@aduffeck good point ... I also realized we have to add it to the compare view as well ... I'll merge this when master isn't frozen any more in order to close the issue and start on a new PR to cover the rest

PR #1634, opened 2015-11-10
Fix remote dir line break

thardeck:
This fixes the remote_dir line break issue which looked like this:
```
  * /mnt/unmanaged/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 0 B
    Files: 1

  * /mnt/unmanaged/remote-dir/ (remote_dir)
  * /remote-dir/ (remote_dir)
  * /root/.bash_history (file)
    User/Group: root:root
    Mode: 600
    Size: 0 B
```

```
  * /mnt/unmanaged/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 0 B
    Files: 1

  * /mnt/unmanaged/remote-dir/ (remote_dir)

  * /remote-dir/ (remote_dir)

  * /root/.bash_history (file)
    User/Group: root:root
    Mode: 600
    Size: 0 B
```

I have changed the unmanaged-files renderer to relate to the extracted status so single lines don't get a break like it used to be before.
```
  * /etc/.pwd.lock (file)
  * /etc/ImageVersion (file)
  * /etc/YaST2/licenses/ (dir)
  * /etc/alternatives/openSUSE-default.xml (link)
  * /etc/auto.remote_dir (file)
```


plugins/unmanaged_files/unmanaged_files_renderer.rb
```diff
@@ -35,7 +35,7 @@ def content(description)
               puts "Files: #{p.files}" if p.files
             end
           else
-            item "#{p.name} (#{p.type})"
+            item "#{p.name} (#{p.type})\n"
```
aduffeck: That doesn't look correct. The `item` call should take care of adding the newline (or not) consistently across all scopes, that's the whole point of having the rendering framework around.

PR #1637, opened 2015-11-11
Handle empty scopes

mauromorales:


plugins/changed_managed_files/changed_managed_files_renderer.rb
```diff
@@ -28,11 +28,11 @@ def content(description)
     list do
       file_status = description["changed_managed_files"].extracted
       if !file_status.nil?
-        puts "Files extracted: #{file_status ? "yes" : "no"}"
-      end
-
-      if files && files.empty?
-        puts "There are no changed managed files."
+        if files.empty?
```
aduffeck: I think we should also show that message when the `file_status` is nil. It's probably a theoretical case but still... ;)

plugins/config_files/config_files_renderer.rb
```diff
@@ -31,10 +31,16 @@ def content(description)
 
     list do
       file_status = description["config_files"].extracted
+      files = description["config_files"].files
+
       if !file_status.nil?
-        puts "Files extracted: #{file_status ? "yes" : "no"}"
+        if files && files.empty?
```
aduffeck: Same here.

plugins/repositories/repositories_renderer.rb
```diff
@@ -20,7 +20,7 @@ def content(description)
     return unless description.repositories
 
     if description.repositories.empty?
-      puts "System has no repositories"
+      puts "System has no repositories."
```
aduffeck: I think this could be made consistent with the other messages (`There are no repositories`).

html/partials/config_files.html.haml
```diff
@@ -8,50 +8,56 @@
...
-                %th.medium Package Version
-                %th.medium Changes
-                %th.small Mode
-                %th.small User
-                %th.small Group
-            %tbody
-              - config_files.files.each do |file|
+        .col-xs-1
+        .col-xs-11
+        - if count == 0
```
aduffeck: This needs to be indented I think:

image

aduffeck:
I suppose you omitted the `users` and `groups` scopes on purpose?

mauromorales:
@aduffeck yup I thought it would be hard to have system without them @tgoettlicher even said he would invite to lunch if I found one ^__^

aduffeck:
@mauromorales ok, if he invites me as well that's fine with me ;)

PR #1642, opened 2015-11-12
Fix remote dir line break2

thardeck:
Squashed #1634

This fixes the remote_dir line break issue which looked like this:
```
  * /mnt/unmanaged/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 0 B
    Files: 1

  * /mnt/unmanaged/remote-dir/ (remote_dir)
  * /remote-dir/ (remote_dir)
  * /root/.bash_history (file)
    User/Group: root:root
    Mode: 600
    Size: 0 B
```

```
  * /mnt/unmanaged/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 0 B
    Files: 1

  * /mnt/unmanaged/remote-dir/ (remote_dir)

  * /remote-dir/ (remote_dir)

  * /root/.bash_history (file)
    User/Group: root:root
    Mode: 600
    Size: 0 B
```

I have changed the unmanaged-files renderer to relate to the extracted status so single lines don't get a break like it used to be before.
```
  * /etc/.pwd.lock (file)
  * /etc/ImageVersion (file)
  * /etc/YaST2/licenses/ (dir)
  * /etc/alternatives/openSUSE-default.xml (link)
  * /etc/auto.remote_dir (file)
```

### DIFF
--- a/spec/unit/unmanaged_files_renderer_spec.rb
+++ b/spec/unit/unmanaged_files_renderer_spec.rb
@@ -97,6 +97,28 @@ describe UnmanagedFilesRenderer do
     EOF
   }
 
+  let(:description_remote_dir) {
+    create_test_description(json: <<-EOF)
+    {
+      "unmanaged_files": {
+        "extracted": true,
+        "files": [
+          {
+            "name": "/mnt/unmanaged/remote-dir/",
+            "type": "remote_dir"
+          },
+          {
+            "name": "/etc/alternatives/awk",
+            "type": "link",
+            "user": "root",
+            "group": "root"
+          }
+        ]
+      }
+    }
+    EOF
+  }
+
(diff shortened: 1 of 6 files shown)


spec/unit/unmanaged_files_renderer_spec.rb
```diff
@@ -97,6 +97,28 @@
     EOF
   }
 
+  let(:description_remote_dir) {
```
mauromorales: Could this be created in the system description factory instead of having the json and then just call the scope?

thardeck: It could but I think it is better to refactor that for all.

mauromorales:
lgtm

PR #1643, opened 2015-11-12
Add SLED12 support (fixes issue#1638)

fcrozat:


cornelius:
Thanks for the patch. We are currently working on a better representation of our support matrix. We'll get back to the patch, when we are done with this. I hope that this doesn't take too long :wink:

PR #1651, opened 2015-11-16
Let Tumbleweed and Leap inherit from OsSuse instead of just Os

### LINKED ISSUE #1627: `export-autoyast` does not support openSUSE Leap
`machinery export-autoyast` does not support openSUSE Leap.

thardeck:
so the analyze and export should now work these.

Fixes #1627

### DIFF
--- a/plugins/os/os_model.rb
+++ b/plugins/os/os_model.rb
@@ -178,7 +178,7 @@ class OsOpenSuse13_2 < OsOpenSuse
   end
 end
 
-class OsOpenSuseTumbleweed < Os
+class OsOpenSuseTumbleweed < OsSuse
   def display_name
     "#{name} (#{architecture})"
   end
@@ -192,7 +192,7 @@ class OsOpenSuseTumbleweed < Os
   end
 end
 
-class OsOpenSuseLeap < Os
+class OsOpenSuseLeap < OsSuse
   def display_name
     "#{name} (#{architecture})"
   end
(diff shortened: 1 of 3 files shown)


tgoettlicher:
LGTM

tgoettlicher:
`export-kiwi` works now.

plugins/os/os_model.rb
```diff
@@ -202,7 +202,7 @@ def self.canonical_name
   end
 
   def self.buildable_systems
-    [OsOpenSuse13_2, OsOpenSuseLeap]
+    [OsOpenSuse13_1, OsOpenSuse13_2, OsOpenSuseLeap, OsOpenSuseTumbleweed, OsSles12]
```
mauromorales: just double checking here ... leap can build 13.1?

plugins/os/os_model.rb
```diff
@@ -188,11 +188,11 @@ def self.canonical_name
   end
 
   def self.buildable_systems
-    [OsOpenSuse13_2]
+    [OsOpenSuse13_2, OsOpenSuseTumbleweed, OsOpenSuseLeap]
```
mauromorales: Just double checking here ... Tumbleweed cannot build 13.1?

plugins/os/os_model.rb
thardeck: I haven't tested it but it is supposed to. We will test it manually.

plugins/os/os_model.rb
thardeck: It will most likely work but might change in the nearby future because of the rolling release character so I don't want to risk it. I would even remove 13.2.

mauromorales:
Lgtm

PR #1653, opened 2015-11-16
Add landing page

vwallfahrer:
Superseds #1652


html/landing_page.html.haml
```diff
@@ -0,0 +1,56 @@
...
+              %th Scopes
+              %th Host
+          %tbody
+            - @all_descriptions.each do |key, value|
+              %tr
+                %td
+                  %a{ :href => "/#{key}" }
+                    #{key}
+                %td #{value["date"]}
+                %td #{value["scopes"].join(", ")}
```
vwallfahrer: Does not work. Or I do something wrong

html/landing_page.html.haml
```diff
@@ -0,0 +1,56 @@
...
+    .container-fluid
+      #nav-bar
+        .row
+          .col-xs-1
+          .col-xs-10
+            %h1
+              System Descriptions
+        .row
+          .filter-input.col-md-3.col-md-offset-1
+            %input.col-md-8#filter{ :placeholder => "Type To Filter" }
```
mschnitzer: The "To" has to be in lowercase. See https://github.com/SUSE/doc-styleguide/blob/develop/prebuilt/current/styleguide_plain.pdf section 4.2 and the 5th point.

I know that we're also using it on the other sites but I'll open a PR for that.

mschnitzer:
Maybe a hint like "No results for your search." is better than an empty table.

vwallfahrer:
@mschnitzer Good point, I will do that

mschnitzer:
I can see only one entry in the host column. Is that a bug in the landing page or in Machinery? The landing page gets the host entry from the `meta` information of a description, right?

cornelius:
The overall approach of the patch is fine. A few comments:

* `machinery serve` (without argument) shows a misleading message now. This needs to be adapted.
* There needs to be an entry in the change log
* The times don't show a time zone on my machine. This can be confusing as they show UTC, not the local time
* The column with the host names is empty, although this information is available in the manifest

cornelius:
The "created by Machinery" text should be right-aligned. At the moment it's shown in the middle of the page, which looks weird.

snapshot1

html/landing_page.html.haml
```diff
@@ -0,0 +1,56 @@
...
+    %link{ :href => "assets/machinery.css", :rel => "stylesheet", :type => "text/css" }
+    %script{ :src => "assets/jquery-2.1.1.min.js" }
+    %script{ :src => "assets/transition.js" }
+    %script{ :src => "assets/collapse.js" }
+    %script{ :src => "assets/modal.js" }
+    %script{ :src => "assets/jquery.searcher.min.js" }
+    %script{ :src => "assets/machinery-base.js" }
+    %script{ :src => "assets/show/machinery.js" }
+    %script{ :src => "assets/bootstrap-tooltip.js" }
+    %script{ :src => "assets/bootstrap-popover.js" }
```
cornelius: Do you need all the assets?

I think we should either use a common template for all pages (which then could include everything for convenience) or we should be specific and only include the required parts.

html/landing_page.html.haml
cornelius: Have you tried `%td= value["scopes"].join(", ")`?

vwallfahrer: Oh yes, now I know my fault. I tried it with a whitespace between the td and the equal

vwallfahrer:
@cornelius why don't you see the hosts? I see all my hosts in the host column

mschnitzer:
@mschnitzer I also don't see it as I mentioned before.

cornelius:
@vwallfahrer I'm not sure why the host names don't show up. I think it would be good to refactor the code to let the system description provide the information about the last update and the host names. This actually is a property of the description itself. When you have done that and there are methods in the description and corresponding unit tests, it will be easier to debug this.

cornelius:
The code looks good now and the hosts also show up on my system now.

Please add some unit tests for the new methods in the `SystemDescription` class.

When you have done that and rebased the pull request based on the current state of master, I think it's ready to be merged.

Issue #1648, opened 2015-11-14
Don't show hint for `show`, when `inspect` is run with `--show` option


cornelius:
When running `machinery inspect SOMETHING --show` the output shows the hint:

```
Hint: To show the data of the system you just inspected run:
machinery show localhost
```

This should only be shown when the `--show` option is not given.

PR #1655, opened 2015-11-16
Don't show hint for `show`, when `inspect` is run with `--show` option

mschnitzer:
Related to #1648


lib/cli.rb
```diff
@@ -616,7 +616,9 @@ def self.parse_inspect_command_options(host, options)
         inspect_options
       )
 
-      Hint.print(:show_data, name: name)
+      if options[:show] == false
+        Hint.print(:show_data, name: name)
```
thardeck: if should be postpositioned if there is only one line

spec/integration/support/command_line_examples.rb
```diff
@@ -68,6 +68,26 @@
           )
         ).to fail.and include_stderr("The following scope is not supported: foobar")
       end
+
+      it "does not show a hint that the inspection result can be shown by using the show " \
```
thardeck: I would reduce the size of the description, for example `does not print show hint if --show is used`

mschnitzer:
Superseded by #1664

PR #1664, opened 2015-11-17
Don't show hint for `show`, when `inspect` is run with `--show` option

mschnitzer:
Related to #1648
Squashes & Supersedes #1655


thardeck:
lgtm

PR #1665, opened 2015-11-17
Escape * in man page properly

mschnitzer:


aduffeck:
That fixes the issue but it makes the asterisks to be highlighted. Can't they be escaped without highlighting somehow?

mschnitzer:
Superseded by #1666

PR #1668, opened 2015-11-18
Add btrfs subvolumes to leap images

thardeck:


cornelius:
@thardeck Could you explain how this setup works? `systemdisk` is adding a LVM based layout. How does this interact with the btrfs setup you write to `/etc/fstab`.

Is this the actual set of sub volumes on Leap? On my Tumbleweed system I have some more.

thardeck:
Systemdisk in Kiwi with Btrfs provides subvolumes: https://doc.opensuse.org/projects/kiwi/doc/#chap.partitions.btrfs

The reason why I have less subvolumes is because Kiwi can only create subvolumes for existing directories and all the databases don't exist in our image. I could fake it of course with mkdir.

cornelius:
I see. The Kiwi documentation is a bit misleading there.

Why do you need to write `/etc/fstab`, if the configuration is done by Kiwi?

thardeck:
Because Kiwi only adds the subvolumes but the installer also adds these fstab entries. The reason for this is according to Arvin the usage of snapshots, because in case of a rollback, the subvolumes would be emptied without mont points.

cornelius:
Ok, the patch looks good to me, if somebody else has run the code.

mschnitzer:
@cornelius I ran the code and I could build and start the image successfully.

thardeck:
As a reference the Jeos images are also build with Kiwi and btrfs and are accepted by PM: https://build.suse.de/package/show/SUSE:SLE-12:Update:JeOS/SLES12-JeOS-for-kvm-and-xen

PR #1670, opened 2015-11-18
Add note for `serve --public` that every description is accessible

mschnitzer:
Supersedes #1669


lib/cli.rb
```diff
@@ -920,6 +920,8 @@ def self.parse_inspect_command_options(host, options)
 
       if options[:public]
         ip = "0.0.0.0"
+
+        Hint.print(:serve_public_option)
```
aduffeck: I think this should rather be a warning instead of a hint so that it is shown even if the user disabled the hints.

spec/unit/cli_spec.rb
```diff
@@ -451,6 +451,26 @@
...
+          "description in ~/.machinery publicly available! Be careful if you have descriptions, " \
+          "what should not be available for everyone.")
+      end
+
+      it "prints not a hint when --public is not used" do
+        allow_any_instance_of(ServeHtmlTask).to receive(:serve)
+
+        run_command(["serve", "description1", "--port=3000"])
+
+        expect(captured_machinery_output).not_to include("Hint: The --public option makes EVERY " \
```
aduffeck: I'd make this check less strict so that it also works if the text is changed.

lib/hint.rb
```diff
@@ -83,5 +83,10 @@ def share_html_contents(options)
       assets_dir = File.join(dir, "assets/")
       "Share the file: '#{index_file}' and the directory: '#{assets_dir}' as needed."
     end
+
+    def serve_public_option(_options)
+      "The --public option makes EVERY description in ~/.machinery publicly available! Be " \
+        "careful if you have descriptions, what should not be available for everyone."
```
aduffeck: That reads a bit rough imo. Maybe `The --public option makes ALL of your system descriptions publicly available. Take care if there are system descriptions that should not be read by others.`?

spec/unit/cli_spec.rb
```diff
@@ -451,6 +451,26 @@
...
+        allow_any_instance_of(ServeHtmlTask).to receive(:serve)
+
+        run_command(["serve", "description1", "--port=3000", "--public"])
+
+        expect(captured_machinery_output).to include("Hint: The --public option makes EVERY " \
+          "description in ~/.machinery publicly available! Be careful if you have descriptions, " \
+          "what should not be available for everyone.")
+      end
+
+      it "prints not a hint when --public is not used" do
```
aduffeck: `does not print...` :)

mschnitzer:
Superseded by #1680

Issue #1632, opened 2015-11-10
`analyze` prints wrong hint if system description is from a container


tgoettlicher:
```
$ machinery analyze opensuse 
The scope 'config-files' is part of the system description but the corresponding files weren't extracted during inspection.
The files are required to continue with this command. Run `machinery inspect --extract-files --scope=config-files opensuse` to extract them.
```

Instead of `inspect` it needs to read `inspect-container`.

PR #1676, opened 2015-11-19
Adjust hint in `analyze` command

mschnitzer:
Issue #1632 
Supersedes #1675


spec/unit/exceptions_spec.rb
```diff
@@ -46,6 +46,20 @@
       )
   end
 
+  it "recognizes if a description is a container and adapts the output message" do
+    description["environment"]["system_type"] = "docker"
```
mauromorales: I believe there is already a factory for docker. Could we use that instead?

mschnitzer:
Superseded by #1678

PR #1673, opened 2015-11-19
Show error instead of exception while creating the `--image-dir` folder

mschnitzer:

Issue #1667, opened 2015-11-18
permission denied error instead of info message


jubalh:
If I am as a normal user at `/` and give a relative path (which does not exist yet) I get:

```
machinery build --image-dir test current
Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
/usr/lib64/ruby/2.2.0/fileutils.rb:252:in `mkdir': Permission denied @ dir_s_mkdir - /test (Errno::EACCES)
        from /usr/lib64/ruby/2.2.0/fileutils.rb:252:in `fu_mkdir'
        from /usr/lib64/ruby/2.2.0/fileutils.rb:226:in `block (2 levels) in mkdir_p'
        from /usr/lib64/ruby/2.2.0/fileutils.rb:224:in `reverse_each'
        from /usr/lib64/ruby/2.2.0/fileutils.rb:224:in `block in mkdir_p'
        from /usr/lib64/ruby/2.2.0/fileutils.rb:210:in `each'
        from /usr/lib64/ruby/2.2.0/fileutils.rb:210:in `mkdir_p'
        from /usr/lib64/ruby/gems/2.2.0/gems/machinery-tool-1.16.1/lib/build_task.rb:31:in `build'
        from /usr/lib64/ruby/gems/2.2.0/gems/machinery-tool-1.16.1/lib/cli.rb:334:in `block (2 levels) in <class:Cli>'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/command_support.rb:126:in `call'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/command_support.rb:126:in `execute'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/app_support.rb:296:in `block in call_command'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/app_support.rb:309:in `call'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/app_support.rb:309:in `call_command'
        from /usr/lib64/ruby/gems/2.2.0/gems/gli-2.13.1/lib/gli/app_support.rb:83:in `run'
        from /usr/lib64/ruby/gems/2.2.0/gems/machinery-tool-1.16.1/bin/machinery:34:in `<top (required)>'
        from /usr/bin/machinery:23:in `load'
[...]

PR #1677, opened 2015-11-19
Show error instead of exception while creating the `--image-dir` folder

mschnitzer:
Issue #1667 
Squashes #1673


aduffeck:
lgtm

PR #1684, opened 2015-11-23
Fix disk-full problem in the ubuntu VM

aduffeck:
The problem was not that the disk was too small, but that the removal of
the temporary /EMPTY file, which is used to fill the disk with zeros for
optimized storage, was not (always) commited to disk before shutting
down the machine, causing the disk to be completely filled.


thardeck:
lgtm

PR #1687, opened 2015-11-23
Move warn message to a better location in `serve`

mschnitzer:
The warn message that all descriptions are accessible over HTTP got moved to a better location.


mschnitzer:
Superseded by #1688

PR #1688, opened 2015-11-23
Move warn message to a better location in `serve`

mschnitzer:
The warn message that all descriptions are accessible over HTTP got moved to a better location.

Supersedes #1687


lib/html.rb
```diff
@@ -32,6 +32,8 @@ def self.run_server(system_description_store, opts, &block)
 Warning:
 The server is listening on all configured IP addresses.
 This could lead to confidential data like passwords or private keys being readable by others.
+
+The --public option makes ALL of your system descriptions publicly available. Take care if there are system descriptions that should not be read by others!
```
mschnitzer: It's a HEREDOC - it doesn't matter.

aduffeck: Shouldn't this warning only be shown when the `--public` option is used?

mschnitzer:
Superseded by #1700

PR #1693, opened 2015-11-23
Add sled support

thardeck:
Supersedes #1643


thardeck:
Since we are planning to get rid of the supported message and change from a whitelist to a blacklist approach this pr might not be handled until the end of the sprint, which is a maximum of three weeks this time.

PR #1700, opened 2015-11-25
Show better warn messages to users when using the `--public ` option in `serve`

mschnitzer:
Supersedes #1688


mschnitzer:
Superseded by #1701

PR #1701, opened 2015-11-25
Rephrase startup and warn messages when starting an HTTP server

mschnitzer:
Supersedes #1700


spec/integration/support/serve_examples.rb
```diff
@@ -74,6 +74,54 @@ def test_basic_html(port)
...
+
+      # wait until the process quits and writes the response of the program
+      # into 'res'
+      times = 0
+      while res.nil? && times < 10
+        times += 1
+        sleep(1)
+      end
+
+      expect(res.stderr).to include("The --public option makes the HTTP server listen on all " \
```
aduffeck: This is again too specific, I'd only check for `--public option` or so, so that the test doesn't fail if small details of the error message are changed.

lib/html.rb
```diff
@@ -26,19 +26,26 @@ def self.run_server(system_description_store, opts, &block)
       Server.set :public_folder, File.join(Machinery::ROOT, "html")
       Server.set :static_cache_control, "no-cache"
 
-      if opts[:ip] != "localhost" && opts[:ip] != "127.0.0.1"
+      if opts[:public]
```
aduffeck: As described I'd catch the case that both `public` and an `ip` is given and raise an exception because it's not clear what the intended behavior is in that case.

I also think the if structure could be flattened to something like

```ruby
if opts[:public] && opts[:ip]
  raise ...
elsif opts[:public]
  ...
elsif opts[:ip] == "0.0.0.0"
  ...
elsif opts[:ip] != "localhost" && opts[:ip] != "127.0.0.1"
  ...
end
```

lib/cli.rb
```diff
@@ -920,10 +920,6 @@ def self.parse_inspect_command_options(host, options)
 
       if options[:public]
```
aduffeck: Here we map the `:public` option to an ip address and in `ServeHtmlTask` it is done the other way around. This could be simplified by directly passing on the `:public` option.

mschnitzer:
Superseded by #1703

PR #1705, opened 2015-11-26
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_151126_adapt_packages_data_model'

aduffeck:
  * 4802858 Fix hound issues
  * 7cf54df Migrate fixture system descriptions
  * 29e10f9 Adapt packages data model for v6 (to add support for dpkg packages)
  * 4326ecb Add unchanged schema files for v6
  * e6166ea Add migration for format version 6

### DIFF
--- a/html/partials/packages.html.haml
+++ b/html/partials/packages.html.haml
@@ -23,7 +23,7 @@
                   %th.medium Vendor
                   %th Checksum
               %tbody
-                - packages.each do |package|
+                - packages.packages.each do |package|
                   %tr
                     %td= package.name
                     %td= package.version
--- a/Rakefile
+++ b/Rakefile
@@ -186,7 +186,7 @@ task :upgrade_test_descriptions do
   # we don't want to upgrade format_vX, invalid-json, so we list the ones
   # we want to upgrade here
   descriptions = [
-    "opensuse131-build",
+    "opensuse_leap-build",
     "validation-error"
   ]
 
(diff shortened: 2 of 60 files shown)


html/partials/packages.html.haml
```diff
@@ -23,7 +23,7 @@
                   %th.medium Vendor
                   %th Checksum
               %tbody
-                - packages.each do |package|
+                - packages.packages.each do |package|
```
mauromorales: I know this is the code following the model but it just looks bad to me to have packages.packages, what do you think?

aduffeck: I agree that it looks a bit weird. We have the same problem with `services.services` and in a less dramatic form with `config_files.files` etc.

Unfortunately I don't know of a good way to solve that right now...

mauromorales:
lgtm

html/partials/packages.html.haml
mauromorales: got it, ok let's look at it later on

cornelius:
Shouldn't this go to the `ubuntu_inspection` branch instead of `master`?

aduffeck:
Yes, definitly. I just checked my history and I told git review to do so, but apparently that's not the proper way to do it:

`TARGET_BRANCH=ubuntu_inspection git review create`

I'll create a new pull request to the right branch once manual fixed the problem with the integration tests.

PR #1706, opened 2015-11-27
Fix `serve` --public option

mschnitzer:
The option was broken before because the IP address was set after the
warning and that's wrong.

### DIFF
--- a/lib/html.rb
+++ b/lib/html.rb
@@ -25,12 +25,6 @@ class Html
     end
 
     Thread.new do
-      Server.set :system_description_store, system_description_store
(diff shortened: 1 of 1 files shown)


lib/html.rb
```diff
@@ -52,6 +46,12 @@ def self.run_server(system_description_store, opts, &block)
         opts[:ip] = "127.0.0.1"
       end
 
+      Server.set :system_description_store, system_description_store
+      Server.set :port, opts[:port] || Machinery::Config.new.http_server_port
+      Server.set :bind, opts[:ip] || "localhost"
```
thardeck: The `opts[:ip]` is obsolete

thardeck:
lgtm

PR #1711, opened 2015-11-27
Services scope for ubuntu systems

greygoo:


cornelius:
Looks good to me.

It actually works out of the box with the Ubuntu Vagrant box I used to inspect.

I suppose QA will cover other scenarios which potentially might create problems.

PR #1715, opened 2015-11-27
Inspect config files ubuntu3

tgoettlicher:


cornelius:
I wouldn't rename the `RpmDatabase` class. The name is concise and accurate and changing it adds quite a bit of noise to the patch.

lib/dpkg_database.rb
```diff
@@ -0,0 +1,54 @@
...
+    Machinery.logger.warn(message)
+    Machinery::Ui.warn("Warning: #{message}")
+
+    @system.run_command("dpkg", "--verify", stdout: :capture)
+  end
+
+  def package_for_file_path(file)
+    package_name = @system.run_command("dpkg", "-S", file, stdout: :capture).split(":").first
+    package_details = @system.run_command("dpkg", "-s", package_name, stdout: :capture)
+    package_version = package_details.lines.find { |line| line =~ /^Version: / }.
```
mauromorales: Not sure but if the string match method would work here I think it would be a bit more readable, something along the lines of

```
package_details.match(/^Version: (.*)$/)[1]
```

lib/managed_files_database.rb
```diff
@@ -0,0 +1,200 @@
...
+
+      path_data[pkg.name].each do |key, value|
+        pkg[key] = value
+      end
+    end
+
+    @changed_files = result
+  end
+
+  def parse_rpm_changes_line(line)
```
mauromorales: Just from the name this seems like it belongs to `RpmDatabase` class

lib/system.rb
```diff
@@ -151,7 +151,15 @@ def locale
...
 
-  def rpm_database
-    @rpm_database ||= RpmDatabase.new(self)
+  def managed_files_database
+    if @managed_files_database
+      return @managed_files_database
+    else
+      if has_command?("rpm")
+        @managed_files_database = RpmDatabase.new(self)
+      else
```
mauromorales: is it safe to assume everything else falls into the dpkg group? Maybe we could specifically have the cases for rpm and dpkg and anything else would raise

tgoettlicher:
superseded by https://github.com/SUSE/machinery/pull/1725

PR #1716, opened 2015-11-30
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_151126_adapt_packages_data_model2'

aduffeck:
  * 451e1ac Migrate fixture system descriptions with adapted migration
  * 3bbb3e5 Make use of the new Machinery::Array attributes in the packages scope
  * 346bb8e Extend Machinery::Arrays so that they can also store attributes
  * 4802858 Fix hound issues
  * 7cf54df Migrate fixture system descriptions
  * 29e10f9 Adapt packages data model for v6 (to add support for dpkg packages)
  * 4326ecb Add unchanged schema files for v6
  * e6166ea Add migration for format version 6
  * 6112791 Merge pull request #1703 from SUSE/better_warn_message_6
  * b50552e Add new unit test for the Html class
  * 0773c51 Changelog: Rephrase startup and warn messages for HTTP server
  * e541c26 Add new integration tests for the HTTP server warn messages
  * 34a351b Remove old unit test code
  * 94097f5 Clean up CLI code for `serve`
  * ba11b68 Clean up HTTP server code and its messages

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -186,7 +186,7 @@ task :upgrade_test_descriptions do
   # we don't want to upgrade format_vX, invalid-json, so we list the ones
   # we want to upgrade here
   descriptions = [
-    "opensuse131-build",
+    "opensuse_leap-build",
     "validation-error"
   ]
 
--- a/html/partials/compare/package_list.html.haml
+++ b/html/partials/compare/package_list.html.haml
@@ -9,7 +9,7 @@
       %th.medium.visible-lg Checksum
       %th.tiny.hidden-lg
   %tbody
-    - list.each do |package|
+    - list.packages.each do |package|
       %tr
         %td= package.name
         %td= package.version
--- a/html/partials/compare/packages.html.haml
+++ b/html/partials/compare/packages.html.haml
@@ -12,10 +12,27 @@
       .col-xs-11
         %h2
           Packages
-          = render_partial "compare/summary",
-            scope: "packages",
-            singular: "package",
-            plural: "packages"
+          .scope-summary
+            - if @diff["packages"].only_in1 && @diff["packages"].only_in1.length > 0
+              %span.summary-part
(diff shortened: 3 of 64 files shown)


aduffeck:
**Note** The tests are failing because all `Machinery::Arrays` are now serialized using the new format, but the code has only been adapted for the `packages` scope.

cornelius:
Commenting on the new array representation:

I think the general concept is quite nice. The code reads much better than before and the representation still is expressive and straightforward.

One open question seems to be how to define the list of possible attributes. It's defined on reading the JSON or creating the object with a hash, but it would be nice, if you could also do something like:

```ruby
packages = PackagesScope.new
packages.package_system = "rpm"
packages.push(allmypackages)
```

Maybe the possible attributes should be set in the constructor of the sub class of the Array which is used for the specific purpose.

What seems to be missing from the current implementation is a setter for the attributes. I suppose we need that.

Personally I would prefer calling the special elements in the array representation in JSON "attributes" and "elements" instead of the versions with the underscores. This makes the JSON more natural to read, and I don't see a big benefit in seeing in the JSON, that these are special attributes.

As already discussed in person, it might make the code more simple and consistent, if we would move all scopes to being arrays with attributes.

aduffeck:
@cornelius I added an explicit `has_properties`  method and support for setting the attributes to the Array class in the last commit, does that work for you?

I'd still prefer the underscore versions of the names. I think it's actually ok if the JSON doesn't read that natural anymore because it is not really natural JSON. It is a JSON object which ends up as an Array object in Machinery and I think it's good to differentiate those special cases in the json.

cornelius:
With the method to define the attributes the concept looks good to me.

I don't have a strong opinion about the underscore in the JSON names. If you think it's better, I'm ok with it.

PR #1725, opened 2015-12-01
Enable inspection of config-files for Ubuntu hosts

tgoettlicher:


lib/managed_files_database.rb
```diff
@@ -0,0 +1,200 @@
...
+      end
+    end
+
+    @changed_files = result
+  end
+
+  def parse_changes_line(line)
+    # rpm provides lines per config file where first 9 characters indicate which
+    # properties of the file are modified
+    @rpm_changes, *fields = line.split(" ")
```
mauromorales: maybe we can rename this so it doesn't imply it's aware of `rpm`

spec/unit/rpm_database_spec.rb
```diff
@@ -18,245 +18,59 @@
...
-    end
-
-    it "it adds other_rpm_changes in case of an unexpected new/additional rpm tag" do
-      line = "S........N  c /etc/pulse/client.conf"
-      file, changes, flag = subject.parse_rpm_changes_line(line)
-      expect(file).to eq("/etc/pulse/client.conf")
-      expect(changes).to match_array(["size", "other_rpm_changes"])
-      expect(flag).to eq("c")
-    end
+  describe "#managed_file_list" do
```
mauromorales: maybe `managed_files_list` ?

lib/managed_files_database.rb
tgoettlicher: dpkg prints rpm format. That's why it's called @rpm_changes.

mauromorales: fair enough ;)

tgoettlicher:
Squashed: https://github.com/SUSE/machinery/pull/1726

Issue #1610, opened 2015-11-09
raise InspectionFailed "Yum repository baseurl is missing. Metalinks are not supported at the moment."  centos system


zhailibao2013:
when test in centos system,
.
Errors while inspecting repositories:
 -> Yum repository baseurl is missing. Metalinks are not supported at the moment.

some yum repository not contain baseurl,but mirrorlist

mauromorales:
Hi @zhailibao2013 

Unfortunately we don't support CentOS just yet so I can't promise we'll get this fixed. Nevertheless if you help me out following up on this issue I'll be happy to look into it. 

Could you send me the CentOS version you are using and if possible how your yum repositories look like. Feel free to change any URL you don't want to make public, I just need to see an example so I can try to replicate locally.

Cheers,

zhailibao2013:
thanks for your replay.
my centos version is 7.my yum repo file like this

# CentOS-Base.repo
#
# The mirror system uses the connecting IP address of the client and the
# update status of each mirror to pick mirrors that are updated to and
# geographically close to the client.  You should use this for CentOS updates
# unless you are manually picking other mirrors.
#
# If the mirrorlist= does not work for you, as a fall back you can try the 
# remarked out baseurl= line instead.
#
#
[base]
name=CentOS-$releasever - Base
mirrorlist=http://mirrorlist.centos.org/?release=$releasever&arch=$basearch&repo=os&
infra=$infra
#baseurl=http://mirror.centos.org/centos/$releasever/os/$basearch/
gpgcheck=1
gpgkey=file:///etc/pki/rpm-gpg/RPM-GPG-KEY-CentOS-7

#released updates 
[updates]
...............


please support mirrorlist，not only baseurl in yum repository

PR #1727, opened 2015-12-01
Ubuntu inspection with repos

thardeck:
The last commit is not directly related to the repo changes but without it the message about empty repositories isn't shown. This can be fixed by adapting the repository renderer code but I think the implementation of the `empty?` method was wrong, hence the fix.

The doc files also need to be adapted but this can be done in a separate pull request for all scopes.


plugins/repositories/repositories_inspector.rb
```diff
@@ -90,7 +92,42 @@ def inspect_yum_repositories
...
+    RepositoriesScope.new(repositories, repository_system: "yum")
+  end
+
+  def inspect_apt_repositories
+    content = system.read_file("/etc/apt/sources.list")
+    begin
+      content += "\n" + system.run_command(
+        "bash", "-c", "cat /etc/apt/sources.list.d/*.list", stdout: :capture
+      )
+    rescue Cheetah::ExecutionFailed
```
aduffeck: Why do we need this? This can easily hide errors and lead to weird and hard to debug behavior, so if there's a cleaner way to handle it that would be great.

plugins/repositories/repositories_inspector.rb
```diff
@@ -90,7 +92,42 @@ def inspect_yum_repositories
...
+    end
+
+    RepositoriesScope.new(
+      parse_apt_cache_poliy_output(content), repository_system: "apt"
+    )
+  end
+
+  def parse_apt_cache_poliy_output(content)
```
aduffeck: The name seems to be wrong, the output is the content of the `.list` files not from `apt-cache policy`.

plugins/repositories/repositories_renderer.rb
```diff
@@ -25,13 +25,21 @@ def content(description)
...
+          puts "URI: #{p.url}" if p.name
+          puts "Alias: #{p.alias}" if p.alias
+          puts "Distribution: #{p.distribution}" if p.distribution
+          puts "Components: #{p.components.join(", ")}" if p.components
+          puts "Enabled: #{p.enabled ? "Yes" : "No"}" if !p.enabled.nil?
           puts "Refresh: #{p.autorefresh ? "Yes" : "No"}" if p.autorefresh != nil
           puts "Priority: #{p.priority}" if p.priority != nil
-          puts "Package Manager: #{p.package_manager}"
+          puts "Type: #{p.type}"
+          puts "Package Manager: #{description.repositories.repository_system}"
```
aduffeck: Do you really think we should show this again for each repository? That's a lot of redundancy and it's not what we're doing for services, for example.

spec/unit/repositories_renderer_spec.rb
```diff
@@ -49,6 +51,7 @@
     Enabled: Yes
     Refresh: No
     Priority: 3
+    Type:
```
aduffeck: Where do the empty types come from?

thardeck: If a repository was added without specifying the type and `zypper refresh` wasn't run the type is `nil`. I can add something in brackets if this happens.

plugins/repositories/repositories_inspector.rb
thardeck: Yeah, list the content of this directory before but we don't have a method for this but I think Chris is working on something similar.

mauromorales:
let's keep in mind #1610 to include the yum mirror list changes in the schema

PR #1730, opened 2015-12-02
Supported systems change

mauromorales:


cornelius:
The patch generally looks good to me. The approach is good and the structure is fine. I haven't run the code, though.

Issue #1717, opened 2015-11-30
Count of Inspected Ubuntu Services Differs from System


svollath:
While the inspection of Ubuntu 14.04 LTS services finisged successful, there is a difference between services found and insstalled.
While for an Ubuntu Desktop machinery listed 100 Services, the running system lists 104 services with "initctl list".
Furthermore the system is running "apparmor", but this service isn't listed by machinery and/or "initctl list" at all - it's shown by e.g. "service --status-all".
So we will need to figure out how to reliably list all services on Ubuntu.

greygoo:
the issue with the differing numbers seems to come from different ways to show some of the services. 
the list from show-config:
  /sbin/initctl show-config -e | grep -e "^[[:alpha:]]" | sort
the list from list
  /sbin/initctl list | awk '{print $1}' | sort

If the output is diffed it shows that the additional services in the list view are simply services that are repeatedly listed like e.g. the network-interface service which is listed for each available network interface. 

For apparmor on the other hand we have to extend the inspector to also track services that are invisible in upstart tools, we might need another service section listing them as sysV services or we ignore them for now.

greygoo:
it is possible to see apparmor and other services with ```service --status-all```
However, that lists some more than just apparmor - but still less than initctl. Should we also evaluate the outpu of this command and then sync the list of services with the srvice list obtained from initctl?
I can do that but wonder if this is not rather a bug in ubuntu and if we should really work around that at this point.

greygoo:
I compared the output of ```services --status--all``` with the output of ```initctl list```

So far on the minimal ubuntu system used for pennyworth these services listed with ```services --status--all``` do not appear in the initctl output:

```
for i in `service --status-all 2>&1 | awk '{print $4}'`; do grep $i /tmp/initctl.list > /dev/null || echo $i; done
apparmor
dns-clean
grub-common
killprocs
ntp
ondemand
pppd-dns
rc.local
rsync
sendsigs
sudo
umountfs
umountnfs.sh
umountroot
urandom
```

I doubt we can say that this is a bug in ubuntu and just ignore it.

thardeck:
This should also be fixed

PR #1731, opened 2015-12-02
Ubuntu service inspector including sysv

greygoo:
Also inspect sysv services that are not handled by upstart in ubuntu
Fixes https://github.com/SUSE/machinery/issues/1717


plugins/services/services_inspector.rb
```diff
@@ -92,10 +92,22 @@ def inspect_sysvinit_services
...
   end
 
-  def inspect_upstart_services
-    output = @system.run_command("/sbin/initctl", "show-config", "-e", stdout: :capture)
+  def inspect_ubuntu_services
+    # Ubuntu is managing its services using upstart but also has some
+    # services still not handled by that. Therefor we need to scan upstart
+    # and sysV to determine all services and their bootup state
+    services = parse_ubuntu_upstart
+    parse_ubuntu_sysv.each do |service|
```
aduffeck: The `sysv` services are part of the same list as the `upstart` ones and the list has the `init_system` set to ` upstart`. Is that still correct or do we need two lists instead?

plugins/services/services_inspector.rb
```diff
@@ -106,8 +118,43 @@ def inspect_upstart_services
...
+    find_output = runlevels.each.map do |runlevel|
+      @system.run_command(
+        "/usr/bin/find",
+        "/etc/rc#{runlevel}.d",
+        "-name",
+        "S\*",
+        stdout: :capture).split
+    end
+
+    sysv_enabled = find_output.flatten.map { |line|
```
aduffeck: I'm not sure if that logic is always correct. You basically consider the union service set of all runlevels as enabled, but it could very well be that some services are only enabled in certain runlevels, no?

plugins/services/services_inspector.rb
greygoo: I was wondering if we should have two lists but as the system handling also the sysv services is upstart i would stick with one list. I'm just calling them sysv as they are technically sysv services. The only problem i see is that beside starting and stopping them upstart does not know anything about them, but that i think is just a design issue with this ubuntu release. Having two lists would imply that there are two init systems when in fact there is only one, upstart.

plugins/services/services_inspector.rb
greygoo: I thought this was our approach for sysv services in other distributions as well, if it gets started in one of the start runlevels, then we consider it as enabled.

PR #1734, opened 2015-12-02
Detect deleted files as deleted, although dpkg reports md5sum changes

tgoettlicher:


lib/rpm_database.rb
```diff
@@ -26,6 +26,10 @@ def package_for_file_path(file)
     [package_name, package_version]
   end
 
+  def determine_changes(line)
```
mauromorales: can this be done with an alias?

lib/dpkg_database.rb
```diff
@@ -45,6 +45,21 @@ def handle_verify_fail(_path)
     # This method is a no-op for DpkgDatabase.
   end
 
+  def determine_changes(line)
```
mauromorales: is it possible to just have a parse_changes_line method for dpkg that does this logic and then calls super?

spec/unit/dpkg_database_spec.rb
```diff
@@ -68,6 +68,16 @@
     end
   end
 
+  describe "#determine_changes" do
```
mauromorales: if the method can be renamed then this could just be a context scenario inside the #parse_changes_line

tgoettlicher:
Superseded by https://github.com/SUSE/machinery/pull/1735

PR #1735, opened 2015-12-02
Detect deleted files as deleted, although dpkg reports md5sum changes

tgoettlicher:


tgoettlicher:
squashed https://github.com/SUSE/machinery/pull/1736

PR #1743, opened 2015-12-03
Ubuntu inspection with repos html

thardeck:


tgoettlicher:
I tried to run it but it crashes with 
```
No such file or directory @ rb_sysopen - /data/machinery/html/partials/repositories_.html.haml
```.

thardeck:
The crash was related to an old migrated description which was done, when the repository migration wasn't merged yet.

tgoettlicher:
Thanks for the explanation. Then it looks good to me.

PR #1746, opened 2015-12-04
Implement ubuntu packages inspection

aduffeck:


plugins/packages/packages_inspector.rb
```diff
@@ -52,7 +68,49 @@ def inspect(_filter, _options = {})
     )
   end
 
-  def summary
-    "Found #{@description.packages.length} packages."
+  def inspect_dpkg
+    dpkg_data = @system.run_command(
+      "dpkg", "-l", stdout: :capture
```
thardeck: What happens if the database is locked here?

plugins/packages/packages_inspector.rb
```diff
@@ -52,7 +68,49 @@ def inspect(_filter, _options = {})
...
+    )
+
+    lines = dpkg_data.lines.reject { |line| !line.start_with?("ii") }
```
thardeck: I would add a trailing space after `ii` just to be sure

plugins/packages/packages_inspector.rb
```diff
@@ -52,7 +68,49 @@ def inspect(_filter, _options = {})
...
+
+    packages = lines.map do |line|
+      name, version, arch = line.split[1..3]
+      version_segments = version.split("-")
+      release = version_segments.pop if version_segments.length > 1
+      version = version_segments.join("-")
```
thardeck: Is there the possibility of more than one dash?

plugins/packages/packages_inspector.rb
```diff
@@ -52,7 +68,49 @@ def inspect(_filter, _options = {})
...
+        "apt-cache",
+        "show",
+        *packages_slice.map { |p| "#{p.name}=#{[p.version, p.release].compact.join("-")}" },
+        stdout: :capture
+      )
+
+      packages_slice.each do |package|
+        name = Regexp.escape(package.name.sub(/:[^:]+$/, ""))
+        apt_cache_output =~
+          /Package: #{name}.*?MD5sum: (\w+).*?Origin: (\w+)/m
```
thardeck: I would add a newline or \s after `Package: #{name}` because there are packages like `zlib1g` and `zlib1g-dev` where the first would also match the second.

spec/unit/packages_inspector_spec.rb
```diff
@@ -31,68 +31,214 @@
...
+  context "dpkg inspection" do
+    let(:dpkg_output) {
+      <<EOF
+Desired=Unknown/Install/Remove/Purge/Hold
+| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
+|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
+||/ Name                                                   Version                          Architecture                     Description
++++-======================================================-================================-================================-==================================================================================================================
+ii  accountsservice:amd64                                  0:0.6.35-foo-0ubuntu7.2          amd64                            query and manipulate user account information
+ii  adduser                                                3.113+nmu3ubuntu3                all                              add and remove users and groups
```
thardeck: I think it would make sense to add one removed package to make sure a refactoring doesn't mess up the parsing. But of course it is not that important with the integration tests, but they might also have no removed packages.

plugins/packages/packages_inspector.rb
```diff
@@ -23,8 +23,24 @@ def initialize(system, description)
   end
 
   def inspect(_filter, _options = {})
-    @system.check_requirement("rpm", "--version")
+    @system.check_requirement(["rpm", "dpkg"], "--version")
```
thardeck: Don't we also need to check for `apt-cache` ?

plugins/packages/packages_inspector.rb
thardeck: Nevermind, I saw the adduser package.

thardeck:
lgtm besides my comments

thardeck:
lgtm

PR #1752, opened 2015-12-14
[Review] Request from 'aduffeck' @ 'SUSE/machinery/ubuntu_service_inspector_sysv'

aduffeck:
  * e801508 Add an additional field to upstart services (legacy sysv flag)
  * 18353bf Extend service inspector to also inspect sysv services not handled by upstart
  * 74af258 Update expected data for integration test to include legacy systemV services
  * 3b38b05 Unit test for inspecting all services on ubuntu, sysv and upstart


thardeck:
lgtm

PR #1754, opened 2015-12-14
Ubuntu pattern inspection

greygoo:
Add functionality to inspect patterns on Ubuntu 14.04

### DIFF
--- a/spec/unit/patterns_inspector_spec.rb
+++ b/spec/unit/patterns_inspector_spec.rb
@@ -71,9 +71,6 @@ EOF
   }
 
   describe "#inspect" do
-#    before(:each) do
-#      allow(system).to receive(:has_command?).with("zypper").and_return(true)
-#    end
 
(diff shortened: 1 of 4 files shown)


spec/unit/patterns_inspector_spec.rb
```diff
@@ -55,11 +71,9 @@
   }
 
   describe "#inspect" do
-    before(:each) do
-      allow(system).to receive(:has_command?).with("zypper").and_return(true)
-    end
 
-    it "parses the patterns list into a Hash" do
+    it "parses the patterns list into a Hash on a zypper based OS" do
+      allow(system).to receive(:has_command?).with("zypper").and_return(true)
```
mauromorales: I see a bunch of repetition with this call. I think it'd look nicer if there was a context separation for "zypper" and "taskel" and they both would have a `before(:each)` with the according setup.

mauromorales:
lgtm just take care of the hound stuff

PR #1760, opened 2015-12-15
Document html overview

### LINKED ISSUE #1749: Document Landing Page
Document that a plain "machinery serve" (landing page) will now work.
A plain "machinery serve" will start a web server for the http server root (landing page):

'# machinery serve'
Trying to start a web server for serving all system descriptions on http://127.0.0.1:7585/

The web server can be closed with Ctrl+C.

This, and any other way to start the server's document root should be evaluated and documented in the man page - e.g. a plain "machinery show" currently doesn't work.

Now that I also tested the "list" command, it's also not documented that it can take the --html option.

cornelius:
Fixes #1749

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -688,7 +688,7 @@ class Cli
     c.switch :short, required: false, negatable: false,
       desc: "List only description names"
     c.switch :html, required: false, negatable: false,
-      desc: "Open all system description in HTML format in your web browser."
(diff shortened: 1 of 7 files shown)


aduffeck:
I couldn't test it because all my descriptions have the v6 format already but code-wise it lgtm besides the type.

cornelius:
OK, I fixed the typo, and will merge it now, so we can do QA on it and test it.

PR #1765, opened 2015-12-15
Ubuntu managed files integration tests5

tgoettlicher:


lib/system.rb
```diff
@@ -102,38 +102,23 @@ def run_script(*args)
...
 
-  # Runs the given script on the inspected machine asynchronously and calls the callback method
+  # Runs the given command on the inspected machine asynchronously and calls the callback method
   # periodically with new output when it occurs.
   #
   # Example:
   #
   #     count = 0
-  #     raw_list = run_script_with_progress("changed_managed_files.sh") do |chunk|
+  #     raw_list = run_command_with_progress("time_consuming_command") do |chunk|
```
thardeck: Why do you change the comments? The method underneath still has the old name.

thardeck:
lgtm besides my comment

lib/system.rb
tgoettlicher: `changed_managed_files.sh` implies that it's a script which is injected. In this example it's about commands in general.

thardeck:
lgtm

PR #1771, opened 2015-12-17
Upgrade nokogiri Version to 1.6.6.4 due to security vulnerabilities

greygoo:
https://github.com/sparklemotion/nokogiri/issues/1374


thardeck:
This should be obsolete since the security issues only affects the bundled xml version which we don't use.

PR #1786, opened 2015-12-21
Add additional yum repo attributes

mauromorales:

PR #1790, opened 2015-12-22
Add additional yum repo attributes

mauromorales:
supersedes #1786


thardeck:
lgtm by the other bug

PR #1795, opened 2016-01-05
Filter duplicate dpkg repositories

### LINKED ISSUE #1793: Duplicate repositories in ubuntu systems are not cleaned
If you inspect an ubuntu system which has the same repository listed twice (e.g. once in `/etc/apt/sources.list` and once in `/etc/apt/sources.list.d/foo.list`) machinery extracts and lists the repository twice.

I think we should rather make sure to only store each repositories once.

thardeck:
Fixes #1793


aduffeck:
lgtm

PR #1796, opened 2016-01-05
Adapt format and data model description to v6

aduffeck:


docs/Data-Model.md
```diff
@@ -31,30 +31,40 @@ For example, a JSON document describing software configuration may look like thi
...
+        "name": "kernel-default",
+        "version": "3.0.101",
+        "release": "0.35.1",
+        "arch": "x86_64",
+        "vendor": "SUSE LINUX Products GmbH, Nuernberg, Germany",
+        "checksum": "2a3d5b29179daa1e65e391d0a0c1442d"
+      },
+      ...
+    ]
+  },
```
cornelius: It would be good to explain the meaning of the special attributes `_attributes` and `_elements` and what consequences they have for the data model.

docs/System-Description-Format.md
```diff
@@ -78,6 +86,44 @@ A simplified example of the manifest can look like this:
...
+        "name": "acl",
+        "version": "2.2.52",
+        "release": "6.4"
+      }
+    ]
+  }
+```
+
+Array elements and Object values can be either primitive values or Arrays or Objects again. Plain
+JSON arrays are not allowed.
```
cornelius: What does "Plain JSON arrays are not allowed" mean? The content of the `_elements` attribute is a plain JSON array, isn't it? What happens if I use e.g. `"version": ["1.0", 2.0"]`?

This whole section is written from an implementation point of view. It probably would fit better into the data model documentation. For the format documentation users have to understand how to read the data, not how to construct new schemas.

docs/System-Description-Format.md
```diff
@@ -523,7 +576,10 @@ The users scope stores all user related information.
 
 JSON Example:
 ```json
-"users": [
+"users": {
+  "_attributes": {
+  },
```
cornelius: Shouldn't the `_attributes` object be removed when it's empty?

docs/System-Description-Format.md
```diff
@@ -605,21 +666,31 @@ JSON Example:
...
+      }
+    ]
+  }
 ```
 
+The attributes of the repositories array are:
+
+| item            | description                          | type                       |
+|-----------------|--------------------------------------|----------------------------|
+| repository_system | The repository manager (e.g. "zypp" or "apt") | string           |
```
cornelius: This is an enum, not a custom string, right? There is an exact list of allowed values ("zypp" and "apt").

docs/System-Description-Format.md
```diff
@@ -605,21 +666,31 @@ JSON Example:
...
+
 Here all information for repositories configured in the inspected system get stored. The repo alias, it repo name, its url, type, if it is enabled and if autorefresh is on, if the gpg key is checked for it, the repository priority and the package manager (zypp or yum).
```
cornelius: The package manager is not stored here anymore, isn't it?

docs/System-Description-Format.md
```diff
@@ -654,9 +729,18 @@ JSON Example:
...
+}
 ```
 
+The attributes of the repositories array are:
+
+| item            | description                          | type                       |
+|-----------------|--------------------------------------|----------------------------|
+| repository_system | The repository manager (e.g. "zypp" or "apt") | string           |
+
+The attributes of the elements are:
```
cornelius: Aren't the attributes different for different repository_systems?

docs/System-Description-Format.md
```diff
@@ -654,9 +729,18 @@ JSON Example:
...
-]
+  ]
+}
 ```
 
+The attributes of the repositories array are:
+
+| item            | description                          | type                       |
+|-----------------|--------------------------------------|----------------------------|
+| repository_system | The repository manager (e.g. "zypp" or "apt") | string           |
```
cornelius: This should be "package_system".

docs/System-Description-Format.md
cornelius: Isn't the schema dependent on the package system?

docs/System-Description-Format.md
```diff
@@ -720,18 +812,30 @@ JSON Example:
         "name": "SuSEfirewall2_init.service",
         "state": "disabled"
       }
-]
+    ]
+  }
 ```
 
-| item        | description          | type   | 
+The attributes of the repositories array are:
```
cornelius: This is the services array, not the repositories array.

docs/System-Description-Format.md
```diff
@@ -720,18 +812,30 @@ JSON Example:
...
+
+| item        | description          | type   |
 |-------------|----------------------|--------|
-| init_system | the init system used | string | 
-| services    | list of services     | array  |
+| init_system | the init system used | string |
```
cornelius: The possible values should be listed here.

cornelius:
The services section is missing an explanation of upstart in the description at the beginning.

docs/System-Description-Format.md
aduffeck: "Plain JSON arrays are not allowed"  means that all arrays containing machinery data should be using the special format, with the elements being contained in the `_elements` array. That itself is just an array of course. 

`"version": ["1.0", 2.0"]` would probably work as well, but I remember that we decided to only use one array format during one of our discussions.

docs/System-Description-Format.md
aduffeck: Yes, that would make sense. I'll change it in the documents and the code.

docs/System-Description-Format.md
aduffeck: The only difference is that `checksum` may be empty in the `dpkg` case.

PR #1797, opened 2016-01-07
Add changelog for support changes

mauromorales:


NEWS
```diff
@@ -1,5 +1,8 @@
 # Machinery Release Notes
 
+* Remove concept of "supported systems" in favor of letting Machinery run
+  everywhere. Weather a system can be built, exported, etc will be validated
```
thardeck: Shouldn't it be "Wether"?

thardeck:
And I would add the related bnc number.

PR #1801, opened 2016-01-07
Revert nokogiri requirement

thardeck:


tgoettlicher:
LGTM, as you made sure that it's not necessary to upgrade from 1.6.1 to 1.6.3 in our case.

PR #1805, opened 2016-01-11
Add updated grub packages to Ubuntu integration test

thardeck:
I haven't activated update repos but the grub packages were still
updated. I am not sure yet  why but Veewee might pull some newer
packages in.


tgoettlicher:
LGTM

PR #1809, opened 2016-01-11
Remove /boot from default filter

thardeck:


cornelius:
Is this the right solution? These files are quite big and they are generated, aren't they? So maybe it would be better to make sure we capture the information from which the files are generated.

In a case where a user really wants to extract these files as well, it should be possible to do with the `--exclude` option, shouldn't it?

thardeck:
The override is at the moment not yet possible and I would prefer that if the user is not interested in it to add a filter to the command line.

mauromorales:
lgtm

PR #1817, opened 2016-01-13
Fixes github issue 1777

greygoo:


plugins/services/services_inspector.rb
```diff
@@ -35,10 +35,17 @@ def inspect(_filter, _options = {})
         init_system: "systemd"
       )
     elsif @system.has_command?("initctl")
-      result = ServicesScope.new(
-        inspect_ubuntu_services,
-        init_system: "upstart"
-      )
+      if @system.has_command?("chkconfig")
```
aduffeck: The redhat case is covered by the `else` branch already, so wouldn't it be sufficient to replace line 37 with `elsif @system.has_command?("initctl") && !@system.has_command?("chkconfig")`?

greygoo: No, as then for RHEL6 still sysvinit would be the recorded init system, which is wrong.

aduffeck: Ah, right.

aduffeck:
lgtm once the unit test passes (which it should when you rebase against the branch).

PR #1827, opened 2016-01-14
Remove extra dash when displaying packages without release

### LINKED ISSUE #1766: Superfluous dash in `show -s packages` for certain Ubuntu packages
Some packages in Ubuntu have only a version but no release. `machinery show -s packages UBUNTU` displays a superfluous dash because the release is empty in this case.

When fixing this issue, the general way of representation of package information should be reconsidered because users who are familiar to deb might have different expectations than rpm users.

mauromorales:
fixes #1766


thardeck:
lgtm

PR #1831, opened 2016-01-14
Make package build filter generic

### LINKED ISSUE #1830: Building SLES12 SP1 images fails
Building SLES12 SP1 images fails with
`Package 'sles-release-DVD' not found.`

mauromorales:
Since we are allowing any system to run the build task so we don't have
to keep track of them then we need a generic way to filter files from
OSs that we don't keep track of.

In general there doesn't seem to be a need of filtering a package in one
OS and not in another and therefore we just made it generic.

Fixes #1830


lib/kiwi_config.rb
```diff
@@ -227,10 +227,9 @@ def generate_config
   end
 
   def apply_packages(xml)
-    build_filter = YAML.load_file(
+    filter = YAML.load_file(
       File.join(Machinery::ROOT, "filters", "filter-packages-for-build.yaml")
     )
```
thardeck: I guess we still need the || []

PR #1832, opened 2016-01-14
Make package build filter generic

### LINKED ISSUE #1830: Building SLES12 SP1 images fails

mauromorales:
Since we are allowing any system to run the build task so we don't have
to keep track of them then we need a generic way to filter files from
OSs that we don't keep track of.

In general there doesn't seem to be a need of filtering a package in one
OS and not in another and therefore we just made it generic.

Fixes #1830
Supersedes #1831


thardeck:
lgtm

PR #1837, opened 2016-01-14
Change warning message regarding inaccurate config-files inspection

### LINKED ISSUE #1811: Inspecting dpkg based warning needs rewording
When inspecting changed-managed file and/or config-files on an Ubuntu system, machinery says:
`The inspection of config-files and changed-managed files is not accurate on Ubuntu systems: Only content changes to a subset of files can be detected. Relevant changes might not be caught.`
The message shouldn't refer to Ubuntu but to dpkg.

thardeck:
Fixes #1811


lib/dpkg_database.rb
```diff
@@ -18,8 +18,7 @@
 class DpkgDatabase < ManagedFilesDatabase
   def managed_files_list(&block)
     message = "The inspection of config-files and changed-managed files is not "\
-     "accurate on Ubuntu systems: Only content changes to a subset of files can "\
-     "be detected. Relevant changes might not be caught."
+     "complete on dpkg systems: https://github.com/SUSE/machinery/wiki/Ubuntu-Inspection"
```
mauromorales: To me this doesn't make it clear if the lists of files are incomplete or if data in those lists are incomplete or both.

Also maybe we could just quickly mention why this is and then have a more detailed explanation in the link.

mauromorales:
lgtm

PR #1856, opened 2016-01-19
[Review] Request from 'aduffeck' @ 'SUSE/machinery/review_160119_fix_unmanaged_files'

aduffeck:
  * a0d1e35 Filter /var/lib/dpkg by default
  * 19313bb Fix inspection of unmanaged files on ubuntu


filters/default_filters.json
```diff
@@ -19,6 +19,7 @@
     "/unmanaged_files/name=/etc/init.d/rc4.d",
     "/unmanaged_files/name=/etc/init.d/rc5.d",
     "/unmanaged_files/name=/etc/init.d/rc6.d",
-    "/unmanaged_files/name=/etc/init.d/rcS.d"
+    "/unmanaged_files/name=/etc/init.d/rcS.d",
+    "/unmanaged_files/name=/var/lib/dpkg"
```
thardeck: I would prefer to group it with rpm but it is not that important

thardeck:
The man page needs to be adapted with the missing commands

thardeck:
lgtm

thardeck:
The integration tests data needs to be adapted

PR #1859, opened 2016-01-19
Upstart rhel6

### LINKED ISSUE #1853: legacy_sysv flag will only be set during a description upgrade
When I inspect a RHEL 6 system with the `ubuntu_inspection` branch, the `legacy_sysv` flag will not be set for services (only when I do an upgrade from schema 5 to schema 6).

The init system `upstart` can be properly recognized.

System Description Diff: https://gist.github.com/mschnitzer/2bcd4c0340af072e0293

(Compared with a a fresh inspection and an upgraded description from 5 to 6)

mauromorales:
fixes #1853
supersedes #1858

### DIFF
--- a/plugins/services/services_inspector.rb
+++ b/plugins/services/services_inspector.rb
@@ -38,7 +38,6 @@ class ServicesInspector < Inspector
         [inspect_sysvinit_services, "sysvinit"]
       end
 
-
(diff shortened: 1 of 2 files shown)


thardeck:
lgtm

PR #1861, opened 2016-01-19
Update man page for ubuntu

thardeck:


aduffeck:
The merge commit shouldn't be in here imo, but other than that it looks good to me except for the duplicate entry.

PR #1863, opened 2016-01-20
Option ssh port and key

mschnitzer:
Supersedes #1862


mschnitzer:
Will create a new PR when I'm done.

PR #1864, opened 2016-01-20

mschnitzer:


lib/remote_system.rb
```diff
@@ -18,14 +18,23 @@
...
   attr_accessor :remote_user
+  attr_accessor :ssh_port
+  attr_accessor :ssh_key_options
 
   def type
     "remote"
   end
 
-  def initialize(host, remote_user = "root")
+  def initialize(host, remote_user = "root", ssh_port = 22, private_key_path = "")
```
mauromorales: would it make sense to group all these into one options argument?

mschnitzer: I also was thinking about that. I agree, I should do that.

lib/remote_system.rb
```diff
@@ -18,14 +18,23 @@
 class RemoteSystem < System
   attr_accessor :host
   attr_accessor :remote_user
+  attr_accessor :ssh_port
+  attr_accessor :ssh_key_options
```
mauromorales: all these can be written in one attr_accessor call separated by comma

lib/cli.rb
```diff
@@ -584,10 +584,19 @@ def self.parse_inspect_command_options(host, options)
...
+    c.flag [:key, :k], type: String, required: false,
+      default_value: @config.default_ssh_key,
+      desc: "The private SSH key location what can be used to authenticate with the remote system.",
+      arg_name: "SSH-KEY"
 
     c.action do |_global_options, options, args|
       host = shift_arg(args, "HOSTNAME")
-      system = System.for(host, options["remote-user"])
+      system = System.for(host, remote_user: options["remote-user"], ssh_port: options["port"],
+        private_key_path: options["key"])
```
cornelius: To be consistent with ssh I would rather name the parameter `ssh_identity_file`.

lib/remote_system.rb
```diff
@@ -16,16 +16,28 @@
...
+
     @host = host
-    @remote_user = remote_user
+    @remote_user = @options[:remote_user]
+    @ssh_port = @options[:ssh_port]
+    @private_key_path = @options[:private_key_path]
+    @ssh_key_options = []
+
+    if @private_key_path.length > 0
+      @ssh_key_options = ["-i", @private_key_path]
```
cornelius: `@ssh_key_options` doesn't really make sense here as a name, as it's a generic array storing arbitrary options. I would rather store a list of identity files.

man/machinery-inspect.1.md
```diff
@@ -48,6 +48,15 @@ trigger errors.
...
     To change the default-user use `machinery config remote-user=USER`
 
+  * `-p SSH-PORT`, `--port SSH-PORT` (optional):
+    Specifies the SSH port of the remote SSH server.
+
+  * `-k SSH-KEY`, `--key SSH-KEY` (optional):
+    Specifies the SSH private key what should be used to authenticate with the
+    remote SSH server. It's important that private keys with a passphrase
+    are already loaded by the `ssh-agent` so that it's possible to connect
+    not interactively.
```
cornelius: Does this make sense? If ssh-agent already has the key then you don't need to provide it with an option, do you?

mschnitzer:
Superseded by #1883

PR #1865, opened 2016-01-20
Remove unnecessary scopes from alert message test

mauromorales:
For this tests it's only important that an alert message is raised so no
scopes are required but just the fact that the version of the schema is
an old one.


aduffeck:
lgtm

PR #1866, opened 2016-01-20
Fix migration issue

mschnitzer:
We decided that we do not want the `_attributes` element during the migration in `users`, `groups`, and `patterns`.


mschnitzer:
Sorry, wrong branch.

PR #1867, opened 2016-01-20
Remove `_attributes` during migration in users, groups, and patterns

mschnitzer:
We decided that we do not want the `_attributes` element in `users`, `groups`, and `patterns`.


mschnitzer:
Superseded by #1868

PR #1881, opened 2016-01-21
Suggest a name for docker images with a solidus

### LINKED ISSUE #1630: `inspect-container` output for image name containing a solidus (`/`) needs improvement
The wording `consist a slash` needs improvement. 
The output should also print a suggestion which command line to run.

```
$ machinery inspect-container foo/bar -s os
Inspecting foo/bar for os...
Error: System description name 'foo/bar' is invalid. By default Machinery uses the image name as description name if the parameter `--name` is not provided.
If the image name consist a slash the `--name=NAME` parameter is mandatory. Valid characters are 'a-zA-Z0-9_:.-'.
```

thardeck:
Fixes #1630


lib/cli.rb
```diff
@@ -150,7 +150,8 @@ def self.check_container_name!(image, name)
         "Error: System description name '#{name}' is invalid. By default Machinery" \
           " uses the image name as description name if the parameter `--name` is not" \
           " provided.\nIf the image name contains a slash the `--name=NAME` parameter" \
-          " is mandatory. Valid characters are 'a-zA-Z0-9_:.-'."
+          " is mandatory. Valid characters are 'a-zA-Z0-9_:.-'.\n\nFor example run:" \
```
aduffeck: A linebreak after `run:` would be helful so that you can directly copy the whole line.

aduffeck:
lgtm besides the newline and the hound issues.

aduffeck:
lgtm

PR #1883, opened 2016-01-22

mschnitzer:
Supersedes #1864 

Difference to the old PR:
* remote_system.rb and its spec file got refactored. I added getters for some member variables of RemoteSystem because the full access doesn't make any sense.
* I implemented a function to better build the command line arguments for `ssh`, `scp`, and `rsync`
* Renamed the options to `--ssh-port` and `--ssh-identity-file`


mschnitzer:
Superseded by #1884

PR #1884, opened 2016-01-22

mschnitzer:
Supersedes #1883

Difference to the old PR:
* remote_system.rb and its spec file got refactored. I added getters for some member variables of RemoteSystem because the full access doesn't make any sense.
* I implemented a function to better build the command line arguments for `ssh`, `scp`, and `rsync`
* Renamed the options to `--ssh-port` and `--ssh-identity-file`


spec/unit/remote_system_spec.rb
```diff
@@ -18,7 +18,15 @@
 require_relative "spec_helper"
 
 describe RemoteSystem do
-  let(:remote_system) { RemoteSystem.new("remotehost") }
+  let(:rsys) { RemoteSystem.new("remotehost") }
```
mauromorales: all these initializations of `RemoteSystem` would be better handled by using the subject and then putting the different arguments in variables that describe what they are ment to be. This way you could always use the subject and depending on the describe/context block define those argument variables

lib/remote_system.rb
```diff
@@ -175,4 +184,42 @@ def remove_file(file)
...
+
+    if @ssh_identity_file
+      command_options.push("-i")
+      command_options.push(@ssh_identity_file)
+    end
+
+    command_options
+  end
+
+  def build_basic_ssh_command
```
mauromorales: these two commands are so much alike that we could handle the differences by passing `:ssh` or `:scp` as an argument

mauromorales: and because the logic of ssh and scp is inside the get_cmd_options maybe it actually belongs there

mschnitzer: Yes, I agree here.

mauromorales: weather the methods get merged or not they sound to me like they belong more to the private side of the API

spec/unit/remote_system_spec.rb
mschnitzer: Would you mind, making an example? :-) I have no idea what you mean by that.

mauromorales: subject { RemoteSystem.new("remotehost", options) }

so you can always use subject and you only define options in the context that you need them like

let(:options_empty) { {} }
let(:options_ssh_port} { {ssh_port: 9999 } }

aduffeck: Wouldn't all the overriden option hashes have to be called `options` to make that work?

mauromorales: yup sorry fixing it now

mschnitzer:
Superseded by #1898

PR #1891, opened 2016-01-22
Use proper check for btrfs subvolume check

thardeck:
/sbin/btrfs is needed because btrfs can't be called by the remote user
without it since /sbin is usually not in the PATH variable for regular
users.


aduffeck:
lgtm

PR #1894, opened 2016-01-22
Prevent autoyast export for non suse

### LINKED ISSUE #1892: it is possible to create an autoyast import from a ubuntu repository
After having inspected a ubuntu system it is possible to export the system description as autoyast files. This fails with an error message about ubuntu being not supported for the kiwi export, so we might also want to block the autoyast export for such systems.

thardeck:
Fixes #1892


greygoo:
LGTM

PR #1929, opened 2016-02-01
Use dpkg instead of apt for repository recognition

thardeck:
because at least on older Debian versions there is no apt
available, only apt-get. Since apt-get is deprecated we
check for dpkg instead, which needs to be installed on
a Debian based system.


aduffeck:
lgtm

PR #1933, opened 2016-02-01
Move broken UTF-8 file from /usr/locale to /usr

mschnitzer:
One of our tests checks that files like `/usr/local/broken\255file` with UTF-8 issues are properly handled. However the folder `/usr/local` is not part of any package in Ubuntu 14.04 and therefore files inside it are never taking into account. Moving it inside a managed directory fixes the test.

Supersedes #1932


mauromorales:
lgtm

PR #1934, opened 2016-02-01
Fix 1928

thardeck:
Fixes 1928

### DIFF
--- a/lib/tee_io.rb
+++ b/lib/tee_io.rb
@@ -1,0 +1,33 @@
+# Copyright (c) 2013-2016 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of version 3 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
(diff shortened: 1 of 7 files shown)


lib/tee_io.rb
```diff
@@ -0,0 +1,33 @@
...
+# you may find current contact information at www.suse.com
+
+# TeeIO
+#
+# Class to allow the storage and forwarding of input at the same time
+# For example if stderr output should be handled by our code but at the
+# same time be put out directly
+class TeeIO < StringIO
+  def initialize(io_object)
+    super()
```
mschnitzer: Are the braces needed here? When yes, where is the difference to line 30?

mschnitzer:
lgtm

lib/tee_io.rb
thardeck: Super does automatically inherits the parameters of the method it is called in, which we don't want here.

PR #1944, opened 2016-02-02
Add filter to TeeIO

thardeck:


lib/tee_io.rb
```diff
@@ -15,19 +15,25 @@
...
+# To prevent for example double error messages data can be filtered
+# before being passed on to the io_object
+# Only the data passed along is filtered, not the data stored in this
+# IO object.
 class TeeIO < StringIO
-  def initialize(io_object)
+  def initialize(io_object, io_filter = [])
     super()
     @io_object = io_object
+    @io_filter = io_filter.is_a?(Array) ? io_filter : [io_filter]
```
aduffeck: This can be simplified to `@io_filter = Array(io_filter)`.

spec/unit/tee_io_spec.rb
```diff
@@ -35,5 +35,31 @@
...
+        expect(io_object.string).to include("anything")
+      end
+
+      it "filters arrays of strings from being forwarded to io_object" do
+        subject = TeeIO.new(io_object, ["filter1", filter_string])
+        subject.write(filter_string)
+        expect(subject.string).to include(filter_string)
+        expect(io_object.string).not_to include(filter_string)
+      end
+
```
aduffeck: There's a superfluous blank line here.

aduffeck:
Eventually it might be helpful to be able to specify regexes for filtering, but since we don't need it right now we can also add it later.

aduffeck:
lgtm

PR #1954, opened 2016-02-08
Add latest release notes to homepage

mauromorales:


mschnitzer:
Improvements:
* Add a loading circle (can be generated here http://www.ajaxload.info/) that the user sees that something is happening right now
* When GitHub is not only right now, we should inform the user that GitHub is down and it's not possible to fetch the latest release notes
* Add the `target="_blank"` attribute to all GitHub issue links

index.html
```diff
@@ -382,6 +403,26 @@ <h1 class="headline-philosophy" align="center">Philosophy</h1>
...
+          $.get(
+              "https://raw.githubusercontent.com/SUSE/machinery/master/NEWS",
+              {},
+              function(data) {
+                var latestEntry = data.split("##")[1].split("* ");
+                var heading = latestEntry[0]
+                $("#latest-release").text(heading.match(/\d+\.\d+.\d+/))
+                for (var i = 1; i < latestEntry.length; i++) {
+                  var issue = latestEntry[i].match(/gh#SUSE\/machinery#(\d+)/)
+                  if(issue) {
```
mschnitzer: I'd add a space after `if` because there is also one after `for`. That makes it more consistent.

aduffeck:
Since the move to mkdocs we keep the source for the homepage in `manual/site` in the master branch, so those changes should go there.

The change itself lgtm, although Manuel's suggestions would be nice to have (but could also be added later)

mauromorales:
Right, will move the changes to `manual/site`

Issue #1616, opened 2015-11-09
inspect experienced an unexpected error during rpm -qf stage, if file missing


tigerfoot:
-> Found 96 config files...Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
Execution of "ssh root@c-3po -o LogLevel\=ERROR LC_ALL\=en_US.utf8 rpm -qf /usr/share/vnc/classes/index.vnc\\\ \\\(replaced\\\)" failed with status 1: error: file /usr/share/vnc/classes/index.vnc (replaced): No such file or directory.
```
Error output:
error: file /usr/share/vnc/classes/index.vnc (replaced): No such file or directory

Backtrace:
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/bundle/ruby/2.0.0/gems/cheetah-0.4.0/lib/cheetah.rb:555:in `check_errors'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/bundle/ruby/2.0.0/gems/cheetah-0.4.0/lib/cheetah.rb:364:in `run'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/logged_cheetah.rb:23:in `run'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/remote_system.rb:84:in `run_command'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/rpm_database.rb:46:in `block in changed_files'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/rpm_database.rb:40:in `each_line'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/rpm_database.rb:40:in `each'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/rpm_database.rb:40:in `map'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/rpm_database.rb:40:in `changed_files'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/plugins/config_files/config_files_inspector.rb:51:in `inspect'
/usr/lib64/ruby/gems/2.0.0/gems/machinery-tool-1.16.0/lib/inspect_task.rb:87:in `block in build_description'
[...]

PR #1962, opened 2016-02-09
Fix treatment of the rpm `(replaced)` flag

mschnitzer:


mschnitzer:
Since this was a pair programming task, we do not need review here.

mauromorales:
fixes #1616

PR #1963, opened 2016-02-09
Extract metadata

mauromorales:


machinery-helper/machinery_helper.go
```diff
@@ -236,11 +263,86 @@ func findUnmanagedFiles(dir string, rpmFiles map[string]string, rpmDirs map[stri
...
 }
 
+func amendMode(entry *UnmanagedFile, perm os.FileMode) {
+	if entry.Type == "link" {
+		return
+	}
+
+	result := int64(perm.Perm())
+
+	if perm&os.ModeSticky > 0 {
```
mschnitzer: I would add a space here before and after the `&`.

machinery-helper/machinery_helper.go
```diff
@@ -236,11 +263,86 @@ func findUnmanagedFiles(dir string, rpmFiles map[string]string, rpmDirs map[stri
...
+		result |= 01000
+	}
+	if perm&os.ModeSetuid > 0 {
```
mschnitzer: I would add a space here before and after the `&`.

machinery-helper/machinery_helper.go
```diff
@@ -236,11 +263,86 @@ func findUnmanagedFiles(dir string, rpmFiles map[string]string, rpmDirs map[stri
...
+		result |= 04000
+	}
+	if perm&os.ModeSetgid > 0 {
```
mschnitzer: I would add a space here before and after the `&`.

machinery-helper/machinery_helper.go
```diff
@@ -236,11 +263,86 @@ func findUnmanagedFiles(dir string, rpmFiles map[string]string, rpmDirs map[stri
...
+	size = int64(0)
+	fileCount = len(files)
+	for _, f := range files {
+		if f.IsDir() {
+			if _, ok := IgnoreList[path + f.Name()]; !ok {
+				subSize, subCount := dirInfo(path + f.Name() + "/")
+				size += subSize
+				fileCount += subCount
+			}
+		} else if f.Mode()&os.ModeSymlink != os.ModeSymlink {
```
mschnitzer: I would add a space here before and after the `&`.

schema/migrations/migrate6to7.rb
```diff
@@ -0,0 +1,38 @@
+# Copyright (c) 2013-2014 SUSE LLC
```
mschnitzer: Adjust the year to 2016 please

spec/unit/migrations/migrate6to7_spec.rb
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) 2013-2014 SUSE LLC
```
mschnitzer: Please adjust the year to 2016.

machinery-helper/machinery_helper.go
mauromorales: I prefer the way it is right now and go fmt doesn't have a preference for this and go sources like the os package use it without the spaces when doing a validation like in

```
func (m FileMode) IsRegular() bool {
	return m&ModeType == 0
}
```
https://golang.org/src/os/types.go?s=3244:3278#L83

They also seem to use it the other way around when just doing the bit operation like in https://golang.org/src/os/types.go?s=3355:3388#L88

We could start defining this kind of thing on a

machinery-helper/file_utils_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"testing"
+)
+
+func TestGetFileOwnerGroup(t *testing.T) {
+	statFile = func(path string) string {
+		return "foo:bar"
+	}
+	path := "/etc/passwd"
+
+	owner, group := getFileOwnerGroup(path)
```
mschnitzer: I would do the variable assignment in one "block" like:
```go
func TestGetFileOwnerGroup(t *testing.T) {
	statFile = func(path string) string {
		return "foo:bar"
	}

	path := "/etc/passwd"
	owner, group := getFileOwnerGroup(path)

    if owner != "foo" {
```

machinery-helper/machinery_helper.go
mschnitzer: OK, I'm fine with that.

PR #1976, opened 2016-02-11
Don't use Nokogiri

cornelius:
Move code to use REXML and Builder instead of Nokogiri, because the Nokogiri gem causes trouble when libxml versions are not matching. There is no heavy XML processing going on, so the Ruby libraries are good enough.


mschnitzer:
I haven't had the time to checkout and test the code but codewise it looks good. lgtm from my side.

aduffeck:
lgtm

aduffeck:
lgtm

Issue #1921, opened 2016-01-29
Exporting the description of a docker image to kiwi fail with several issues


greygoo:
Description:

I tried to export the description created from inspecting a docker container to kiwi. This triggered a backtrace and error message as the docker image has no init system:

------------------snip-----------------------
```
chuller@lagrange:/data/chuller/machinery(master)> ./bin/machinery export-kiwi --kiwi-dir /tmp/opensuse_container opensuse
Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
/data/chuller/machinery/lib/kiwi_config.rb:336:in `apply_services': Unsupported init system: "none". (RuntimeError)
        from /data/chuller/machinery/lib/kiwi_config.rb:223:in `block (2 levels) in generate_config'
        from /usr/lib64/ruby/gems/2.1.0/gems/nokogiri-1.6.7.1/lib/nokogiri/xml/builder.rb:391:in `call'
        from /usr/lib64/ruby/gems/2.1.0/gems/nokogiri-1.6.7.1/lib/nokogiri/xml/builder.rb:391:in `insert'
        from /usr/lib64/ruby/gems/2.1.0/gems/nokogiri-1.6.7.1/lib/nokogiri/xml/builder.rb:375:in `method_missing'
        from /data/chuller/machinery/lib/kiwi_config.rb:196:in `block in generate_config'
        from /usr/lib64/ruby/gems/2.1.0/gems/nokogiri-1.6.7.1/lib/nokogiri/xml/builder.rb:293:in `initialize'
        from /data/chuller/machinery/lib/kiwi_config.rb:195:in `new'
        from /data/chuller/machinery/lib/kiwi_config.rb:195:in `generate_config'
        from /data/chuller/machinery/lib/kiwi_config.rb:35:in `initialize'
        from /data/chuller/machinery/lib/cli.rb:472:in `new'
        from /data/chuller/machinery/lib/cli.rb:472:in `block (2 levels) in <class:Cli>'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/command_support.rb:126:in `call'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/command_support.rb:126:in `execute'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/app_support.rb:296:in `block in call_command'
[...]

PR #1977, opened 2016-02-12
Detect correct init system for containers in `export-kiwi`

mschnitzer:
This fixes a part of issue #1921


mschnitzer:
Superseded by #1978

PR #1981, opened 2016-02-15
Make openSUSE Leap the default build target

cornelius:


thardeck:
You have to change the Test Environment documentation but besides that lgtm

cornelius:
I already changed the documentation. Will merge now.

Issue #1983, opened 2016-02-16
Performance down for processing descriptions (e.g. unmanaged files)


svollath:
There's a drop in performance for processing descriptions (e.g. for upgrade-format, show) from current SLES/Leap version (1.16.4) to current master (1.17.0).
I tested with two full SLES-12 inspections, while the difference is about unmanaged-files:

"parse-speed-normal": 1506 unmanaged files and trees.
 "parse-speed-manyfiles": 41507 unmanaged files and trees. - ~18sec vs. ~7min40sec
((upgrade-format for "parse-speed-manyfiles" also took a very long already))

See the timings:
```
machinery-1.16.4-6.1.x86_64
skylab:~ # time machinery show parse-speed-normal
real	0m2.063s
user	0m1.336s
sys	0m0.164s

skylab:~ # time machinery show parse-speed-manyfiles
real	0m18.318s
user	0m17.552s
sys	0m0.400s


machinery-1.17.0.20160216T120248ZopenSUSELeap421git71be1ce-1.x86_64
skylab:~ # machinery upgrade-format parse-speed-normal
Reading 'parse-speed-normal' ... Successfully upgraded from version 5 to 7.
skylab:~ # machinery upgrade-format parse-speed-manyfiles
Reading 'parse-speed-manyfiles' ... Successfully upgraded from version 5 to 7.

skylab:~ # time machinery show parse-speed-normal
real	0m2.659s
user	0m1.904s
sys	0m0.100s

skylab:~ # time machinery show parse-speed-manyfiles
real	7m42.310s
user	7m34.632s
sys	0m0.708s
```

cornelius:
git bisect shows that the problem is this commit 
https://github.com/SUSE/machinery/commit/7ce74e42c2f3e9e3f3a70d295fd58a1fd622ab69

This is the schema change to allow meta data without extracting files.

PR #1984, opened 2016-02-18
json-schema monkey patch for a performance bug with certain schemas

aduffeck:


cornelius:
Looks good to me. This fixes the issue and performance is back to the original values.

mauromorales:
fixes #1983

PR #1985, opened 2016-02-19
Activate helper for all architectures and remove old code path

thardeck:


machinery.spec.erb
```diff
@@ -35,7 +35,9 @@ Release:        0
 %endif
 
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-%ifarch x86_64
+%if 0%{?suse_version} == 1315 && 0%{?is_opensuse} != 1
```
mauromorales: It would be nice to have a comment saying what these numbers mean

spec/unit/helper_builder_spec.rb
```diff
@@ -284,15 +285,55 @@
...
+      allow(subject).to receive(:run_uname_p).and_return("x86_64")
+    end
+
+    context "for Go 1.4.2" do
+      before(:each) do
+        allow(subject).to receive(:run_go_version).and_return("go version go1.4.2 linux/amd64")
+      end
+
+      it "returns only the local arch if it is x86_64" do
+        expect(subject).to receive(:run_uname_p).and_return("x86_64")
```
mauromorales: is this guy necessary? seems to be the result of the stub on the "before" clause

thardeck: You are right but it is easier to understand I guess but I don't mind removing it.

tools/helper_builder.rb
```diff
@@ -85,15 +88,39 @@ def go_available?
...
+      elsif version <= 1.6
+        archs = ["i686", "x86_64", "ppc64le"]
+        # the two following lines are temporary until Go 1.7 with s390x support is released
+        archs.push("s390x") if File.exist?("/usr/share/go/src/cmd/asm/internal/arch/s390x.go")
+        archs
+      else
+        ["i686", "x86_64", "ppc64le", "s390x"]
+      end
+    end
+    @archs
```
mauromorales: I (we) think this should be extracted into different methods inside a class

tools/helper_builder.rb
```diff
@@ -85,15 +88,39 @@ def go_available?
...
+  end
+
   def arch_supported?
     arch = run_uname_p
-    if !["x86_64"].include?(arch)
+    if buildable_archs.include?(arch)
```
mauromorales: In general all these specific checks sound like they would suit better to be handled by a different class so we can reduce the amount of nested helps and having all the logic in one place

PR #1997, opened 2016-03-01
Remove RPM spec sanity check

thardeck:
because if the Go compilation fails the rpm build would raise anyway.


cornelius:
I haven't run the code but the changes looks good to me.

Issue #2001, opened 2016-03-03
machinery analyze - exception for package "grub2-2.02~beta2"


svollath:
Acceptance testing for current SLES release (1.18.0):
Inspecting a Leap system, and perform machinery analyze will fail for package "grub2-2.02~beta2".

```
# machinery analyze skylab.suse.de
Setting up repository access...
Generating diffs...
Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
Execution of "zypper --non-interactive --no-gpg-checks --root /tmp/machinery_zypper20160303-4356-1bxbod0 -x download grub2-2.02\~beta2" failed with status 4 (no error output).

Standard output:
<?xml version='1.0'?>
<stream>
<message type="error">Error building the cache:
[repo-debug-non-oss|http://download.opensuse.org/debug/distribution/leap/42.1/repo/non-oss/] Valid metadata not found at specified URL
</message>
<message type="warning">The metadata cache needs to be built for the &apos;repo-debug-non-oss&apos; repository. You can run &apos;zypper refresh&apos; as root to do this.</message>
<message type="info">Disabling repository &apos;repo-debug-non-oss&apos;.</message>
<message type="error">Error building the cache:
[repo-debug-update-non-oss|http://download.opensuse.org/debug/update/leap/42.1/non-oss/] Valid metadata not found at specified URL
</message>
<message type="warning">The metadata cache needs to be built for the &apos;repo-debug-update-non-oss&apos; repository. You can run &apos;zypper refresh&apos; as root to do this.</message>
<message type="info">Disabling repository &apos;repo-debug-update-non-oss&apos;.</message>
<message type="error">Some of the repositories have not been refreshed because of an error.</message>
<message type="info">Loading repository data...</message>
<message type="info">Reading installed packages...</message>
<message type="info">Retrieving package grub2-2.02~beta2-76.1.x86_64 (1/1),   2.4 MiB ( 19.8 MiB unpacked)</message>
<progress id="" name="(1/1) grub2-2.02~beta2-76.1.x86_64 (download.opensuse.org-oss_1)" value="0"/>
<progress id="" name="(1/1) grub2-2.02~beta2-76.1.x86_64 (download.opensuse.org-oss_1)" value="1"/>
<progress id="" name="(1/1) grub2-2.02~beta2-76.1.x86_64 (download.opensuse.org-oss_1)" value="12"/>
[...]

PR #2003, opened 2016-03-04
Handle failed zypper downloads in analyze

thardeck:


svollath:
JFYI: IIRC for me this zypper error came up due to a "pilot error" only. It was when I executed a zypper command ran by machinery, that referred to a temporary zypper root, which was not in place anymore - in this case any package isn't found.
For this fix to make sense, the question would be, if it will ever be the case for analyze, that these zypper commands are run, while the temporary zypper root isn't present.
((Also see updated Issue#2001))

mauromorales:
@svollath let's keep #2001 open and handle that one fully on it's own ... in the mean time we can merge this since it helps with the error when analyzing an inspection from our testing VMs

@thardeck lgtm

PR #2007, opened 2016-03-07
Optimize assert_scope to be copyable by the user

### LINKED ISSUE #2002: Analyze command recommends incorrect scope name for config files
```
 $ bin/machinery analyze grub2
The system description misses the following scopes : os, repositories, config_files.
 $ bin/machinery inspect e11.suse.de -n grub2 -s os,repositories,config_files                                                                                                                 
The following scope is not valid: 'config_files'. Scope names must start with a letter and contain only lowercase letters and digits separated by dashes ('-').
```

thardeck:
Fixes #2002

### DIFF
--- a/spec/unit/system_description_spec.rb
+++ b/spec/unit/system_description_spec.rb
@@ -197,30 +197,21 @@ describe SystemDescription do
   end
 
   describe "#assert_scopes" do
-    let(:working_description) {
-      create_test_description(
-        name: @name, scopes: ["packages", "repositories", "config_files"]
+    it "raises if one of the required scopes is missing" do
+      description = create_test_description(scopes: ["packages", "repositories"])
+      expect {
+        description.assert_scopes("repositories", "packages", "config_files")
+      }.to raise_error(
+        Machinery::Errors::SystemDescriptionError,
+        /The system description misses the following scope: config-files/
       )
-    }
+    end
 
-    it "checks the system description for completeness" do
-      [
-        ["repositories"],
-        ["packages"],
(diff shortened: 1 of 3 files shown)


spec/unit/system_description_spec.rb
```diff
@@ -197,22 +197,29 @@ class SystemDescriptionFooConfig < Machinery::Object; end
...
           missing.each do |element|
             description.send("#{element}=", nil)
           end
-          description.assert_scopes("repositories", "packages")
+          description.assert_scopes("repositories", "packages", "config_files")
         }.to raise_error(
            Machinery::Errors::SystemDescriptionError,
-           /: #{missing.join(", ")}\./)
+           /: #{missing.map { |scope|
+             Machinery::Ui.internal_scope_list_to_string(scope)
```
mauromorales: I'd prefer to have the specif text checked here

spec/unit/system_description_spec.rb
```diff
@@ -197,22 +197,29 @@ class SystemDescriptionFooConfig < Machinery::Object; end
...
+    }
+
     it "checks the system description for completeness" do
-      full_description = create_test_description(name: @name, json: @description)
       [
         ["repositories"],
         ["packages"],
-        ["repositories", "packages"]
+        ["repositories", "packages", "config_files"]
       ].each do |missing|
```
mauromorales: Do we really need to test each case? we could just test the case when the required scopes are there and when they are not

mauromorales:
lgtm

PR #2009, opened 2016-03-10
shows getting_started hint if 'machinery' is called without arguments

### LINKED ISSUE #1286: Show 'getting_started' hint if `machinery` is called without arguments
If you call `machiney help` the following hint is shown:
```
Machinery can show hints which guide through a typical workflow.
These hints can be switched off by '/usr/bin/machinery config hints=off'.

Hint: You can get started by inspecting a system. Run:
/usr/bin/machinery inspect HOSTNAME
```

The same hint should be shown if `machinery` is called without arguments.

fschueller:
Fixes #1286, in which the getting started hint is not printed when calling ```machinery``` without arguments

### DIFF
--- a/lib/cli.rb
+++ b/lib/cli.rb
@@ -41,7 +41,7 @@ class Cli
   end
 
   post do |global_options,command,options,args|
-    if command.is_a?(GLI::Commands::Help) && !global_options[:version] && ARGV == ["help"]
+    if command.is_a?(GLI::Commands::Help) && !global_options[:version] || ARGV == ["help"]
 
       Machinery::Ui.puts "\nFor more detailed information, open the documentation by typing " \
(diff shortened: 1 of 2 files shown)


lib/cli.rb
```diff
@@ -41,7 +41,7 @@ class Cli
   end
 
   post do |global_options,command,options,args|
-    if command.is_a?(GLI::Commands::Help) && !global_options[:version] && ARGV == ["help"]
+    if command.is_a?(GLI::Commands::Help) && !global_options[:version] || ARGV == ["help"]
```
aduffeck: I think it would be good to add some parentheses here to make the logic more obvious.

And how about adding an integration test for that hint?

aduffeck:
If you remove the string interpolation (see the hound comment), this pr lgtm.

Issue #1927, opened 2016-01-29
Fix scope naming "config_files" vs. "changed configuration files" in output and documentation


svollath:
The scope "config_files" is referred to in different naming throughout machinery:
- different output for:
```machinery show <descr>```: title is "Changed configuration files"
 but for ```machinery show --html <descr>```: title is "Config Files"
- The man page even includes the scope "config_files", and talks about "changed configuration files" at the same time
- Other places?

We should only refer to the scope using one and the same wording.

PR #2010, opened 2016-03-10
Issue #1927: Change output of 'show <name>' to keep consistency

promulo:
_This is my first PR :), so I apologize if any project guideline is not being followed_

Fixes https://github.com/SUSE/machinery/issues/1927

I have changed the output of `machinery show <name>` to be consistent with what is showed in the HTML output from `machinery show --html <name>`. The issue also mentions the man page, but I was not able to find any naming inconsistency inside `manual/docs`.

In the future though, I would suggest to rename the scope name, HTML and command line outputs from _config files_ to _changed configuration files_, like it is already done with managed files. The meaning is more clear, IMHO.

In case I missed anything, I would be glad to know. Thanks. :)


thardeck:
Hi @promulo, thx for your pull request.
I am currently working on renaming our config-file inspector to changed-config-file inspector and one of the tasks is also cleaning up the config vs configuration usage so it will conflict with your pr.

promulo:
Hi @thardeck, thanks for the quick response. 

I'll close the PR then. :)

mauromorales:
Thanks for you PR @promulo :) much appreciated

Just to keep you on the loop, for this fix we decided that in places where the config-files are represented as words (sentences, headers, etc) we are going to call it "Changed Configuration Files" and in places where reference is made to the argument used in the command line we are going to use "changed-config-files". This wasn't clear from the issue :/

If there is any other issue you feel like picking up just ping me in there and I can give you a bit more context if necessary ;)

Cheers,

Issue #897, opened 2015-05-20
Global `--exclude` and `--exclude-scope` option can be confused


cornelius:
It's possible to abbreviate `--exclude-scope` as `--exclude`, so that the meaning of the option is only specified by the position on the command line. This can result in very confusing errors, when the user uses the options at the wrong place with the expectation that it is the other one.

fschueller:
Fixed in https://github.com/SUSE/machinery/pull/2011

PR #2011, opened 2016-03-10
renaming of --exclude-scope option to avoid confusion with global --exclude

fschueller:
Referencing #897, in which the coexistence of  ```--exclude-scope``` and the global option ```--exclude``` might result in undesired behaviour.

Thank you for reviewing (:

### DIFF
--- a/manual/docs/machinery-compare.1.md
+++ b/manual/docs/machinery-compare.1.md
@@ -3,7 +3,7 @@
 
 ## SYNOPSIS
 
-`machinery compare` [-s SCOPE | --scope=SCOPE] [-e ignore-scope | --ignore-scope=ignore-scope] [--no-pager] [--show-all] [--html] NAME1 NAME2
+`machinery compare` [-s SCOPE | --scope=SCOPE] [-e IGNORE-SCOPE | --ignore-scope=IGNORE-SCOPE] [--no-pager] [--show-all] [--html] NAME1 NAME2
 
 `machinery` help compare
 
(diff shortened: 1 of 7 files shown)


manual/docs/machinery-compare.1.md
```diff
@@ -3,7 +3,7 @@
 
 ## SYNOPSIS
 
-`machinery compare` [-s SCOPE | --scope=SCOPE] [-e EXCLUDE-SCOPE | --exclude-scope=EXCLUDE-SCOPE] [--no-pager] [--show-all] [--html] NAME1 NAME2
+`machinery compare` [-s SCOPE | --scope=SCOPE] [-e IGNORE-SCOPE | --ignore-scope=IGNORE-SCOPE] [--no-pager] [--show-all] [--html] NAME1 NAME2
```
thardeck: Shouldn't the short form also be renamed to `-i` to be not confusing?

thardeck:
besides that lgtm

manual/docs/machinery-compare.1.md
fschueller: Later on I thought so too, but ```-i``` is already reserved for the ```--ssh-identity-file``` option. Maybe we need to discuss this seperately.

thardeck: Maybe that existing `-i` can be replaced by `-s` since it is most likely less commonly used and "s" is also the first later of the full name.

Issue #1939, opened 2016-02-02
machinery serve url example causes backtrace in browser


greygoo:
When running 
```
machinery serve
```

you get the following output?

```
chuller@lagrange:/data/chuller/machinery(master)> ./bin/machinery serve
Trying to start a web server for serving a view on all system descriptions.

The overview of all descriptions is accessible at:

    http://127.0.0.1:7585/

A specific descriptions with the name NAME is accessible at:

    http://127.0.0.1:7585/NAME

The web server can be closed with Ctrl+C.
```

when pasting ```http://127.0.0.1:7585/NAME``` in the browser - a unhandled exception happens and you see a backtrace. We should catch that.

PR #2012, opened 2016-03-11
Issue #1939: Machinery serve url example causes backtrace in browser

promulo:
Fixes https://github.com/SUSE/machinery/issues/1939

The call is `GET` to `/:id`. When `:id` does not exist, it will redirect to the landing page.


promulo:
Hi @mauromorales, thanks for the clarification regarding my other pull request.

Feel free to review this one, then. :)

Thanks a lot and have a nice weekend,
--
Paulo

mauromorales:
@promulo looks good but as a user I think it would be a bit confusing to just be redirected without knowing what is going on. Could we maybe add a message to the alert box on the home page giving the reason why he is on the home page instead of the system view?

promulo:
@mauromorales Hi Mauro, thanks for reviewing. Your suggestion makes a lot of sense, so I have put the error message inside the alert in the landing page. The way I did that was including the error inside a session (we lose context when redirecting, check the commit message :). However, since I'm not very experienced with Sinatra, I'm not sure this is the best way.

Thanks!

mauromorales:
@promulo nice this works pretty well for the use case we're talking about unfortunatelly after going back to the list of descriptions I'm getting the following issue displayed in my alert as well.

```
Couldn't find a system description with the name 'favicon.ico'
```

promulo:
@mauromorales Thanks again for reviewing. 

Sometimes browsers do those `GET`s to /favicon.ico (I could only reproduce using `localhost` instead of `127.0.0.1` on my machine), and since machinery does not serve one, this causes a route conflict with `/:id`, that's why you see the message. I have included an empty favicon in order to avoid this.

Also, one thing that I missed was to clear the session after saving the error into `@errors`. That will prevent the alert to be showed more than once.

Feel free to review it again and let me know if it works for you.

Cheers,
Paulo

lib/server.rb
```diff
@@ -328,7 +330,13 @@ def diff_to_object(diff)
   end
 
   get "/:id" do
-    @description = SystemDescription.load(params[:id], settings.system_description_store)
+    begin
+      @description = SystemDescription.load(params[:id], settings.system_description_store)
+    rescue Machinery::Errors::SystemDescriptionNotFound => e
+      session[:error] = e.to_s
+      status 404
```
cornelius: There is no point in setting the status, when the response is a redirect, or is there?

spec/unit/server_spec.rb
```diff
@@ -70,6 +70,14 @@ def app
       end
     end
 
+    describe "GET /:id with non-existent id" do
+      it "redirects to landing page if description is not found" do
+        get "/does_not_exist"
+
+        expect(last_response).to be_redirect
+      end
+    end
```
cornelius: It would be good to also have a feature test, which checks, that the error message is actually displayed.

lib/server.rb
promulo: Thanks for reviewing. 

I wondered the same thing. The final response will be a 302 anyway. I have removed it. :)

spec/unit/server_spec.rb
promulo: I agree. I've also added a check for the error message.

cornelius:
Thanks for the patch. I merged it to master now.

PR #2014, opened 2016-03-14
Add changed to config files

### LINKED ISSUE #1927: Fix scope naming "config_files" vs. "changed configuration files" in output and documentation

thardeck:
Fixes #1927


plugins/changed_config_files/changed_config_files_inspector.rb
```diff
@@ -37,25 +37,27 @@ def initialize(system, description)
...
-    do_extract = options[:extract_changed_config_files]
+    do_extract = options[:extract_changed_changed_config_files]
     system.check_retrieve_files_dependencies if do_extract
 
     count = 0
     files = @system.managed_files_database.changed_files do |chunk|
       count += chunk.lines.count { |l| !l.chomp.end_with?(":") && l.split(" ")[1] == "c" }
-      Machinery::Ui.progress(" -> Found #{Machinery.pluralize(count, "%d changed config file")}...")
+      Machinery::Ui.progress(
+        " -> Found #{Machinery.pluralize(count, "%d changed configuration file")}..."
```
mauromorales: Just as a comment I think this would read better if it said "changed configuration file(s)" but we can do that in a different PR otherwise this would be the only scope that reads this way

schema/migrations/migrate7to8.rb
```diff
@@ -0,0 +1,49 @@
...
+    if File.directory?(File.join(@path, "analyze", "config_file_diffs"))
+      FileUtils.mv(
+        File.join(@path, "analyze", "config_file_diffs"),
+        File.join(@path, "analyze", "changed_config_files_diffs")
+      )
+    end
+
+    if @hash.key?("config_files")
+      @hash["changed_config_files"] = @hash["config_files"]
+      @hash.delete("config_files")
```
mauromorales: This could be shortened to

```
@hash["changed_config_files"] = @hash.delete("config_files")
```

schema/migrations/migrate7to8.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    meta = @hash["meta"]
+    if meta.key?("config_files")
+      meta["changed_config_files"] = meta["config_files"]
+      meta.delete("config_files")
```
mauromorales: same here

schema/migrations/migrate7to8.rb
thardeck: I know but I have asked andred because I did not find the short form that readable and he also did like the current one more.

plugins/changed_config_files/changed_config_files_inspector.rb
mauromorales: ignore previous comment

mauromorales:
lgtm

PR #2015, opened 2016-03-14
mention remote-user option in hint

### LINKED ISSUE #1813: Mention `--remote-user` parameter for inspection
Hint doesn't mention `--remote-user` parameter when needed for an inspection.

fschueller:
Fixes #1813, in which the hint doesn't mention the remote-user option.


thardeck:
We start git commit titles with an uppercase letter

lib/hint.rb
```diff
@@ -44,7 +44,10 @@ def which_machinery
     end
 
     def get_started(_options)
-      "You can get started by inspecting a system. Run:\n#{program_name} inspect HOSTNAME"
+      "You can get started by inspecting a system. Run:\n'#{program_name} inspect HOSTNAME'\n" \
+      "Specify user for connecting via SSH:\n" \
+      "'#{program_name} inspect --remote-user USER HOSTNAME'\n" \
+      "(Note: specified user needs sudo access or root on remote machine)"
```
thardeck: What do you mean by `or root on remote machine` ?

Sudo is mandatory for each user except of root but if a user wants to inspect as root they do not need the option so I would not mention it here.

thardeck: Maybe you can even shorten it by mention `Specify user with sudo rights for conenction via SSH:\n`

fschueller: Thanks for the feedback, i basically interpreted what the ```-h``` option told me but I will implement your feedback.
Also thanks for mentioning your git commit conventions, very helpful!

lib/hint.rb
```diff
@@ -44,7 +44,9 @@ def which_machinery
     end
 
     def get_started(_options)
-      "You can get started by inspecting a system. Run:\n#{program_name} inspect HOSTNAME"
+      "You can get started by inspecting a system. Run:\n'#{program_name} inspect HOSTNAME'\n" \
+      "Specify user with sudo rights for connecting via SSH:\n" \
```
mauromorales: I'd mention somehow that this is only necessary if you don't want to run it with root so that it's clear that you should never do `inspect --remote-user root ...`

thardeck:
lgtm

PR #2018, opened 2016-03-14
Fix some style issues2

thardeck:


thardeck:
Just a test pr to check prophet

PR #2027, opened 2016-03-16
Remove repositories scope from check in autoyast-export to avoid missing scope error

### LINKED ISSUE #2024: exporting a description with no repo scope to autoyast does not work.
When I inspect a system which has no repos set up, I can do an export:
-------------------------------snip---------------------------------
chuller@lagrange:/data/big/chuller/machinery(master)> ./bin/machinery --debug export-autoyast no_repos_2_sle11_removed --autoyast-dir /tmp

Unmanaged files following these patterns are not exported:
var/lib/rpm/*
etc/passwd*
etc/shadow*
etc/group*
etc/fstab
lib/mkinitrd
etc/udev/rules.d/70-persistent-net.rules
var/log/*
etc/default/grub*
etc/zypp/repos.d/*

Note: The permssions of the AutoYaST directory are restricted to be only accessible by the current user. Further instructions are provided by the README.md in the exported directory.
Exported to '/tmp/no_repos_2_sle11_removed-autoyast'.
-------------------------------snip---------------------------------

However, when skippig the repo inspection altogether, an error message prevents me from exporting to autoyast:

---------------------------------snip-----------------------------------
chuller@lagrange:/data/big/chuller/machinery(master)> ./bin/machinery --debug export-autoyast no_repos_2_sle11 --autoyast-dir /tmp
The system description misses the following scope : repositories.
---------------------------------snip-----------------------------------

Is this the intended behaviour?

fschueller:
Fixes #2024, in which exporting a description with no 'repositories scope' to autoyast does not work, although it's supposed to work nevertheless.

### DIFF
--- a/lib/autoyast.rb
+++ b/lib/autoyast.rb
@@ -24,7 +24,6 @@ class Autoyast < Exporter
     @system_description = description
     @system_description.assert_scopes(
       "os",
-      "repositories",
       "packages"
     )
     check_exported_os


mauromorales:
lgtm

PR #2028, opened 2016-03-17
Filter broken utf-8 characters in renderer

thardeck:
The HTML output did already escaped broken utf-8 characters but there was a regression found during testing which still used the old path before migration.


tgoettlicher:
Thanks for the quick fix. The changelog doesn't mention the fix for the crash when processing invalid utf-8 characters in config files. That's okay, as it's a minor fix and the issue wasn't reported by a customer through an official channel. Looks good to me.

PR #2031, opened 2016-03-18
Issue #1747: Missing Link from System Descriptions back to Landing Page

### LINKED ISSUE #1747: Missing Link from System Descriptions back to Landing Page
When testing the new machinery landing page (machinery serve) I can easily navigate to existing system descriptions by clicking on their link - however, if I'm watching a system description, there's no link to the landing page present.
The user needs to use the browsers navigation buttons, or to modify the URL - a link would be more intuitive.
Also, if you "show --html" or "serve" a specific description, the landing page can be accessed just by modifying the URL to the html server root - without a link to the server root/landing page nobody will be aware of that.
Off course such links would need to support current server IP and Port.

promulo:
Fixes #1747


cornelius:
I think the styling is not ideal. There is missing some margin at the top. It's also not clear what this link means, when you start from the individual description (by using `machinery show --html`).

We are working on a redesign of the overall navigation. This will change the structure of the pages, so that the issue with the missing link will go away.

But thanks for the patch anyway.

PR #2035, opened 2016-03-21
Issue # #2034: serve command fails without hostname specified in /etc/hosts

fschueller:
Fixes #2034, in which running ```machinery serve``` with a hostname not defined in ```/etc/hosts``` results in a exception.

PR #2036, opened 2016-03-21
Provide host for public serve command url

### LINKED ISSUE #2034: `machinery serve` fails
When running `machinery serve` I get an error:

```
Machinery experienced an unexpected error. Please file a bug report at: https://github.com/SUSE/machinery/issues/new
/home/cs/suse/alfred/machinery/lib/serve_html_task.rb:21:in `gethostbyname': getaddrinfo: Name or service not known (SocketError)
        from /home/cs/suse/alfred/machinery/lib/serve_html_task.rb:21:in `serve'
        from /home/cs/suse/alfred/machinery/lib/cli.rb:951:in `block (2 levels) in <class:Cli>'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/command_support.rb:126:in `call'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/command_support.rb:126:in `execute'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/app_support.rb:296:in `block in call_command'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/app_support.rb:309:in `call'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/app_support.rb:309:in `call_command'
        from /usr/lib64/ruby/gems/2.1.0/gems/gli-2.13.4/lib/gli/app_support.rb:83:in `run'
        from /home/cs/suse/alfred/machinery/bin/machinery:41:in `<main>'
```

The host name of the system is a local host name, which is not known to DNS or defined in `/etc/hosts`.

fschueller:
Fixes  #2034, in which running machinery serve with a hostname not defined in /etc/hosts results in an exception.
Supersedes #2035


spec/integration/support/serve_examples.rb
```diff
@@ -27,7 +27,7 @@
     def test_basic_html(port)
       wait_time = 0
       loop do
-        curl_command = @machinery.run_command("curl http://localhost:#{port}/opensuse131")
+        curl_command = @machinery.run_command("curl http://127.0.0.1:#{port}/opensuse131")
```
aduffeck: Does `localhost` not work anymore ?

fschueller: It does work, it just looked more consistent to me that way. Should I change it back?

lib/serve_html_task.rb
```diff
@@ -16,13 +16,17 @@
...
 class ServeHtmlTask
-  def serve(system_description_store, opts)
-    hostname = Socket.gethostbyname(Socket.gethostname).first
-    url = "http://#{hostname}:#{opts[:port]}/"
-    if opts[:public]
-      remote_ip = Socket.ip_address_list.find { |ip| ip.ipv4? && !ip.ipv4_loopback? }.ip_address
-      remote_url = "http://#{remote_ip}:#{opts[:port]}/"
+  def assemble_url(opts)
+    host = if opts[:public]
+      Socket.gethostname
```
aduffeck: Wouldn't it still make sense to try to get the "real" hostname and only fall back to the alias if it fails? I think it's not very unrealistic that the locally configured hostname is not the one under which the machine is reachable from the outside.

spec/integration/support/serve_examples.rb
aduffeck: No, it doesn't really matter I guess. It just wasn't clear to me if/why this change was necessary.

lib/serve_html_task.rb
fschueller: I guess your suggestion is a little closer to what I did before, but it was decided against doing so in the end. If it's still unclear maybe we have to discuss separately how we want to proceed.

PR #2039, opened 2016-03-22
Fix kiwi bootstrap

thardeck:
Use a static list to find packages relevant for bootstrap instead of just adding all of them.


fschueller:
LGTM

PR #2040, opened 2016-03-24
Add changelog entry for kiwi export bootstrap selection

thardeck:


NEWS
```diff
@@ -1,5 +1,6 @@
 # Machinery Release Notes
 
+* Only use necessary packages for bootstrap in Kiwi export
```
fschueller: I'd phrase it "required packages", but basically it's ok the way it is and LGTM

mauromorales:
Lgtm

PR #2042, opened 2016-03-29
Issue #1820: wording of patterns on Debian-like systems

### LINKED ISSUE #1820: Description of patterns scope for non RPM based systems
Our patterns scope title is not be very clear for users who have dpkg based tasks

- [x] HTML view
- [x] System inspection
- [x] Error message when there is the `taskel` program isn't installed (see: https://github.com/SUSE/machinery/pull/1781/files#diff-5ca3896c8c50f9d57c744d8a0cccc1dcR35)

fschueller:
Fixes #1820, where the explanation for the patterns scope for descriptions on Debian-like systems was not clear enough.


lib/inspect_task.rb
```diff
@@ -79,6 +79,9 @@ def build_description(store, name, system, scopes, filter, options)
     scopes.each do |scope|
       inspector = Inspector.for(scope).new(system, description)
       Machinery::Ui.puts "Inspecting #{Machinery::Ui.internal_scope_list_to_string(inspector.scope)}..."
+      if scope == "patterns"
```
thardeck: This message would be shown every time even when you inspect an rpm based system.

lib/inspect_task.rb
```diff
@@ -79,6 +79,9 @@ def build_description(store, name, system, scopes, filter, options)
...
+        Machinery::Ui.puts "Note: Patterns on Debian-like systems are treated as tasks."
```
thardeck: Don't we treat Debian tasks as patterns?

plugins/patterns/patterns_inspector.rb
```diff
@@ -32,20 +32,25 @@ def inspect(_filter, _options = {})
         inspect_with_tasksel
       else
         @patterns_supported = false
-        @status = "For a patterns inspection please install the package tasksel " \
+        @status = "For a tasks inspection please install the package tasksel " \
```
thardeck: I would use `patterns/tasks` here unless we also allow the scope tasks

spec/integration/machinery_matcher_spec.rb
```diff
@@ -78,6 +78,22 @@
 
     expect(actual_output).to_not match_machinery_show_scope(expected_output)
   end
+
+  it "matches '/' in show output" do
```
thardeck: I would rephrase it as it `allows slashes in the scope name`

plugins/patterns/patterns_inspector.rb
thardeck: Reading all these / I guess it might make sense to also allow `tasks` as a synonym for `patterns`. But I would ask others if they agree.

PR #2043, opened 2016-03-29
Issue #1940: machinery serve with an old format version triggers exception

### LINKED ISSUE #1940: machinery serve with an old format version triggers exception
Running   
```machinery serve```

and then connecting to

``` http://127.0.0.1:7585/$description_with_old_format_version ```

 runs into a backtrace shown in the browser after the error message:

```
Machinery::Errors::SystemDescriptionIncompatible at /ubuntu
ubuntu: format version 5, needs to be upgraded. Try './bin/machinery upgrade-format ubuntu' to upgrade it to the current version.

BACKTRACE

(expand)

JUMP TO: GET POST COOKIES ENV

    /data/chuller/machinery/lib/system_description.rb in rescue in from_hash
                  raise Machinery::Errors::SystemDescriptionIncompatible.new(name, json_format_version)
    /data/chuller/machinery/lib/system_description.rb in from_hash
              begin
    /data/chuller/machinery/lib/system_description.rb in load 
...
```

### PR DESCRIPTION
**Pull request by fschueller**
Fixes #1940, in which an outdated format of a description throws an exception when pasted directly into the browser after calling ``machinery serve``.
The suggested fix was catching the error and displaying an HTML view with instructions on how to upgrade.

### CODE REVIEW

**Comment by mauromorales** [comment]
Can we add a test for this?

**Comment by fschueller** [comment]
Sure!

File: html/upgrade.html.haml [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+      #nav-bar
+        .row
+          .col-xs-1
+          .col-xs-10
+            %h1
+              System Description incompatible!
+        .row
+          .col-xs-1
+          .col-xs-4
+            Outdated format version of system description
```
thardeck: You are handling two cases, one in which case the description has an old format version and one which the format version is newer than the one machinery supports.

In the second case this message is not correct.

spec/unit/server_spec.rb
```diff
@@ -61,13 +61,28 @@ def app
...
+
       it "returns the page" do
         get "/#{description_a.name}"
-
         expect(last_response).to be_ok
         expect(last_response.body).
           to include("#{description_a.name} - Machinery System Description")
       end
+
+      it "returns update instruction page" do
```
thardeck: I would add a context here why an update instruction page is shown and also test the case for newer system descriptions.

spec/unit/server_spec.rb
```diff
@@ -60,28 +60,49 @@ def app
...
-        create_test_description(
-          name: "description_d",
-          store_on_disk: true,
-          scopes: ["os"],
-          format_version: 7
-        )
-      }
+    context "description with incompatible format version" do
+      describe "GET /:id" do
+        let(:description_d) {
```
thardeck: I would use speaking names like "older_format_version" and for the other "unknown_format_version" or "newer_format_version"

thardeck:
lgtm

PR #2051, opened 2016-04-04
Issue #1820: wording of patterns on Debian-like systems

### LINKED ISSUE #1820: Description of patterns scope for non RPM based systems

fschueller:
Fixes #1820, where the explanation for the patterns scope for descriptions on Debian-like systems was not clear enough.

Supersedes #2050


NEWS
```diff
@@ -4,6 +4,8 @@
   intuitive
 * Following the previous change the analyze operation was also renamed to
   changed-config-files-diffs
+* Includes explanation for patterns-tasks relationship for Debian based
```
aduffeck: Maybe `Add explanation...` to better match the general style in this file?

lib/inspect_task.rb
```diff
@@ -79,6 +79,10 @@ def build_description(store, name, system, scopes, filter, options)
     scopes.each do |scope|
       inspector = Inspector.for(scope).new(system, description)
       Machinery::Ui.puts "Inspecting #{Machinery::Ui.internal_scope_list_to_string(inspector.scope)}..."
+      if scope == "patterns" && description.packages &&
+          description.packages.package_system == "dpkg"
```
aduffeck: I think this should go to the patterns inspector instead. That also has the benefit that you already know whether it's a `dpkg` based system or not, so you don't have to rely on the packages scope to be around.

html/partials/patterns.html.haml
```diff
@@ -6,6 +6,11 @@
         :scope => "patterns",
         :title => "Patterns",
         :count => "#{patterns.length} #{Machinery.pluralize(patterns.length, "pattern")}"
+      - if @dpkg_note
```
aduffeck: I don't like having a global variable around for that. Could we pass along the description and check for the package_system here directly instead? That way you don't have to touch the `server.rb` anymore, which you shouldn't really for that kind of change.

plugins/patterns/patterns_inspector.rb
```diff
@@ -32,20 +32,25 @@ def inspect(_filter, _options = {})
...
 
   def summary
     if @patterns_supported
-      "Found #{Machinery.pluralize(@description.patterns.count, "%d pattern")}."
+      concept = if @system.has_command?("dpkg")
+        "task"
+      else
+        "pattern"
+      end
+      "Found #{Machinery.pluralize(@description.patterns.count, "%d #{concept}")}."
```
aduffeck: Should we really talk about `tasks` in the `patterns` inspector output?

aduffeck:
lgtm

PR #2053, opened 2016-04-04
Issue #2046: HTML View doesn't List Name of a Broken System Description

### LINKED ISSUE #2046: HTML View doesn't List Name of a Broken System Description
While ```machinery list``` will warn you about a broken system description, e.g.:
```
test-props: This description is broken.
```
The warning differs for html view (for ```machinery serve``` and ```machinery list --html```):
```
Machinery::Errors::SystemDescriptionError
```
1. The description name is missing in html view (so, error for which description?)
2. The message in html view should be the same: "```<Name>```: This description is broken."

FYI: I broke this description by adding ```,"blafasel": true``` to the unmanaged_files section:
```
  "unmanaged_files": {
    "_attributes": {
      "extracted": false,
      "has_metadata": false,
      "blafasel": true
    },
```

promulo:
Fixes #2046 

I've included the same message as the one showed when `machinery list` is executed in order to keep consistency.


spec/unit/server_spec.rb
```diff
@@ -191,6 +191,55 @@ def app
         end
       end
     end
+
+    context "broken description" do
+      before do
+        store_raw_description("foo", <<-EOT
+        {
```
fschueller: Good morning @promulo, thank you for taking on this issue! 

Instead of defining a long description in the spec file itself, take a look at the ```system_description_factory.rb``` file. You can also create a broken example scope there to test this specific case.

promulo: Hi @fschueller, thanks for reviewing! 

I didn't see how could I use the factory's `create_test_description` without having it [validating](https://github.com/SUSE/machinery/blob/master/spec/support/system_description_factory.rb#L87) the JSON. That's why I used `store_raw_description` instead. You can see that I've included the `foo` attribute which would make another error to be raised.

Or maybe I just missed something. ;)

P.S.: Maybe some sections of that broken JSON can be removed.

promulo: hey @fschueller I've made the test description much smaller :)

but if you know how can I create a broken description using the factory, just let me know!

thanks!

spec/unit/server_spec.rb
```diff
@@ -194,40 +194,25 @@ def app
...
-                "unmanaged_files": {
-                    "modified": "2016-04-05T12:07:33+02:00",
-                    "hostname": "example.com"
-                }
+        store_raw_description(
+          "foo", <<-EOT
+            {
+              "unmanaged_files": {
+                  "_attributes": {
+                      "extracted": true,
```
fschueller: You were right about the validation, this was my mistake. In this case it's fine to use ```store_raw_description``` here. 
Just strip down the example JSON to the utmost necessary attributes only, which means you can also get rid of the ```extracted``` attribute, the ```_elements``` array and everything inside the ```meta``` scope. Other than that it ltgm!

promulo: Thanks again @fschueller! I will create another PR to not pollute the commit history.

promulo:
Closed in favor of #2055

PR #2056, opened 2016-04-06
Arm support

thardeck:
I have tested this against armv7l and aarch64, I had not armv6l but it should be the same as for armv7l.


fschueller:
lgtm

Issue #1904, opened 2016-01-25
Fields of text output are not explained


cornelius:
When the data of a system description has content which is not speaking for itself it's hard to tell what it means when looking at the text output. There is no documentation or help anywhere, which explains what the fields are.

E.g. packages on a dpkg system are listed like this:

```
 * cron-3.0pl1-124ubuntu2.amd64 (N/A)
 * curl-7.35.0-1ubuntu2.5.amd64 (N/A)
 * dash-0.5.7-4ubuntu1.amd64 (N/A)
``` 

It's not clear what the "N/A" is supposed to mean.

Another example from a users scope:

```
* bitnami (,,,, uid: 1000, gid: 1000, shell: /bin/bash)
* bitnamiftp (N/A, uid: 1000, gid: 1000, shell: /bin/bitnami_ftp_false)
 ```

It's not clear what the four commas or the "N/A" are supposed to mean.

PR #2057, opened 2016-04-07
 Issue #1904: Fields of text output are not explained

fschueller:
Fixes https://github.com/SUSE/machinery/issues/1904, in which 'N/A' fields are not further explained.


thardeck:
I would only show the message when there are N/A elements, otherwise it might be annoying. Or making it optional somehow.

greygoo:
LGTM

lib/renderer.rb
```diff
@@ -266,4 +266,9 @@ def indent
     yield
     @indent -= 2
   end
+
+  def na_note(s)
```
thardeck: I would use a speaking variable name.

spec/support/system_description_factory.rb
```diff
@@ -631,6 +649,25 @@ def build_description(name, store, options)
...
   EOF
 
+  EXAMPLE_SCOPES["users_without_comment"] = <<-EOF.chomp
+    "users": {
+      "_attributes": {},
+      "_elements": [
+        {
+          "name": "bin",
+          "password": "x",
+          "uid": null,
```
thardeck: Is this case possible? I do not think that a user without a uid is possible.

lib/renderer.rb
fschueller: I followed the example of other methods, but I can change it nevertheless.

spec/support/system_description_factory.rb
fschueller: The code at least covers this case, in ``` plugins/users/users_renderer.rb```, line 25.

PR #2058, opened 2016-04-07
Issue #2047: Error for Accessing Broken System Description

### LINKED ISSUE #2047: Machinery Error for Accessing Broken System Description
If I start ```machinery serve``` and try to access a broken system description via e.g. ```http://127.0.0.1:7585/test-props``` (see related Issue#2046), I'll get an error:

Backtrace:
[Machinery-Errors-SystemDescriptionError-at_test-props.html.gz](https://github.com/SUSE/machinery/files/197865/Machinery-Errors-SystemDescriptionError-at_test-props.html.gz)

fschueller:
Fixes #2047, in which directly accessing a broken description after calling ```machinery serve``` results in an exception page. 
We now show a HTML view with the error, but maybe we should clarify the error further?


html/broken_desc.html.haml
```diff
@@ -0,0 +1,49 @@
...
+        .row
+          .col-xs-11
+            - if @error
+              .scope#alert_container
+                .row
+                  .col-xs-10.col-xs-offset-1
+                    .well
+                      %span
+                        %p
+                          = @error
```
thardeck: Can't we do this with partials so at least the header is not redundant on many pages?

fschueller: Sure!

html/exception.html.haml
```diff
@@ -2,7 +2,7 @@
 %html
   %head
     %title
-      Machinery Error: incompatible system description
+      Machinery error
```
thardeck: I would capitalize "error"

html/partials/exception/broken_desc.html.haml
```diff
@@ -0,0 +1,6 @@
+%h1
+  System Description broken!
+.row
+  .col-xs-1
+  .col-xs-4
+    Broken manifest of system description
```
thardeck: I do not think that this explanation is needed since it is kind of redundant in regard to the real error message.

html/partials/exception/upgrade.html.haml
```diff
@@ -0,0 +1,6 @@
+%h1
+  System Description incompatible!
+.row
+  .col-xs-1
+  .col-xs-4
+    Incompatible format version of system description
```
thardeck: Same here.

thardeck:
lgtm

PR #2059, opened 2016-04-08
Issue #328: Use more discoverable and standard locations to store Machinery data

### LINKED ISSUE #328: Use more discoverable and standard locations to store Machinery data
We currently store all Machinery data in `~/.machinery`. This includes the system descriptions, the log file, and the configuration file. This approach has two issues, on one hand it doesn't follow the XDG Base Directory standard for configuration files, and on the other hand it gives the impression that the system descriptions are not to be accessed by the user, because they are in a hidden directory.

Following up on a discussion in #268 I would like to propose the following solution:

* Use the XDG standard to store the configuration file. By default it would end up in `~/.config/machinery.conf`.
* Ask the user when running Machinery where to store the system descriptions and provide a default of `~/machinery-descriptions`. This would give users the option to choose according to their preferences, and make it clear and explicit for users, where the data will be stored.
* Store the log file in the descriptions directory. The log file is associated with the descriptions, so this is a natural place for it.

promulo:
Fixes #328 

This PR splits Machinery's config and data directories in order to follow the [XDG standard](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html).

The default config directory would now be `~/.config/machinery` and the default data directory (where the descriptions and logs are stored) would now be `~/.local/share/machinery`.

Thanks for reviewing!


cornelius:
This patch doesn't address storing the descriptions in a visible directory. It also doesn't take care of migrating existing data. I think we would need both to be able to apply that.

Given that we have used the `.machinery` directory for quite some time now, it's also the question, if it wouldn't be more confusing for users to move it, than keeping it as it is.

promulo:
Hi @cornelius , sorry about the delay. I have used the default options of the XDG standard, so changing that to a visible directory would not be a big deal. But of course, there should be just an agreement on which directory would it be. :) The data migration issue could be addressed by creating a new command.

Let me know your thoughts, if you think that keeping the status quo is the better option for now, we can close the PR and maybe reopen in the future. :)

suse-tests-pass:
Well Done! Your tests are still passing. (Merged 5b07b0a2f594adef09d9523d32cf9910d38b9b0d into 189b0e2c4742d6f3cd307a6dd3de42768ae41501)

opensusegithubapi:

opensusegithubapi:

thardeck:
We had some issues with our ci which created too many statuses for longer lasting prs, so I will disable this pr for now.
This should not happen for future prs, if the last commit has changed.

PR #2061, opened 2016-04-12
Issue #1904 Adjustment of integration tests

### LINKED ISSUE #1904: Fields of text output are not explained

fschueller:
Fixes #1904, adding correct integration test data
Supersedes #2060

### DIFF
--- a/spec/data/groups/opensuse131
+++ b/spec/data/groups/opensuse131
@@ -1,3 +1,6 @@
+Note: 'N/A' represents a missing group ID which could not
+be gathered during inspection.
+
 # Groups [192.168.121.137] (2014-11-25 18:21:52)
 
   * + (gid: N/A)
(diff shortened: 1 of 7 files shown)


thardeck:
lgtm

PR #2066, opened 2016-04-21
Use /etc/sudo for Tumbleweed integration test

thardeck:
because /etc/auto.master is not available if remote fs mounts are
disabled.


aduffeck:
lgtm

PR #2069, opened 2016-04-22
Disable unnecessary syntax checks

thardeck:
We have unit and integration tests and if we want to optimize syntax
issues we should run this as a separate test in Jenkins.


mauromorales:
Codewise LGTM but if it's something we think is useful we should put the check somewhere else like in our long tests

PR #2070, opened 2016-04-22
Add missing release checks

thardeck:
The integration test json link did not work in our Jenkins
implementation previously but it seems to be fixed.
So we add the check to make sure that all external tests are green
before release.


aduffeck:
lgtm

PR #2075, opened 2016-04-28
Fix syntax warnings

thardeck:


lib/kiwi_config.rb
```diff
@@ -105,9 +105,11 @@ def inject_users_and_groups(output_location)
         u.max_days, u.warn_days, u.disable_days, u.disabled_date, ""].join(":")
       "['#{passwd}', '#{shadow}']"
     end.join(",\n")
+    passwd_entries = passwd_entries # This variable is used in erb - prevent Syntax warning
```
aduffeck: I don't think we should add hacks like that just to prevent a syntax warning. In this case I'd rather just ignore the warning.

lib/kiwi_config.rb
```diff
@@ -105,9 +105,11 @@ def inject_users_and_groups(output_location)
...
     group_entries = @system_description.groups.map do |g|
       "'#{g.name}:#{g.password}:#{g.gid}:#{g.users.join(",")}'"
     end.join(",\n")
+    group_entries = group_entries # This variable is used in erb - prevent Syntax warning
```
aduffeck: Same here.

PR #2078, opened 2016-05-09
Filter .dockerinit in integration tests

thardeck:
because the new Docker version does not seem to create this file
anymore.


mauromorales:
lgtm

PR #2079, opened 2016-05-10
Inspect service instances

mauromorales:
There is a problem with the way machinery inspects for services in a system using systemd. Machinery extracts the information by listing the unit files and filtering any of those who are template files  (identified by having an @ sign) i.e. `getty@.service`.

By doing this Machinery doesn't show any service that was instantiated. What is worse the general status of the service is not even listed which can prove to be conflicting when comparing to packages. For example in the case of the "redis" package on openSUSE Tumbleweed the package will be present but the service isn't listed at all.

With this change I propose that whenever there is a service which is meant to be started as instances we either display:

- **When NO instance was enabled:** The template showing a "disabled" state
- **When at least one instance is enabled:** the detail of which instances where enabled and which disabled.

**Example**: Given a system with 2 services that can be instantiated "getty" and "redis". 

- Two instances of "getty" have been started but only one is enabled.
- No instance of "redis" was enabled.

Then Machinery's result would be the following:

```
getty@tty1.service    enabled
getty@tty2.service    disabled
redis@.service    disabled
```


suse-tests-pass:
Well Done! Your tests are still passing. (Merged 0f0699129d45c5d8c561178266816379683039f0 into 189b0e2c4742d6f3cd307a6dd3de42768ae41501)

plugins/services/services_inspector.rb
```diff
@@ -51,24 +51,55 @@ def summary
...
   def inspect_systemd_services
-    output = @system.run_command(
+    services = list_unit_files.map do |line|
+      name, state = line.split(/\s+/)
+
+      Service.new(name: name, state: state)
+    end
+
+    list_unit_files(true).map { |line| line.split(/\s+/) }.each do |name, state|
+      services << Service.new(name: name, state: state) if state == "disabled"
```
aduffeck: It's not very clear from the code why `disabled` services need to be treated separately here. Do you think we can clear this up somehow?

plugins/services/services_inspector.rb
```diff
@@ -51,24 +51,55 @@ def summary
...
+
+      prefix, suffix = name.split("@")
+      list_unit_instances.select { |i| i =~ /#{prefix}.*#{suffix}/ }.each do |instance|
```
aduffeck: Should we maybe include the `@` in the regex in order to make sure to map a wrong service?

mauromorales:
@aduffeck thanks with your review I realized the regex not only was missing the @ but also needs to enforce that at least there is 1 character between the @ and the .

Also for the other suggestion let me know if it makes it clearer now

aduffeck:
lgtm

PR #2083, opened 2016-05-11
Support old systemd version

mauromorales:


thardeck:
lgtm

thardeck:
I would ask a systemd maintainer or experienced person if something else is missing.

PR #2091, opened 2016-05-15
Remove Tumbleweed image date test

thardeck:
This test is just annoying and does not add much value. We had no issues
with old Tumbleweed images since the introduction of the test so I think
it is safe to remove it.


mauromorales:
lgtm

PR #2096, opened 2016-05-20
Issue #2095 Fix headline of initial dialogue

### LINKED ISSUE #2095: Title of initial dialog of HTML view is incosistent
When starting the HTML view without a description being selected, the dialog says "Select a description from the list below to show". It should be "Select a description from the list below to show its details." to be consistent with what we show, when a description already is selected.

fschueller:
Fixes #2095


html/partials/description_selector.html.haml
```diff
@@ -8,7 +8,7 @@
         %h4.modal-title#description-selector-label
           Select a description from the list below to
           %span.description-selector-action
-            show
+            show its details
```
mauromorales: Could you add a dot at the end so that it's exactly the same with the show view?

mauromorales:
lgtm just squash before merging

PR #2098, opened 2016-05-24
Issue #2095 Fix headline of initial dialogue

### LINKED ISSUE #2095: Title of initial dialog of HTML view is incosistent

fschueller:
Fixes #2095
Supersedes #2096


mauromorales:
lgtm

PR #2100, opened 2016-05-24
Issue #2097 Inconsistent HTML view for unmanaged files

### LINKED ISSUE #2097: Different output of "show" vs. "show --html" command for file_objects/files/dirs
Related to "Separate directories and files count in metadata" from format version 10 on.
While old descriptions are converted correctly, and the text output of show lists file_objects, e.g.:

Converted from former version ("machinery show" output snippet):
```
  * /boot/grub2/fonts/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 1.4 MiB
    File Objects: 2
```

Inspected with format version 10 ("machinery show" output snippet):
```
  * /boot/grub2/fonts/ (dir)
    User/Group: root:root
    Mode: 755
    Size: 1.4 MiB
    Files: 1
    Subdirectories: 0
```

the html view is statically listing "Files in dir." which in this example is empty, since it can't be calculated.

Converted from former version ("machinery show --html" output snippet):
```
/boot/grub2/fonts/,dir,755,root,root,1.4 MiB,
```

Inspected with format version 10 ("machinery show --html" output snippet):
```
/boot/grub2/fonts/,dir,755,root,root,1.4 MiB,1
```

To have html view consistent with the text output we need:
- in html view also vary between "Files in dir." (files) and "File Objects" (file_objects)
- introduce "Subdirectories" (dirs) to html view, too

fschueller:
Fixes #2097 
Supersedes #2099

### DIFF
--- a/html/partials/unmanaged_files.html.haml
+++ b/html/partials/unmanaged_files.html.haml
@@ -25,7 +25,11 @@
                     %th.medium User
                     %th.medium Group
                     %th.medium Size
-                    %th.medium Files in dir.
+                  - if unmanaged_files.any?(&:file_objects)
+                    %th.medium File Objects
+                  - else
+                    %th.medium Subdirectories
+                    %th.medium Files in Directory
               %tbody
                 - unmanaged_files.each do |file|
                   %tr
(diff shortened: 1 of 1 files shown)


mauromorales:
lgtm

PR #2102, opened 2016-05-30
Issue #2094 Comparison view displays single results inconsistently

### LINKED ISSUE #2094: Comparison view displays single results inconsistently
When viewing the comparison of two descriptions in the HTML view, the results get shown underneath their respective description title. However, if there's only a single result (e.g. there are only patterns existing in one of the descriptions), all content gets shifted to the left. This breaks the consistency of the view for the description shown on the right.

snapshot3

fschueller:
Fixes #2094 

HAML makes it hard here to add a condition for shifting the container around properly (e.g. add an offset). The easiest solution I found is to add a message which expands the bootstrap container in case there is only a single `@diff`` element (refer to screenshot in linked issue).


html/partials/compare/changed_config_files.html.haml
```diff
@@ -23,10 +23,14 @@
             - if @diff["changed_config_files"].only_in1
               = only_in_a
```
mauromorales: I'm thinking we could keep the titles out of the `if` clause so we always show them. Otherwise having one block with a title and one without might look a bit off

html/partials/compare/changed_config_files.html.haml
```diff
@@ -23,10 +23,14 @@
...
               = render_partial "compare/changed_config_files_list", list: @diff["changed_config_files"].only_in1
+            - else
+              No individual changed config files found.
```
mauromorales: For the message maybe we could use a helper method so in case we do a change we don't need to change it on every instance. We could pass the scope name as an argument.

Also not sure if the word "individual" is necessary here.

fschueller: That's probably a far better approach, I will look into that. 
I chose to add the "individual" to stress that it's a unique pattern etc. that's missing, but that could also be phrased differently of course.

mauromorales:
lgtm

PR #2103, opened 2016-05-31
Issue #2094 Comparison view displays single results inconsistently

### LINKED ISSUE #2094: Comparison view displays single results inconsistently

fschueller:
Fixes #2094 
Supersedes #2102


mauromorales:
lgtm

PR #2106, opened 2016-06-09
Add query command feature

felixsch:
Hey Hey,

Due lack of creativity what to program next (at home and break),  I looked up the mentoring page of openSUSE and found [this](https://github.com/openSUSE/mentoring/issues/33) nice project.
I implemented a basic version using [ragel](http://www.colm.net/open-source/ragel/) to parse a simple query language.

A query consists of _scope element[:attribute] operator value_ such as
- `packages zypper:version > 1.9.10`
- `users root exist`
- `packages zypper exist && packages zypper:version > 1.9`

Example:

```
x250 ❯ machinery query "packages zypper:version > 1.10.0 || users felixsch exist" 
Matched: 192.168.2.112
  -- felixsch exists
  ++ 1.12.37 > 1.10.0

Matched: 192.168.2.115
  ++ felixsch exists
  ++ 1.12.37 > 1.10.0

Matched: opensuse132
  -- felixsch exists
  ++ 1.11.14 > 1.10.0

Matched: opensuse_leap
  -- felixsch exists
  ++ 1.12.23 > 1.10.0

Total found: 4
```

**Current problems**:
- **Casting issues**:
  -  Currently its possible to run a query without attribute tag -> How to convert a element when compared to an integer `e.g packages zypper >= 1.9.10` (default to element.name?)
  - I'm not really sure how to safely parse version numbers without breaking all other number comparisions. Currently the code uses a regex to determine if a string has the form "1.9xx" or "1.9.1xx", everything else is casted to integer.
- **Error handling**: Currently the code prints out where an parsing error was found or when parsing failed. I bet this can improved
- **Output**: I implemented a basic output which is able to show what matched and what failed to match. `e.g ++ 1.9.16 == 1.9.16` maybe this can be made more readable
- **Performance**: Because it works file based, speed depends on the performance of the json parser. I implemented threaded reading but this will not solve the problem if, say more than 200 manifest needs to be read

What do you think is needed to make this a successfull pull request?


mschnitzer:
@felixsch Try to indent with 2 spaces instead with > 2 - Then it's consistent to the rest of the code.

felixsch:
This file is generated by ragel (see query_parser.rl) and every time anything is changed, ragel will regenerate the file. Does it make sense to reformat every time by hand? (You will never edit this file anyway, or you doing something wrong)

thardeck:
Closing for now, I guess we have to sit together and discuss this with @cornelius.

felixsch:
yes definitely. ping me if there is time!

PR #2107, opened 2016-06-09
Add attribute in patterns scope to identify the patterns manager

fschueller:
* The patterns scope includes an attribute identifying how patterns are managed (tasksel on Debian systems, zypper on SUSE systems)
* Existing descriptions get the patterns manager flag added based on the information from the packages scope.
* If during the migration the packages scope is not there the flag falls back to zypper and the user is shown a warning about it.


html/partials/patterns.html.haml
```diff
@@ -6,7 +6,7 @@
         :scope => "patterns",
         :title => "Patterns",
         :count => "#{patterns.length} #{Machinery.pluralize(patterns.length, "pattern")}"
-      - if @description.packages && @description.packages.package_system == "dpkg"
+      - if @description.patterns.patterns_system == "dpkg"
```
thardeck: What happens if the value is not there for older descriptions? Is this protected by the format version check?

plugins/patterns/patterns_renderer.rb
```diff
@@ -23,10 +23,13 @@ def content(description)
       puts "There are no patterns or tasks."
     end
 
-    if description.packages && description.packages.package_system == "dpkg"
+    if description.patterns.patterns_system == "tasksel"
       puts "Note: Tasks on Debian-like systems are treated as patterns."
     end
 
+    puts "Pattern Manager: #{description.patterns.patterns_system}" unless \
+      description.patterns.empty?
```
thardeck: The "\" should not be needed, but personally I would prefer to put the if statement on top if it does not fit the line. But I am not sure about the style guide.

thardeck: After talking to the team I would just remove the \

plugins/patterns/schema/system-description-patterns.schema-v10.json
```diff
@@ -7,7 +7,22 @@
...
-      "type": "object"
+      "type": "object",
+      "required": [
+        "patterns_system"
+      ],
+      "properties": {
+        "patterns_system": {
+          "oneOf": [
+            {
+              "enum": ["zypper"]
```
thardeck: Shouldn't it be?
```
"enum": ["zypper", "tasksel"]
```

schema/migrations/migrate9to10.rb
```diff
@@ -29,5 +31,25 @@ def migrate
         end
       end
     end
+
+    if @hash.key?("packages")
```
thardeck: I would only run this part if patterns are there, otherwise people might get the warning without them.

schema/migrations/migrate9to10.rb
```diff
@@ -29,5 +31,25 @@ def migrate
...
+      patterns_system = if @hash["packages"]["_attributes"]["package_system"] == "dpkg"
+        "tasksel"
+      else
+        "zypper"
+      end
+    else
+      patterns_system = "zypper"
+      Machinery::Ui.puts("Warning: No packages scope found. Patterns system defaults to zypper.")
```
thardeck: Use `Machinery::Ui.puts` should be `Machinery::Ui.warning`

plugins/patterns/schema/system-description-patterns.schema-v10.json
fschueller: I basically copied the concept from the v10 packages schema (there's a lot more duplication in there as it doesn't use ```oneOf```) but I can also change it I guess.

spec/support/system_description_factory.rb
```diff
@@ -473,6 +477,21 @@ def build_description(name, store, options)
...
+  EXAMPLE_SCOPES["patterns_with_dpkg"] = <<-EOF.chomp
+    "patterns": {
+      "_attributes" :
+        {
+          "patterns_system": "tasksel"
+        },
+      "_elements": [
+        {
+          "name": "base",
+          "version": "13.1",
```
thardeck: Does tasksel also have these version and releases?

fschueller: No, it actually doesn't. My bad

spec/unit/migrations/migrate9to10_spec.rb
```diff
@@ -92,4 +190,28 @@
...
+    migration.migrate
+
+    expect(description_hash_with_rpm["patterns"]["_attributes"]["patterns_system"]).to eq("zypper")
+  end
+
+  it "adds the attribute patterns_system and sets it to tasksel for dpkg based systems" do
+    migration = Migrate9To10.new(description_hash_with_dpkg, description_base)
+    migration.migrate
+
+    expect(description_hash_with_dpkg["patterns"]["_attributes"]["patterns_system"]).to \
```
thardeck: no `\` needed

spec/unit/patterns_inspector_spec.rb
```diff
@@ -125,6 +125,13 @@
         expect { patterns_inspector.inspect(filter) }.to raise_error(
           Machinery::Errors::ZypperFailed, /Zypper is locked./)
       end
+
+      it "returns patterns_system zypper" do
+        expect(system).to receive(:run_command).and_return(zypper_output)
+        patterns_inspector.inspect(filter)
+        patterns_system = description.patterns.attributes["patterns_system"]
```
thardeck: I don't think two lines are needed here unless it is bigger than 100 characters.

spec/unit/patterns_inspector_spec.rb
```diff
@@ -149,6 +156,13 @@
 
         expect(patterns_inspector.summary).to include("Found 5 patterns")
       end
+
+      it "returns patterns_system tasksel" do
+        expect(system).to receive(:run_command).and_return(zypper_output)
```
thardeck: Shouldn't it be `dpkg_output`?

thardeck:
lgtm

PR #2108, opened 2016-06-13
Add attribute in patterns scope to identify the patterns manager

fschueller:
* The patterns scope includes an attribute identifying how patterns are managed (tasksel on Debian systems, zypper on SUSE systems)
* Existing descriptions get the patterns manager flag added based on the information from the packages scope.
* If during the migration the packages scope is not there the flag falls back to zypper and the user is shown a warning about it.

Supersedes #2107


thardeck:
lgtm

PR #2110, opened 2016-06-14
Adapt patterns test JSON

fschueller:
Related to https://github.com/SUSE/machinery/pull/2108. 
Just a minor fix in the test data to align with the actual patterns JSON schema


thardeck:
The patterns example in the factory was not adapted.

fschueller:
https://github.com/SUSE/machinery/blob/b1646e635cb71e9f97bd2442e7a7e6c5fae8f856/spec/support/system_description_factory.rb
Fixed in #2108?

thardeck:
lgtm

Issue #2101, opened 2016-05-24
Compare does not handle new types of file_objects/dirs/files


svollath:
Also related to "Separate directories and files count in metadata" from format version 10 on.

```machinery compare --html <desc1(v10)> <desc2(v10)>``` still lists "Files in dir." only, and does not list "Subdirectroies" and "Files in Directory", like with ```machinery show --html <desc1(v10)>```

In addition, users will be able to compare
- a converted description (file_objects) vs. a current (dirs/files) ((I guess we can't list any counts then))
- two converted description (file_objects) - where we could list "File Objects"

The text output (compare --show-all) lists either "file_objects" or "dirs/files" - while currently also lists just file_objects, if a new description is compared to a converted one.

I suggest that all 3 cases listed above will result in the same output for
```compare --show-all``` and
```compare --html```

PR #2114, opened 2016-06-17
Issue #2101 Show file_objects/subdirs/files in comparison view

fschueller:
Fixes https://github.com/SUSE/machinery/issues/2101

### DIFF
--- a/html/partials/unmanaged_files.html.haml
+++ b/html/partials/unmanaged_files.html.haml
@@ -25,11 +25,11 @@
                     %th.medium User
                     %th.medium Group
                     %th.medium Size
-                  - if unmanaged_files.any?(&:file_objects)
-                    %th.medium File Objects
-                  - else
-                    %th.medium Subdirectories
-                    %th.medium Files in Directory
+                    - if unmanaged_files.any?(&:file_objects)
+                      %th.medium File Objects
+                    - else
+                      %th.medium Subdirectories
+                      %th.medium Files in Directory
               %tbody
                 - unmanaged_files.each do |file|
                   %tr
--- a/html/partials/unmanaged_files.html.haml
+++ b/html/partials/unmanaged_files.html.haml
@@ -20,16 +20,16 @@
                 %tr
                   %th Name
                   %th.medium Type
-                  - if unmanaged_files.has_metadata
+                  - if unmanaged_files.contains_metadata?
                     %th.medium Mode
                     %th.medium User
                     %th.medium Group
                     %th.medium Size
-                    - if unmanaged_files.any?(&:file_objects)
-                      %th.medium File Objects
-                    - else
+                    - if unmanaged_files.has_subdir_count?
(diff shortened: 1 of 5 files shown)


svollath:
lgtm

plugins/unmanaged_files/unmanaged_files_model.rb
```diff
@@ -73,6 +73,14 @@ def compare_with(other)
     comparison.push(common)
   end
 
+  def has_metadata?
+    @metadata ||= has_metadata || elements.any?(&:user)
+  end
+
+  def file_objects?
```
aduffeck: I think this could be a more speaking name, maybe negate the value and call it `has_subdir_counts` or something like that? `file_objects` is just an internal name and doesn't really say what it means.

plugins/unmanaged_files/unmanaged_files_model.rb
```diff
@@ -73,6 +73,14 @@ def compare_with(other)
     comparison.push(common)
   end
 
+  def has_metadata?
+    @metadata ||= has_metadata || elements.any?(&:user)
```
aduffeck: Why do we need this `elements.any?(&:user)` part? The user attribute should only be there when the metadata was extracted (and thus `has_metadata` is true) anyway, no?

html/partials/unmanaged_files.html.haml
```diff
@@ -25,11 +25,11 @@
                     %th.medium User
                     %th.medium Group
                     %th.medium Size
-                  - if unmanaged_files.any?(&:file_objects)
-                    %th.medium File Objects
-                  - else
-                    %th.medium Subdirectories
-                    %th.medium Files in Directory
+                    - if unmanaged_files.any?(&:file_objects)
```
aduffeck: This could also make use of the new method.

plugins/unmanaged_files/unmanaged_files_model.rb
```diff
@@ -73,12 +73,12 @@ def compare_with(other)
     comparison.push(common)
   end
 
-  def has_metadata?
-    @metadata ||= has_metadata || elements.any?(&:user)
+  def contains_metadata?
+    @metadata ||= has_metadata || !!elements.any?(&:user)
```
aduffeck: You shouldn't need the double bang because `any?` returns a boolean already.

plugins/unmanaged_files/unmanaged_files_model.rb
```diff
@@ -73,12 +73,12 @@ def compare_with(other)
...
   end
 
-  def file_objects?
-    @file_objects ||= elements.any?(&:file_objects)
+  def has_subdir_count?
```
aduffeck: I think `has_subdir_counts?` might be more appropriate because it's a method of the list, and the list has multiple counts, one per element.

aduffeck:
lgtm :tada:

PR #2115, opened 2016-06-21
 Issue #2101 Show file_objects/subdirs/files in comparison view

### LINKED ISSUE #2101: Compare does not handle new types of file_objects/dirs/files

fschueller:
Workaround for showing number of file objects and/or number of files and
subdirectories in unmanaged file comparison.

Fixes #2101

### DIFF
--- a/html/partials/compare/unmanaged_file_list.html.haml
+++ b/html/partials/compare/unmanaged_file_list.html.haml
@@ -7,20 +7,28 @@
       %tr
         %th Name
         %th.small Type
-        - if list.has_metadata
+        - if list.contains_metadata?
           %th.medium Mode
           %th.medium User
           %th.medium Group
           %th.medium Size
-          %th.medium Files in dir.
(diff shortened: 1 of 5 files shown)


aduffeck:
:shipit:

PR #2120, opened 2016-07-05
Fix filter pathes in documentation

thardeck:
We seem to have not updated the filter documentation after changing the
data model.


mauromorales:
lgtm

PR #2121, opened 2016-07-05
Fix scope names in documentation

### LINKED ISSUE #2117: Scopes are not displayed correctly in HTML manual
Hi, 

a user on our public mailing list figured out, that the scopes are named incorrectly or at least are just displayed incorrectly in our HTML manual.

When going to http://machinery-project.org/machinery_main_scopes.1/ the name of the scopes are shown with underscores instead of dashes. In the markdown file, they're named correctly. So it's an error while generating the HTML manual.

I'm also a bit confused why the file `manual/docs/machinery_main_scopes.1.md` is gone. It's actually in my local checkout and I'm in the master branch and up-to-date.

thardeck:
The scope names on the user side should have dashes instead of
underscores.

Fixes #2117

### DIFF
--- a/lib/man_task.rb
+++ b/lib/man_task.rb
@@ -18,7 +18,7 @@
 class ManTask
   def self.compile_documentation
     docs = Inspector.all_scopes.map do |scope|
-      scope_doc = "* #{scope}\n\n"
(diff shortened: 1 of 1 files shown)


aduffeck:
lgtm

PR #2130, opened 2016-07-19
Fix check_retrieve_files_dependencies for remote system

### LINKED ISSUE #2118: rsync is not installed on minimal systems
A user reported an issue that `rsync` is not always installed on openSUSE systems and thus Machinery fails when extracting files. For example, on minimal installation systems.

fschueller:
Remote files are transferred using the local rsync so the check needs to
happen on the local system.

Fixes #2118 
Supersedes #2129

### DIFF
--- a/lib/remote_system.rb
+++ b/lib/remote_system.rb
@@ -145,6 +145,10 @@ class RemoteSystem < System
     end
   end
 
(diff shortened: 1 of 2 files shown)


aduffeck:
:shipit:

PR #2142, opened 2016-09-09
Replace click_on with trigger("click") to make capybara work in docker

greygoo:


greygoo:
source of problem was missing fonts in the docker container.

PR #2144, opened 2016-09-22
Fix broken exclude files command with new tar

manno:
The  system unit spec fails, because the excluded files are not excluded from the tar ball.
OpenSuse Tumbleweed uses tar 1.29, moving the exclude arguments in front of the file list allows tar to actually exclude the files.


thardeck:
Wouldn't this break the unit test for Leap and other systems which still have the old tar?

Issue #2119, opened 2016-06-28
machinery export description to html format


MalloZup:
### How To Reproduce:

I'm using machinery for automation in a QA-automation environment, machinery works well together in a Jenkins server, plus other stuff. 

I really like the inspect functionality of machinery, and for QA-automation is a good feature.

The  only one issue i found, is that, after i get the description json, i cannot really have an html file of the description.

So : 

```
machinery inspect IP_ADRESS_OF_SUT(Systems under Tests)
```
I have the description. This is stored in json format.

I would like to have a feature like this:

```
machinery convert-html IP_ADRESS_OF_SUT  
```
With this feature i could then copy the .html file and put it in the workspace dir of Jenkins, or to a web-server.

In my env. i cannot use machinery serve  or show --html because: 
*1) i don't want to expose the jenkins server ip
*2) this would be not nice for  the testing workflow. 
3) i don't want to create a web-server

mschnitzer:
A temporary workaround might be: Start a web server, download with curl the HTML view.

(If you want a quick solution)

MalloZup:
@mschnitzer  thx. i can make a short hack for that :) helps a quick solution :)

thardeck:
This feature was added to master and a new version will be released soon.

thardeck:
@MalloZup The version for openSUSE was released and SLES is in the maintenance queue.

PR #2145, opened 2016-09-23
Export Html

manno:
Add command to create an HTML export of a system description. This should solve https://github.com/SUSE/machinery/issues/2119


thardeck:
I would add a unit test which checks the static html code of a small description against a static file under `spec/data`, to make sure that nothing breaks in the code.

html/static_index.html.haml
```diff
@@ -0,0 +1,76 @@
+!!!
```
aduffeck: Couldn't the HTML file be shared between the export and the server-rendered version?

lib/cli.rb
```diff
@@ -513,6 +513,33 @@ def self.check_port_validity(port)
     end
   end
 
+  desc "Export static HTML view of a system description"
+  long_desc <<-LONGDESC
+    Export system description as AutoYaST profile
```
aduffeck: That doesn't sound right ;)

lib/cli.rb
```diff
@@ -513,6 +513,33 @@ def self.check_port_validity(port)
...
+
+    The HTML view will be placed in a subdirectory at the given location by the 'html-dir'
+    option.
+  LONGDESC
+  arg "NAME"
+  command "export-html" do |c|
+    c.flag ["html-dir", :d], type: String, required: true,
+      desc: "Location where the HTML view will be stored", arg_name: "DIRECTORY"
+    c.switch :force, default_value: false, required: false, negatable: false,
+      desc: "Overwrite existing profile"
```
aduffeck: `profile` also doesn't make sense here

spec/unit/static_html_spec.rb
```diff
@@ -44,5 +41,21 @@
...
+  end
+
+  describe "#write" do
+    it "renders an HTML report" do
+      static_html = StaticHtml.new(description)
+      Dir.mktmpdir do |tmp_dir|
+        static_html.write(tmp_dir)
+        index_file = File.join(tmp_dir, "index.html")
+        expect(File.readable?(index_file)).to be_truthy
+        expect(File.read(index_file)).to include "<html", "openSUSE"
```
aduffeck: Hound doesn't cover it, but according to our [style guide](https://github.com/SUSE/style-guides/blob/master/Ruby.md) you should use parentheses for all method calls (with exceptions for statement like calls like `puts foo` etc.)

lib/static_html.rb
```diff
@@ -17,12 +17,33 @@
...
+  include HamlHelpers
+
   def initialize(description)
     @description = description
-    Haml::Engine.new(File.read(File.join(Machinery::ROOT, "html/static_index.html.haml"))).render(Object.new, description: @description)
   end
 
-  def render_scope(partial)
-    Haml::Engine.new(File.read("_#{partial}.html.haml")).render
+  def haml(source, locals:)
```
aduffeck: Since this is only used in the `HamlHelpers` module, could it go there instead?

lib/static_html.rb
```diff
@@ -32,13 +32,22 @@ def haml(source, locals:)
...
   def write(directory)
     FileUtils.mkdir_p(directory)
+    render_html(directory)
+    copy_assets(directory)
+  end
+
+  private
+
+  def copy_assets(directory)
+    FileUtils.cp_r File.join(TEMPLATE_DIR, "assets"), directory
```
aduffeck: Again, there should be parentheses to comply to the style guide.

lib/cli.rb
```diff
@@ -530,13 +530,9 @@ def self.check_port_validity(port)
     c.action do |_global_options, options, args|
       name = shift_arg(args, "NAME")
       description = SystemDescription.load(name, system_description_store)
-      exporter = StaticHtml.new(description)
-
-      task = ExportTask.new(exporter)
```
aduffeck: I'm just curious, why doesn't ExportTask work here?

manual/docs/machinery-export-html.1.md
```diff
@@ -0,0 +1,32 @@
+# export-autoyast — Export System Description as HTML
```
aduffeck: This should be `export-html` instead.

lib/static_html.rb
```diff
@@ -0,0 +1,72 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+class StaticHtml < Exporter
+
```
aduffeck: There shouldn't be empty lines before or after the class body.

lib/server.rb
```diff
@@ -40,40 +40,7 @@ def call(env)
...
-
-    def scope_title(scope)
-      scope_info(scope)[:name]
-    end
-
-    def scope_initials(scope)
-      scope_info(scope)[:initials].upcase
-    end
-
+    include HamlHelpers
```
aduffeck: What's the separation logic for having helpers here or in `HamlHelpers`? Arent they all related to HAML?

html/static_index.html.haml
manno: Yes, I thought so too, but couldn't come up with a good switch to toggle the two blocks which are missing from the static html template.

Additionally static export is using an instance variable and the server is using `description`, but that might be fixed by an `attr_reader`.

lib/cli.rb
manno: ExportTask contains some unrelated logic, like checking for scopes, adding a sub directory below the target directory and applying filters.

lib/static_html.rb
manno: That might interact with Sinatra, because in `lib/server.rb` `haml` is a Sinatra method.

lib/server.rb
manno: While the remaining methods might be used in HAML templates, they're not required to render HAML templates with partials.

lib/static_html.rb
aduffeck: There's still the superfluous empty line.

html/static_index.html.haml
aduffeck: I think it would be good to do that if possible, otherwise we'll always have to think about doing chances in two places. I'm pretty sure that will be forgotten sooner or later.

lib/cli.rb
aduffeck: I see, thanks. Maybe adding it to the commit message would be nice to make that clearer.

thardeck:
lgtm

PR #2146, opened 2016-09-26
Remove Ruby 2.1 required keyword argument

manno:
The gemspec allows Ruby 2.0, which does not support required
keyword arguments. However 2.0 is end of live.

This PR removes the keyword argument. I'd prefer to bump the Ruby requirement, though.


thardeck:
lgtm

PR #2147, opened 2016-09-26
Bump Ruby version to 2.1.0

manno:
Try to run integration tests with Ruby 2.1


thardeck:
I would prefer to keep the requirement  like it is for now although Ruby 2.1 seems to be installed on SLES12.

PR #2153, opened 2016-10-04
Filter diversions from dpkg output

### LINKED ISSUE #2138: wrong dpkg-query call when executing "machinery inspect localhost" on Ubuntu 16.04
Hello there,

I recently installed Machinery 1.21 (along with all requirements) on Ubuntu 16.04 running in a KVM virtual machine.

When running 'machinery inspect localhost' as root, the program works for a while and then crashes at the step 'Inspecting changed-config-files' with an error that seems to indicate machinery tried to get information on a package named 'diversion by branding-ubuntu from' using dpkg-query, so the problem seems to be that this line sneaked into the list of package names somehow and the crash occurs because dpkg-query does not accept package names that contain a space. The log of the crash including backtrace is appended below.

Greetings and thanks in advance
Supersonic

[machinery-dpkg-error.txt](https://github.com/SUSE/machinery/files/455002/machinery-dpkg-error.txt)

**EDIT:**
_.machinery/machinery.log_ reveals that the last successful command was `dpkg -S /usr/share/gnome-mahjongg/themes/postmodern.svg` which returns these three lines:
`diversion by branding-ubuntu from /usr/share/gnome-mahjongg/themes/postmodern.svg`
`diversion by branding-ubuntu to: /usr/share/gnome-mahjongg/themes/postmodern.svg.unbranded`
`gnome-mahjongg, branding-ubuntu: /usr/share/gnome-mahjongg/themes/postmodern.svg`

thardeck:
Diversions are files owned by more than one package. The original file
gets renamed.

Fixes #2138

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -3,6 +3,8 @@
 * Add remaining systemd service states in Kiwi export (gh#SUSE/machinery#2122)
 * Add export of system descriptions as HTML page. This provides an offline
   view of descrptions (gh#SUSE/machinery#2119)
+* Fix changed files inspections on dpkg systems in case of diversions
+  (gh#SUSE/machinery#2138)
 
 ## Version 1.21.0 - Tue Jun 21 16:28:31 CEST 2016 - thardeck@suse.de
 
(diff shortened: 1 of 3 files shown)


manno:
LGTM

PR #2160, opened 2016-10-12
Adapt read file

thardeck:


NEWS
```diff
@@ -1,5 +1,6 @@
 # Machinery Release Notes
 
+* Only use sudo were appropriate for reading files (gh#SUSE/machinery#2077)
```
aduffeck: `where appropriate when reading`?

lib/remote_system.rb
```diff
@@ -150,8 +150,12 @@ def check_retrieve_files_dependencies
   end
 
   # Reads a file from the System. Returns nil if it does not exist.
-  def read_file(file)
-    run_command("cat", file, stdout: :capture, privileged: true)
+  def read_file(file, options = {})
+    if options[:privileged]
+      run_command("cat", file, stdout: :capture, privileged: true)
```
aduffeck: How about

```ruby
def read_file(file, options = {})
  command_options = {
    stdout: :capture,
    privileged: false
  }.merge(options)

  run_command("cat", file, command_options)
...
```

That makes the intention to run unprivileged by default explicit and removes a conditional and some duplication.

NEWS
aduffeck: Or `when necessary`?

spec/unit/remote_system_spec.rb
```diff
@@ -225,6 +225,16 @@
     end
 
     describe "#read_file" do
+      it "allows retrieving files without privileged mode by default" do
```
aduffeck: That reads a bit weird, what about `retrieves files in unprivileged mode by default`?

lib/remote_system.rb
aduffeck: Not exposing the cheetah options makes sense, but that could also be circumvented using

```ruby
command_options = {
  stdout: :capture,
  privileged: options.fetch(:privileged, false)
}
```
for example.

Having to adapt the tests is a relatively weak argument against clearer and more DRY code  imo ;)

PR #2162, opened 2016-10-13
Fix centos inspection

thardeck:


lib/exceptions.rb
```diff
@@ -124,6 +124,17 @@ def to_s
       end
     end
 
+    class SudoMissingTTY < MachineryError
+      def initialize(host)
+        @host = host
+      end
+
+      def to_s
+        "sudo isn't configured on the inspected host '#{@host}'." \
```
aduffeck: I think `isn't configured properly` would sound a bit better here.

PR #2166, opened 2016-10-14
Fix retrieve files dependencies

thardeck:


lib/remote_system.rb
```diff
@@ -130,7 +130,7 @@ def retrieve_files(filelist, destination)
   end
 
   def check_retrieve_files_dependencies
-    LocalSystem.new.check_requirement("rsync", "--version")
+    LocalSystem.validate_existence_of_command("rsync", "rsync")
```
aduffeck: Is that supposed to say `"rsync", "rsync"`?

thardeck: Yes, the second is the package name for the error message.

PR #2169, opened 2016-10-18
Fix nfs analyze issue

thardeck:


lib/zypper.rb
```diff
@@ -86,10 +91,12 @@ def version
     [found[1].to_i, found[2].to_i, found[3].to_i] if found
   end
 
-  private
-
   def call_zypper(*args)
-    cmd = ["zypper"]
+    cmd = if args.last.is_a?(Hash) && args.last[:sudo]
```
aduffeck: From reading the code I'd think that you have to pop the last elements in the positive case because it would otherwise end up in the command. But why doesn't the test fail then? ...

lib/zypper.rb
```diff
@@ -27,32 +27,47 @@ class Zypper
 
   class <<self
     def isolated(options = {}, &block)
-      zypper_base = Dir.mktmpdir("machinery_zypper")
+      @zypp_base = Dir.mktmpdir("machinery_zypper")
```
aduffeck: Since this is a class method you're essentially creating an instance variable on the class object itself. That's bad and not thread safe. I'd suggest to revert to passing the base directory path around as parameters again or make it an attribute of the zypper object instead.

lib/zypper.rb
```diff
@@ -91,8 +91,6 @@ def version
     [found[1].to_i, found[2].to_i, found[3].to_i] if found
   end
 
-  private
```
aduffeck: What makes this obsolete? I think it still makes sense to restrict access to the actual `call_zypper` method.

lib/zypper.rb
```diff
@@ -27,32 +27,47 @@ class Zypper
...
+      end
+      false
     end
 
     private
 
-    def clean_up(base_path)
-      LoggedCheetah.run("sudo", "rm", "-rf", base_path)
+    def clean_up
+      if @zypp_base =~ /^\/tmp\/./
```
aduffeck: I'd prefer a guard clause here instead of the `if..else` construct.

aduffeck: And I'd even check for `/tmp/machinery_zypper` instead.

lib/zypper.rb
```diff
@@ -27,32 +27,47 @@ class Zypper
...
+        cmd = ["rm", "-rf", @zypp_base]
+        cmd = cmd.insert(0, "sudo") if nfs_repos?
+        LoggedCheetah.run(*cmd)
+      else
+        raise("This is supposed to never happen, only tmp directories should be used")
```
aduffeck: This message is not very helpful, how about `The zypper base directory is not inside of '/tmp'. Aborting...`.

lib/zypper.rb
```diff
@@ -92,7 +107,15 @@ def version
   end
 
   def call_zypper(*args)
```
aduffeck: I'd rewrite this to

```ruby
sudo = args.delete(:sudo) if args.last.is_a?(Hash)
...
cmd.unshift("sudo") if sudo
```

lib/zypper.rb
```diff
@@ -27,32 +27,47 @@ class Zypper
...
+        zypper.zypp_config = create_zypp_config(options[:arch])
       end
 
       block.call(zypper)
     ensure
-      clean_up(zypper_base)
+      clean_up
+    end
+
+    def nfs_repos?
```
aduffeck: This should not be a class method imo.

aduffeck: What about

```ruby
def contains_nfs_repos
  files = Dir.glob(File.join(@zypp_base, "etc/zypp/repos.d", "*"))
  files.any? do |file|
    File.readlines(file).any? { |line| line.start_with?("baseurl=nfs://") }
  end
end
```

PR #2196, opened 2016-11-10
Enhance debian lenny compatibility

thardeck:


thardeck:
All integration tests were green.

plugins/packages/packages_inspector.rb
```diff
@@ -90,12 +90,22 @@ def inspect_dpkg
...
-        stdout: :capture
-      )
+      begin
+        apt_cache_output = @system.run_command(
+          "apt-cache",
+          "show",
+          *packages_slice.map { |p| "#{p.name}=#{[p.version, p.release].compact.join("-")}" },
+          stdout: :capture
+        )
+      rescue
```
aduffeck: Using exceptions for things like that is quite a hack. Would it be possible to use the version of the command or so to know which format to use beforehand?

plugins/packages/packages_inspector.rb
```diff
@@ -90,12 +90,22 @@ def inspect_dpkg
...
+        # Older Debian systems do not support the version parameter for `apt-cache show`
+        apt_cache_output = @system.run_command(
+          "apt-cache",
+          "show",
+          *packages_slice.map(&:name),
+          stdout: :capture
```
aduffeck: I think we explicitly added the version and release to the query in order to not end up with confused data or something like that. Are you sure that this code works as expected?

plugins/packages/packages_inspector.rb
```diff
@@ -108,6 +118,7 @@ def inspect_dpkg
         name = package.name.sub(/:[^:]+$/, "")
         apt_cache_element = apt_cache_elements[name]
         if apt_cache_element
+          package.arch = apt_cache_element[/Architecture: (\w+)\n/, 1]
```
aduffeck: Why did you add this? It's not clear to me from the commit message.

plugins/repositories/schema/system-description-repositories.schema-v10.json
```diff
@@ -37,7 +37,6 @@
               },
               "url": {
                 "type": "string",
-                "format": "uri",
```
aduffeck: Could we somehow use a stricter schema, e.g. by using `oneOf`?

lib/exceptions.rb
```diff
@@ -28,7 +28,12 @@ class IncompatibleHost < StandardError; end
     class MachineryError < StandardError; end
 
     class UnknownScope < MachineryError; end
-    class UnknownOs < MachineryError; end
+    class UnknownOs < MachineryError
+      def to_s
+        "Unable to determine the operating system. There does not seem to be an" \
```
aduffeck: Should this be `a /etc/os-release` file? And it's not really correct because we also look at `/etc/SuSE-release` and `/etc/redhat-release`.

plugins/packages/packages_inspector.rb
thardeck: We could try to determine the version, but since it just calls the command again a little different and if it fails again, the inspection stops so I think it is Ok.

plugins/packages/packages_inspector.rb
thardeck: It worked for my Debian inspection and the alternative would be no results to my knowledge.

plugins/repositories/schema/system-description-repositories.schema-v10.json
thardeck: I think Cornelius wanted to be less strict if there are issues. For most parts we only require strings with at least one character. We though url should work here but like I mentioned Debian does not follow this schema.

I am not sure if there is a schema for the Debian cdrom urls. If not I am not sure how `oneOf` would help.

lib/exceptions.rb
thardeck: That is true and I thought about it but it would make the error message more confusing. The new standard is /etc/os-release which is used if it is there.
If it is not there we also look for the SUSE and RHEL-Version. Do you think this should be mentioned?

plugins/packages/packages_inspector.rb
manno: Is there a way to not have an exception for every hundred packages? 

This is bad for performance and I find it hard to read, exceptions shouldn't be part of the normal execution flow.
If the performance is no issue, it would be sufficient to extract the `apt_cache_output` assignment to methods to improve readability.

Also this method is too long, it would benefit from some "extract to method". 

Not really part of this PR, but the method names 'inspect_*' do hint at some hidden classes. In the future we might want to split this into a `DpkgInspector` and a `RPMInspector`.

manno:
LGTM

PR #2198, opened 2016-11-11
Do not extract Metadata for missing unmanaged files

thardeck:
Make sure that files exist before extracting the metadata and remove
them from the unmanaged-files list if not.

This should allow filtering to work properly because the filters are
applied before extraction so it is possible for the user to filter
directories with changing files.
Before that machinery-helper could crash if a file was removed during
the first phase of inspection and there was no way for the user to
filter these directories because the helper does not support filtering
yet.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,6 @@
 # Machinery Release Notes
 
+* Prevent machinery-helper from crashing when files are removed during inspection
 * Fix analyze of changed-config-files when NFS or SMB repositories are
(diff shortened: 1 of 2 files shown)


thardeck:
To test it you can run the following command inside a managed repository: `for i in `seq 1 600`; do touch "$i"; sleep 1; rm "$i"; done`

manno:
LGTM

PR #2204, opened 2016-11-15
Handle unextracted unmanaged files as extracted

thardeck:
Fixed regression where the machinery-helper reported different number of
files when extraction was used or not used.

Additionally Lstat was used to make sure that the check happens for the
actual link and not the referred file.

### DIFF
--- a/machinery-helper/machinery_helper.go
+++ b/machinery-helper/machinery_helper.go
@@ -357,7 +357,7 @@ func getUnmanagedFilesList(files []string, unmanagedFiles map[string]string, ext
 	i := 0
 	for j := range files {
 		// only add accessible files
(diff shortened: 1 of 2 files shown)


machinery-helper/machinery_helper.go
```diff
@@ -356,20 +356,21 @@ func getUnmanagedFilesList(files []string, unmanagedFiles map[string]string, ext
 	unmanagedFilesList := make([]UnmanagedFile, len(unmanagedFiles))
 	i := 0
 	for j := range files {
-		entry := UnmanagedFile{}
-		entry.Name = files[j]
-		entry.Type = unmanagedFiles[files[j]]
+		// only add accessible files
+		if _, err := os.Lstat(files[j]); err == nil || (os.IsNotExist(err) == false && os.IsPermission(err) == false && strings.Contains(err.Error(), "no such device") == false) {
```
manno: I think this line is too long. I'd prefer to get rid of the semicolon and do it in two lines.

svollath:
LGTM

PR #2205, opened 2016-11-16
Only search inside the Machinery namespace

thardeck:
Make sure that we only search for our scope classes inside of our
namespace.
This fixes a regression on 13.1 systems.


manno:
LGTM

PR #2216, opened 2016-12-23
Remove outdated documentation "Overall-Class-Design"

abodry:


cornelius:
@abodry Why is this outdated?

PR #2222, opened 2017-01-13
Error handling for a disrupted ssh connection

abodry:
Throwing an exception with an error message and aborting the
inspection if the ssh connection is disrupted.


lib/exceptions.rb
```diff
@@ -170,6 +170,7 @@ class UpgradeFailed < MachineryError; end
     class BinaryDiffError < MachineryError; end
 
     class SshConnectionFailed < MachineryError; end
+    class SshConnectionDisrupted < SystemExit; end
```
aduffeck: Yes, why do you inherit from SystemExit here?

lib/remote_system.rb
```diff
@@ -94,7 +94,25 @@ def run_command(*args)
       "LogLevel=ERROR", sudo, "LANGUAGE=", "LC_ALL=#{locale}", *piped_args, options
     ].compact.flatten
 
-    cheetah_class.run(*cmds)
+    begin
+      cheetah_class.run(*cmds)
+    rescue Cheetah::ExecutionFailed => e
+      if e.stderr && !(e.stderr =~ /Connection refused/).nil?
+        puts "\n#{e.stderr}"
+        raise Machinery::Errors::SshConnectionDisrupted.new("\nSSH ERROR: #{e.stderr}")
```
aduffeck: That's a lot of duplicated code. Can we merge this together somehow?

lib/remote_system.rb
```diff
@@ -140,13 +158,15 @@ def read_file(file, options = {})
...
       privileged: options.fetch(:privileged, false)
     }
-    run_command("cat", file, command_options)
-  rescue Cheetah::ExecutionFailed => e
-    if e.status.exitstatus == 1
-      # File not found, return nil
-      return
-    else
-      raise
+    begin
```
aduffeck: The `begin ... end`  should not be necessary. Is there a reason why you added it here?

lib/exceptions.rb
abodry: So that machinery exits and doesn't show a backlog.

lib/remote_system.rb
abodry: Well it looked cleaner indenting the rescue and wrapping it

lib/exceptions.rb
aduffeck: That's what we have the `MachineryError` exception class for, though. If you look at https://github.com/SUSE/machinery/blob/master/lib/cli.rb#L105 you'll see that these exceptions are handled properly and no backlog is shown.

lib/remote_system.rb
```diff
@@ -94,7 +94,19 @@ def run_command(*args)
       "LogLevel=ERROR", sudo, "LANGUAGE=", "LC_ALL=#{locale}", *piped_args, options
     ].compact.flatten
 
-    cheetah_class.run(*cmds)
+    begin
+      cheetah_class.run(*cmds)
+    rescue Cheetah::ExecutionFailed => e
+      if e.stderr && !(e.stderr =~ /Connection refused/).nil? \
```
aduffeck: How about

```ruby
ssh_error_regex = /Connection refused|Connection reset by peer|Broken pipe|Network is unreachable/
if e.stderr && e.stderr =~  ssh_error_regex
```

?

spec/unit/remote_system_spec.rb
```diff
@@ -117,6 +117,48 @@
     end
 
     describe "#run_command" do
+      context "when ssh connection is disrupted" do
+        it "raises SshConnectionDisrupted when sshd is stopped on remote machine" do
```
aduffeck: These tests could also be DRYed up, e.g. by doing something like

```ruby
["Connection refused", "Broken pipe", ...].each do |error|
  it "raises an error" do
    expect(Cheetah) blabla
    .and_raise(Cheetah::ExecutionFailed.new(nil, nil, nil, error))

    expect { blabla }.to raise_error(Machinery::Errors::SshConnectionDisrupted, /#{error}/)
  end
end
```

or so.

lib/exceptions.rb
aduffeck: Ah, I See. In this case I think we should rather handle `SshConnectionDisrupted` differently. Using the raw `SystemExit` class for that seems a bit crude to me.

aduffeck: Also, is it really a problem if the inspection continues? The problem could have been temporary and the next inspection works just fine, no?

abodry: No this can cause other functions to fail because they depend on data that should have been gathered.
* I meant i already tried it with RuntimeException but i get a backtrace when i use that.
I also thought about taking something different than SystemExit but i would probably need to look at the code that suppresses exceptions while inspecting and that would probably be too much for this card.

abodry: The probability of an error of this sort occurring while inspecting is already very low, i had to manually kill the tcp connection and shut down the sshd deamon to even replicate it. If it does get back up again you can try again, otherwise you would have to fix the connection first anyway. The error mentioned in the bug report isn't a timeout but rather a configuration problem because the machine shuts down the connection on purpose.

aduffeck: But we also have the problem of dependencies in every other case where on inspector failed, so I think we could just handle this case the same.

aduffeck:
lgtm

PR #2223, opened 2017-01-18
Catching Exception when a wrong Filter Path is given

abodry:


spec/unit/filter_spec.rb
```diff
@@ -317,6 +317,29 @@ def expect_file_scope_filter_change(scope, filter, before, after)
...
     end
+
+    it "raises error when given wrong path" do
+      expect {
+        expect_file_scope_filter_change(
+          "changed_managed_files",
+          Machinery::Filter.new(["/changed_managed_files/files/name=element_a,element_b"]),
+          [
+            "/etc/deleted changed managed",
+            "/etc/cron.d",
```
thardeck: I would use less examples, two are more than enough because they are not causing the issue. Even one line would be Ok.

lib/filter.rb
```diff
@@ -151,7 +151,11 @@ def apply!(system_description)
         container = nil
         steps.each do |step|
           break unless pointer
-          pointer = pointer[step]
+          begin
+            pointer = pointer[step]
+          rescue TypeError
+            raise Machinery::Errors::WrongFilterPath.new("Error: Check if the path: \"#{path}\" is correct.")
```
thardeck: We usually use single quotes for internal quotes.

spec/unit/filter_spec.rb
```diff
@@ -317,6 +317,29 @@ def expect_file_scope_filter_change(scope, filter, before, after)
...
+          ],
+          [
+            "/etc/deleted changed managed",
+            "/etc/cron.d",
+            "/etc/cron.daily/cleanup",
+            "/etc/cron.daily/logrotate",
+            "/usr/bin/replaced_by_link"
+          ]
+        )
+      }.to raise_error(Machinery::Errors::WrongFilterPath)
```
thardeck: I would also check for the message or at least a part of it.

lib/filter.rb
```diff
@@ -154,7 +154,7 @@ def apply!(system_description)
           begin
             pointer = pointer[step]
           rescue TypeError
-            raise Machinery::Errors::WrongFilterPath.new("Error: Check if the path: \"#{path}\" is correct.")
+            raise Machinery::Errors::WrongFilterPath.new('Error: Check if the path: \"#{path}\" is correct.')
```
thardeck: You have to remove the backslashes now.

spec/unit/filter_spec.rb
```diff
@@ -338,7 +338,7 @@ def expect_file_scope_filter_change(scope, filter, before, after)
             "/usr/bin/replaced_by_link"
           ]
         )
-      }.to raise_error(Machinery::Errors::WrongFilterPath)
+      }.to raise_error(Machinery::Errors::WrongFilterPath, /Error: Check if the path/)
```
thardeck: I would add the complete line, or at least with the path, then the issue with the backslashes would have been shown by the test.

PR #2232, opened 2017-06-19
Support Sinatra version 2

thardeck:
`bundle --standalone` seems to have an issue with multiple version
requirements, that's why we allow all versions >= 1.4.


dragonchaser:
lgtm

PR #2236, opened 2017-08-16
Fix Tumbleweed expectation in integration test

thardeck:
rc-local.service is not static anymore but enabled-runtime


svollath:
LGTM

PR #2238, opened 2017-08-23
Allow installation of machinery test package in Tumbleweed

thardeck:
Zypper in Tumbleweed does not allow to install unsigned packages anymore
with the non-interactive tag. Disabling the gpg-checks fixes the issue.

The RPMs are installed in a vm which is resetted all the time so it
should be not a significant problem from a security point of view.


thardeck:
I have talked to the developer and created a bug against zypper to prevent us from using this work around. https://bugzilla.suse.com/show_bug.cgi?id=1055533

thardeck:
Merging temporarily because fixing of the issue will take longer.

PR #2241, opened 2017-10-17
Remove deleted package from Tumbleweed test vm

thardeck:


aduffeck:
lgtm if all tests still pass with it

PR #2245, opened 2017-11-21
Allow inspection of 32-bit systems which have no SSE

### LINKED ISSUE #2243: machinery-helper-i686: illegal instruction
Hello,

I am trying to do `inspect -x` and see the following:

```
Inspecting changed-managed-files...
 -> Extracted 0 changed managed files.
Inspecting unmanaged-files...
 -> Inspection failed!


Errors while inspecting unmanaged-files:
 -> There is no machinery-helper available for the remote system architecture i586.
```

I understand that i586 is a bit outdated architecture. But machinery is a tool to copy system configurations, so I think there is a reason to support it in order to copy configs from very old legacy systems to be migrated.

thardeck:
SSE processor instructions are used by default when go cross compiles 32-bit on
a modern system.
This prevents the executable from working on older systems so we now compile
explicitly without these instructions.

Fixes #2243

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 # Machinery Release Notes
 
+* Allow inspection of old 32-bit systems which have no
+  SSE instructions (gh#SUSE/machinery#2243)
 
 ## Version 1.23.0 - Mon Aug 28 14:59:52 CEST 2017 - thardeck@suse.de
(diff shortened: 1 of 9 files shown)


greygoo:
LGTM

PR #2246, opened 2018-01-09
Use helper for old architectures

thardeck:
The last fix did create a helper which can run on old systems but the architecture was reported as i586 instead of i686. This pr addresses this issue.


thardeck:
It addresses the second part of the issue https://github.com/SUSE/machinery/issues/2243

PR #2250, opened 2018-06-01
Export kickstart

ksanjeet:
This implements export-kickstart.


opensusegithubapi:

cornelius:
Hey Sanjeet! This is great. Thanks for the contribution. I'll let another developer look at the code in more detail. I think we will need some unit tests and some documentation before we can merge it. This probably can follow the lines of the other export formats so it shouldn't be a big deal. In general the approach looks good to me.

thardeck:
Closing for now, please reopen if you find the time to fix the issues.

PR #2264, opened 2019-04-17
Implement kickstart

ksanjeet:


opensusegithubapi:

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+      raise Machinery::Errors::ExportFailed.new(
+        "Export is not possible because the operating system " \
+        "'#{@system_description.os.display_name}' is not supported."
+      )
+    end
+  end
+
+  def apply_repositories
+    return unless @system_description.repositories
+    ret_val=""
```
thardeck: we use spaces around `=`

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+        @system_description.repositories.each do |repository|
+        ret_val += "\nrepo --name=" + repository.alias.to_s + " --baseurl=" +  repository.url.to_s.chop.reverse.chop.reverse
```
thardeck: No need for `.to_s` if the value is already a string.

Insteaf of concatenation you can also use variables inside of strings, for example

```
ret_val += "\nrepo --name=#{repository.alias} --baseurl=#{repository.url}..."
```

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+    end
+  end
+
+  def apply_repositories
+    return unless @system_description.repositories
+    ret_val=""
+        @system_description.repositories.each do |repository|
+        ret_val += "\nrepo --name=" + repository.alias.to_s + " --baseurl=" +  repository.url.to_s.chop.reverse.chop.reverse  
+      end
+    return ret_val
```
thardeck: Using return at the end of methods is unusual for Ruby: http://ruby-for-beginners.rubymonstas.org/writing_methods/return_values.html

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+    end
+
+  def apply_packages
+    return unless @system_description.packages
+    ret_val="\n%packages"
+      @system_description.packages.each do |package|
+        ret_val += "\n" + package.name.to_s
```
thardeck: For simple cases like this you could use `@system_description.packages.join("\n")`

spec/support/system_description_factory.rb
```diff
@@ -352,6 +352,21 @@ def build_description(name, store, options)
       ]
     }
   EOF
+  EXAMPLE_SCOPES["groups_centos"] = <<-EOF.chomp
```
thardeck: I do not think that we need a special group for centos, since they should be linux/unix universal. I would use the group which is there and adapt the kickstart simple.cfg accordingly.

spec/support/system_description_factory.rb
```diff
@@ -373,6 +388,20 @@ def build_description(name, store, options)
       "architecture": "x86_64"
     }
   EOF
+  EXAMPLE_SCOPES["os_centos7"] = <<-EOF.chomp
+    "os": {
+      "name": "CentOS Linux 7 (Core)",
+      "version": "7 (Core)",
```
thardeck: If you inspect a CentOS 7 with Machinery, this is what is generated in the machinery manifest? I am wondering because CentOS 7 has a code name but RHEL 7 not.

spec/support/system_description_factory.rb
```diff
@@ -782,6 +867,27 @@ def build_description(name, store, options)
       ]
     }
   EOF
+
+  EXAMPLE_SCOPES["users_with_passwords_centos"] = <<-EOF.chomp
```
thardeck: same as for the groups, I would use the already existing example user scope `users_with_passwords`.

spec/support/system_description_factory.rb
```diff
@@ -810,6 +916,10 @@ def build_description(name, store, options)
           "name": "sshd.service",
           "state": "enabled"
         },
+        {
+          "name": "NetworkManager.service",
+          "state": "enabled"
+        },
```
thardeck: You can not just extend an example, unless you adapt all the others tests which are based on it.

For the unit tests it is ok, if some services are missing from the kickstart manifest, as long as the result is consistent.

spec/unit/kickstart_spec.rb
```diff
@@ -0,0 +1,195 @@
...
+        to include("var/log/*")
+    end
+
+    it "adds the kickstart export readme" do
+      expect(File.exist?(File.join(@output_dir, "README.md"))).to be(true)
+    end
+
+    it "adds the ip of the outgoing network to the readme" do
+      file = File.read(File.join(@output_dir, "README.md"))
+      expect(file).to include("ks=http://#{ip}:8000/ks.cfg")
```
thardeck: I know that we also do this in the Autoyast test, but I think it is better to expect the whole string for readability and to prevent issues if the value is empty by accident:
```
expect(file).to include("ks=http://192.168.0.35:8000/ks.cfg")
```

export_helpers/kickstart_export_readme.md
```diff
@@ -0,0 +1,41 @@
+# README for KickStart export from Machinery
```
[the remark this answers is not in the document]
ksanjeet: commited the suggested change.

export_helpers/kickstart_export_readme.md
```diff
@@ -0,0 +1,41 @@
...
+The export directory contains both the KickStart profile and additional data that
+is required during installation. This directory needs to be made available to
+the installer via network, e.g. by running:
+
+  cd <path>; python -m SimpleHTTPServer
+
+You can then point the installer to the profile by specifying the KickStart
+option on the kernel command line. You also need to mention the required setting for setting up network.
+
+For RHEL 6&7 and CentOS 6&7:
```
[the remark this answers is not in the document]
ksanjeet: Committed the suggested change.

export_helpers/kickstart_export_readme.md
```diff
@@ -0,0 +1,41 @@
...
+The installation via for example an HTTP server is only possible if all files
+and sub directories are readable by the HTTP server user.
+To make the export directory readable for all users run:
+
+  chmod -R a+rX <path>
+
+## Reference
+
+Additional information regarding KickStart is available in the official documentation:
+
```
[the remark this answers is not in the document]
ksanjeet: Committed the suggested change.

lib/cli.rb
```diff
@@ -587,6 +578,39 @@ def self.check_port_validity(port)
...
+
+      The profile will be placed in a subdirectory at the given location by
+      the 'kickstart-dir' option.
+    LONGDESC
+    arg "NAME"
+    command "export-kickstart" do |c|
+      c.flag ["kickstart-dir", :a],
+        type:     String,
+        required: true,
+        desc:     "Location where the autoyast profile will be stored",
```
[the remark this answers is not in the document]
ksanjeet: commited the suggested change.

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+      "packages"
+    )
+    check_exported_os
+    unless description.users
+      Machinery::Ui.puts(
+        "\nWarning: Exporting a description without the scope 'users' as KickStart" \
+        " profile will result in a root account without a password which prevents" \
+        " logging in.\n" \
+        "So either inspect or add the scope 'users' before the export or" \
+        " add a section for the root user to the KickStart profile."
```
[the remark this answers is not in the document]
ksanjeet: I'll test it. However, I 've one observation that the kickstart autoinstallation doesn't proceed without a root password which is the reason I had to put separate "rootpw" section in the kickstart profile apart from the users.

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+  end
+
+
+  def profile(output_dir)
+    File.open(File.join(output_dir, "ks.cfg"),"w") do |f|
+      f.write("lang en_us.UTF-8" +
+        "\ngraphical" +
+        "\ninstall" +
+        "\neula --agreed" +
+        "\nauth --enableshadow --passalgo=sha512" +
```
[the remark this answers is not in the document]
ksanjeet: I've tested it with CentOS 7. From my past experience I guess it should work with the previous versions as well. I'll anyhow test it and confirm.

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+        "\ncdrom" +
+        "\nfirstboot --enabled" +
+        "\nclearpart --all --initlabel" +
+        "\npart /boot --fstype ext4 --size=500" +
+        "\npart swap --size=1024" +
+        "\npart / --fstype ext4 --size=13480 --grow --asprimary" +
```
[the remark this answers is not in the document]
ksanjeet: There is an "autopart" option. I'll test it and confirm.

lib/kickstart.rb
ksanjeet: I've made the suggested change. I'll commit it once all changes in this file are made.

lib/kickstart.rb
```diff
@@ -0,0 +1,254 @@
...
+        "Export is not possible because the operating system " \
+        "'#{@system_description.os.display_name}' is not supported."
+      )
+    end
+  end
+
+  def apply_repositories
+    return unless @system_description.repositories
+    ret_val=""
+        @system_description.repositories.each do |repository|
```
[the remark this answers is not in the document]
ksanjeet: I've made the suggested change. I'll commit it once all changes in this file are made.

thardeck:
@ksanjeet could you rebase against `master` so Travis can verify your commits?
I think it would be also better to use a new pr with squashed commits so the review will be easier.
Thanks in advance.

PR #2269, opened 2019-06-25
Fix raise_error expectation without specific error messages

thardeck:


thardeck:
Prophet is failing because of a ruby gem issue, but rspec tests were ok. I am planning to switch to travis to be independent of this.

PR #2279, opened 2020-11-04
Fix file encoding when creating archive files

guangyee:
When creating the archive files, we need to open them with "wb"
instead of "w" as the underneath cheetah library is doing partial
read, which always result int ASCII-8BIT encoding [1]. Otherwise,
we'll get an ugly error like this.

"\x8B" from ASCII-8BIT to UTF-8"

in cheetah [2] when attempt to do buffer write to the archive file.

[1] https://ruby-doc.org/core-2.7.0/IO.html
[2] https://github.com/openSUSE/cheetah/blob/master/lib/cheetah.rb#L640

### DIFF
--- a/lib/system.rb
+++ b/lib/system.rb
@@ -74,7 +74,7 @@ class Machinery::System
       "The following files are packaged in #{archive}: " + Array(file_list).join(", ")
     )
     created = !File.exist?(archive)
-    out = File.open(archive, "w")
+    out = File.open(archive, "wb")
     begin
       run_command(
         "tar", "--create", "--gzip",


lib/system.rb
```diff
@@ -74,7 +74,7 @@ def create_archive(file_list, archive, exclude = [])
       "The following files are packaged in #{archive}: " + Array(file_list).join(", ")
     )
     created = !File.exist?(archive)
-    out = File.open(archive, "w")
+    out = File.open(archive, "wb")
```
jeremy-moffitt, line 77: the ruby docs are not particularly helpful on this (https://ruby-doc.org/core-2.7.2/File.html  doesnt describe the modes available) , I assume thats writing as binary or something similar?

jeremy-moffitt: oops, your IO link does have it
```
"b"  Binary file mode
     Suppresses EOL <-> CRLF conversion on Windows. And
     sets external encoding to ASCII-8BIT unless explicitly
     specified.
```

bear454: https://ruby-doc.org/core-2.7.2/IO.html#method-c-new-label-IO+Open+Mode

jeremy-moffitt (approved):
LGTM, thanks for linking the relevant docs

PR #2281, opened 2021-02-05
Statically linked machiner-helper executables

guangyee:
We want them to be portable and not depended on glibc DLL.


jeremy-moffitt (approved):
looks ok to me... from what I can google about CGO_ENABLED=0 , this should be fine

mbelur (approved):
LGTM! Tested the PR and no longer have the GLIBC version incompatibility issue.

bear454 (approved):

PR #2282, opened 2021-02-11
Use dnf instead of yum for CentOS 8 repo discovery

guangyee:
Yum had been deprecated in favor of dnf in CentOS 8. Therefore,
we should be use dnf for repo discovery.


mbelur (approved):
LGTM

jeremy-moffitt (approved):
LGTM, thanks for addressing this so quickly

chajain (approved):
LGTM

PR #2283, opened 2021-02-20
Add an options to skip exporting UID, GID, and package version

guangyee:
If a given UID has already been taken by a different user on a target machine,
user creation will fail. Therefore, we need to provide an option to skip
exporting the UIDs from source machine to avoid potential UID conflicts
when recreating the user on the target machine. In this case, a new UID will
be allocated during user creation as per

https://docs.saltproject.io/en/latest/ref/states/all/salt.states.user.html

To skip migrating UID, use '--skip-uid' option.

Likewise, use '--skip-gid' option to skip GID migration.

When migrating to a higher OS version, we may not want to migration to package
version as that would constitutes a package downgrade. Furthermore, the
specific package version may no longer be supported in the higher OS version.
Therefore, we need an option to skip migrating the package version.

To skip migrating package version, use '--skip-package-version' option.

### DIFF
--- a/export_helpers/unmanaged_files_salt_excludes
+++ b/export_helpers/unmanaged_files_salt_excludes
@@ -7,6 +7,7 @@ etc/shadow*
 etc/sysconfig/network/*
 etc/group*
 etc/udev/rules.d/70-persistent-net.rules
+etc/zypp/repos.d/*
 lib/mkinitrd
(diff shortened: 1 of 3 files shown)


mbelur (approved):
LGTM!

jeremy-moffitt (approved):

PR #2290, opened 2025-04-14
Make Tumbleweed recognized again ... without side effects

mge1512:


jeremy-moffitt (approved):
Looks ok to me, I also asked a team member more familiar with Ruby than I am and they didn't see any issues. I'm not sure if the hound flags need resolved or if thats just a nag as its been a couple years since my team did anything with the machinery repo.

bear454 (approved):

